PR #92, opened 2015-01-09
Lower CPU usage

adamgreen:
When I had a device halted under pyOCD, the CPU usage on my development
machine would sky rocket.  The effect of this was very noticeable when
running off of batteries.  This was due to pyOCD's usage of
non-blocking socket reads in a tight loop while it tries to read
packets sent from GDB.  I added a sleep() call to the run() method as
already used in the resume() method.  I set both of these sleeps to be
100 milliseconds. For me, I didn't find this decreased the perceived
responsiveness of pyOCD when I was using GDB and it greatly reduces the
CPU usage.

**Edit**: Below feature is no longer required after PR #93 so it has been
removed from this PR.

~~As I have been working to implement pyOCD support for the LPC4330, I
would hit issues where resume() would hit an excpeption as it tried to
read the hard fault handler address from the interrupt vector table at
an address of 0x12.  This happened because the external SPIFI FLASH
isn't always available and mapped into this shadow region.  I added
code which handles this exception and treats it the same as when no
hardware breakpoints are available.~~

### DIFF
--- a/pyOCD/flash/flash.py
+++ b/pyOCD/flash/flash.py
@@ -1,6 +1,6 @@
 """
  mbed CMSIS-DAP debugger
- Copyright (c) 2006-2013 ARM Limited
+ Copyright (c) 2006-2015 ARM Limited
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
@@ -54,7 +54,7 @@ class Flash(object):
         # check the return code
         result = self.target.readCoreRegister('r0')
         if result != 0:
-            logging.error('eraseAll error: %i', result)
+            logging.error('init error: %i', result)
 
         return
 
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -143,6 +143,7 @@ class GDBServer(threading.Thread):
                     if (new_command == True):
                         new_command = False
                         break
+                    sleep(0.1)
                     try:
                         if self.shutdown_event.isSet() or self.detach_event.isSet():
                             break
@@ -301,7 +302,7 @@ class GDBServer(threading.Thread):
         val = ''
         
         while True:
-            sleep(0.01)
+            sleep(0.1)
             if self.shutdown_event.isSet():
                 return self.createRSPPacket(val), 0, 0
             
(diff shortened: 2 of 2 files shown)


c1728p9:
I use eclipse (KDS) as a front end to GDB and this patch makes stepping noticeable slower for me.  Below is some timing information I recorded on my machine.  I used the same test C code for both runs, and recorded the time it took to step over the first two functions.

Before Patch:
Step - 33 gdb commands, 0.809000015 s to complete
Step - 32 gdb commands, 0.80099988 s to complete

With patch:
Step - 33 gdb commands, 4.209000111 s to complete
Step - 32 gdb commands, 4.003999949 s to complete


Would it be possible to modify your patch to only start sleeping a fixed amount of time after the last command completed?  I think this would keep CPU usage at a minimum, and also keep the faster performance when using eclipse.  Something like the below code:

```
while True:
    ...
    if time() - last_command > 0.5
        # Sleep if no command has been sent for 500ms
        sleep(0.1)
    ... 
    if command:
        ...
        last_command = time()
    ...
```

0xc0170:
@c1728p9 Thanks for testing this PR.

adamgreen:
@c1728p9 Thanks for the feedback.  I will incorporate your sleep change later today.

I would be happy to see you cleanup that hard fault code.  I had switching to the vector catch on my to do list but if you do it then that is even better :)

adamgreen:
@c1728p9 I have added the back-offs that you mentioned in your above comments.  Hopefully it helps with your single stepping issue with this commit.

c1728p9:
I'll give this patch a try and let you know.  Also, I have the vector catch implemented in #93.

c1728p9:
Looks like this fixes the eclipse slow downs.  I looks like its even a little faster.  Thanks for fixing this.

Before Patch:
Step - 31 gdb commands, 0.811000109s to complete
Step - 30 gdb commands, 0.789000034s to complete

With patch:
Step - 31 gdb commands, 0.784999847s to complete
Step - 30 gdb commands, 0.778999805s to complete

adamgreen:
Excellent!

Thanks again,

Adam

adamgreen:
Rebased and retested on NRF51822 and LPC1768.

0xc0170:
Thanks 01000001 01000111 :)

PR #93, opened 2015-01-10
Fix resume, prevent stepping into interrupts and add hard fault vector catch

c1728p9:
When using GDB, a typical way of moving the execution address is to set
a breakpoint with the command 'break [line]' and then jump to that
address with the command 'jump [line]'.  This does not work because
the function maybeSkipBreakpoint will cause GDB to skip the breakpoint.
This patch fixes that problem by removing the function
maybeSkipBreakpoint.

GDB single stepping still works as expected and does not get stuck
when trying to step over breakpoints.  This is because the GDB
client automatically removes the breakpoint when stepping over it
and restores it after.

When using Eclipse as a GDB front end this fixes the problem where
where "Move to line" causes the device to begin executing code,
rather than just moving the PC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,9 @@ A series of tests are provided in the test directory:
   * "-l", "--list", help = "List all the connected board"
   * "-d", "--debug", help = "Set the level of system logging output, the available value for DEBUG_LEVEL: debug, info, warning, error, critical"
   * "-t", "--target", help = "Override target to debug"
-  * "-n", "--nobreak", help = "Disable breakpoint at hardfault handler. Required for nrf51 chip with SoftDevice based application."
+  * "-n", "--nobreak", help = "Disable halt at hardfault handler."
+  * "-r", "--reset-break", help = "Halt the target when reset."
+  * "-s", "--step-int", help = "Allow single stepping to step into interrupts."
 
 
 ### Hello World example code
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -68,6 +68,9 @@ class GDBServer(threading.Thread):
         else:
             self.port = port_urlWSS
         self.break_at_hardfault = bool(options.get('break_at_hardfault', True))
+        self.board.target.setVectorCatchFault(self.break_at_hardfault)
+        self.break_on_reset = options.get('break_on_reset', False)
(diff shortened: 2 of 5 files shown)


adamgreen:
What happens now if you try to step over code which actually has a hard coded breakpoint in the machine code?

c1728p9:
If you try to step over a breakpoint in flash you will always halt at the breakpoint and will not be able step over it.  This was the same behavior before my patch.  The function maybeSkipBreakpoint only checks hardware breakpoints.

c1728p9:
Verified basic_test.py still runs without problems on a K22F freedom board.

adamgreen:
There are some very cool commits in this PR.

I tried it on my LPC1768 with no problems.  On my NRF51822 I couldn't use the "load" command from within GDB.  When I do, it gets into an infinite loop between your various reset routines.  Here is a snippet of the callstack:
```
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/target_nrf51822.py", line 73, in resetStopOnReset
    CortexM.resetStopOnReset(self, True)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/cortex_m.py", line 652, in resetStopOnReset
    self.reset(software_reset)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/target_nrf51822.py", line 46, in reset
    self.resetStopResume()
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/target_nrf51822.py", line 53, in resetStopResume
    self.resetStopOnReset()
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/target_nrf51822.py", line 73, in resetStopOnReset
    CortexM.resetStopOnReset(self, True)
```

c1728p9:
Thanks for testing and good catch.  I updated the PR to fix the NRF51822.  Unfortunately I don't have a NRF51822 development board so I am unable to run the fix to confirm it works.  Could you give it another try and let me know if it fixes the problem?

adamgreen:
Yep, that fixed the "load" issue on my NRF51822 board.

Thanks!

c1728p9:
This pull request should be ready to merge unless anyone wants changes to it.

0xc0170:
@c1728p9 
Thanks, I have this in my queue for review/merge.

c1728p9:
Excellent, just making sure no one was waiting on me.

PR #99, opened 2015-02-01
Improve dapTransfer speed by queuing packets

c1728p9:
Send multiple transfer block commands before reading any responses.
This allows multiple packets to be buffered up, which improves
 transfer speeds.  On the K22F with firmware 0203 and a clock of 10MHz
this patch brings the data transfer rate from ~25,000 B/s (before) to
~45,000 B/s (after).

### DIFF
--- a/test/speed_test.py
+++ b/test/speed_test.py
@@ -1,0 +1,116 @@
+"""
+ mbed CMSIS-DAP debugger
+ Copyright (c) 2015 ARM Limited
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+import os, sys
+from time import sleep, time
+from random import randrange
+
+parentdir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(0, parentdir)
+
+import pyOCD
+from pyOCD.board import MbedBoard
+
+import logging
+logging.basicConfig(level=logging.INFO)
+
+with MbedBoard.chooseBoard(frequency = 1000000) as board:
+    target_type = board.getTargetType()
+
+    test_clock = 10000000
+    if target_type == "kl25z":
+        ram_start = 0x1ffff000
+        ram_size = 0x4000
+        rom_start = 0x00000000
+        rom_size = 0x20000
+    elif target_type == "kl46z":
+        ram_start = 0x1fffe000
+        ram_size = 0x8000
+        rom_start = 0x00000000
+        rom_size = 0x40000
+    elif target_type == "k22f":
+        ram_start = 0x1fff0000
+        ram_size = 0x20000
+        rom_start = 0x00000000
+        rom_size = 0x80000
+    elif target_type == "lpc11u24":
+        ram_start = 0x10000000
+        ram_size = 0x2000
+        rom_start = 0x00000000
+        rom_size = 0x8000
+    elif target_type == "lpc1768":
+        ram_start = 0x10000000
+        ram_size = 0x8000
+        rom_start = 0x00000000
+        rom_size = 0x80000
+    elif target_type == "lpc4330":
+        ram_start = 0x10000000
+        ram_size = 0x20000
+        rom_start = 0x14000000
+        rom_size = 0x100000
+    elif target_type == "nrf51822":
+        ram_start = 0x20000000
+        ram_size = 0x4000
+        rom_start = 0x00000000
+        rom_size = 0x40000
+        # Override clock since 10MHz is too fast
+        test_clock = 1000000 
+    else:
+        raise Exception("The board is not supported by this test script.")
+
+    target = board.target
+    transport = board.transport
+    flash = board.flash
+    interface = board.interface
+
+    transport.setClock(test_clock)
+
+    print "\r\n\r\n------ TEST RAM READ / WRITE SPEED ------"
+    test_addr = ram_start
+    test_size = ram_size
+    data = [randrange(1, 50) for x in range(test_size)]
+    start = time()
+    target.writeBlockMemoryUnaligned8(test_addr, data)
+    stop = time()
+    diff = stop-start
+    print("Writing %i byte took %s seconds: %s B/s" % (test_size, diff,  test_size / diff))
+    start = time()
+    block = target.readBlockMemoryUnaligned8(test_addr, test_size)
+    stop = time()
+    diff = stop-start
+    print("Reading %i byte took %s seconds: %s B/s" % (test_size, diff,  test_size / diff))
+    error = False
+    for i in range(len(block)):
+        if (block[i] != data[i]):
+            error = True
+            print "ERROR: 0x%X, 0x%X, 0x%X!!!" % ((addr + i), block[i], data[i])
+    if error:
+        print "TEST FAILED"
+    else:
+        print "TEST PASSED"
+
+    print "\r\n\r\n------ TEST ROM READ SPEED ------"
+    test_addr = rom_start
+    test_size = rom_size
+    start = time()
+    block = target.readBlockMemoryUnaligned8(test_addr, test_size)
+    stop = time()
+    diff = stop-start
+    print("Reading %i byte took %s seconds: %s B/s" % (test_size, diff,  test_size / diff))
+    print "TEST PASSED"
+    
+    target.reset()
--- a/pyOCD/board/board.py
+++ b/pyOCD/board/board.py
@@ -53,6 +53,12 @@ class Board(object):
         """
         logging.debug("init board %s", self)
         self.interface.init()
+        packet_count = self.getPacketCount()
+        logging.info("board allows %i concurrent packets", packet_count)
+        if packet_count < 1:
+            logging.error('packet count of %i outside of expected range', packet_count)
+            packet_count = 1
+        self.interface.setPacketCount(packet_count)
         self.transport.init(self.debug_clock_frequency)
         self.target.init()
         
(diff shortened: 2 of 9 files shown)


c1728p9:
Verified this patch on windows with a K22F board.  Before merging this pull request it would probably be good to have someone try it on linux (or using something other than pywinusb) to make sure the other supported platforms support having multiple packets queued up in the HID driver.

Also, using tweaked firmware and pywinserial as the interface layer, along with this patch, I was able to get data transfer rates of +90KB/s on a KL25 freedom board.  Is there any interest in investigating other non-HID interfaces so the 64KB/s data transfer limit can be bypassed?

0xc0170:
@c1728p9 how did you tweak the firmware? mbed cmsis-dap?

I will tests this. I believe @adamgreen as well :)

c1728p9:
Yeah, I modified the cmsis-dap firmware so HID was disabled and the serial port was instead used for debug data.  If you want to try it I put up the patches I was using:
https://github.com/c1728p9/CMSIS-DAP/commit/0c2801b7db4668d4ccafaa6599b00ca7880c9cdb
and
https://github.com/c1728p9/pyOCD/commit/3afb8ad2ad6e63772c5686739aee9fb0625e2f9d

ECNU3D:
Hi @c1728p9 

I'm trying to use your patch with other target, and I noticed that you're using the existing cdc data port, which means there will be a mixture of debug data and user data. So it's only a patch for test, and in the formal patch we may need two serial port right?

c1728p9:
Hello @ECNU3D 

You are correct, this those two patches are for testing only.  The uart data has been disabled so only debugging data is sent over the serial port.  In a more format patch a different protocol all together might be used.  The reason I chose CDC for testing was because it used bulk endpoints rather than interrupt endpoints, so it was not limited to a 64KB/s speed.

ECNU3D:
Thanks for your clarification, and it's more clear to me now. I think such an approach is quite reasonable. Actually if pyOCD can support semi-hosting, then only port is needed.

ECNU3D:
Thanks for your clarification, and it's more clear to me now. I think such an approach is quite reasonable. Actually if pyOCD can support semi-hosting, the uart data might be merged with the debug data(the only problem is user might still want a standalone uart port).

adamgreen:
I tested this PR on a Ubuntu 12.04 VM with mixed results:
* On the LPC1768 and LPC11U24 "load" was successful but no faster than without the PR.
* On the KL25Z and NRF51822 "load" hangs and the only way to recover is to unplug the USB cable.

I don't know if this is a side effect of the CMSIS-DAP firmware differences between the two classes of devices (the official mbed devices use a closed source interface firmware and the others use the open source CMSIS-DAP firmware) or something different about the block transfer activity based on differences in the target/flash Python code for these devices.

adamgreen:
>Is there any interest in investigating other non-HID interfaces so the 64KB/s data transfer limit can be bypassed?

I thought pyOCD was supposed to be the daemon for CMSIS-DAP and by design CMSIS-DAP was supposed to be HID based so that it required no drivers?

c1728p9:
Thanks for testing @adamgreen .  It is unfortunate that it hangs when running on some boards.  Was it hanging inside dapTransferBlock?  Also, which version of CMSIS-DAP firmware were you using on the KL25 board?

As for the LPC11U24, did you run a block transfer speed test similar speed_test.py in my patch?  Load might not occur much faster if the majority of the time is spent on sector writes/erases and not transferring the data over USB.

As for your second question, yes, I think the goal was to keep CMSIS-DAP configured so it didn't require any extra drivers.  It might still be possible to use CMSIS-DAP without any additional drivers, if another existing interface similar to HID (but with bulk endpoints) is used.  Maybe this has already been checked by someone and there are none suitable, but I figured I would ask in case anyone had ideas about this.

adamgreen:
>Was it hanging inside dapTransferBlock?

It hangs in the interface.read() call within dapTransferBlock().

I was able to reduce the MAX_PENDING_READS to 6 and have it work on the KL25Z but had to drop it to 4 to get it working with the NRF51822.  It appears to be sensitive to the particular board.

> Also, which version of CMSIS-DAP firmware were you using on the KL25 board?

I was originally running 0201 but I upgraded to 0203 with the same result.

>As for the LPC11U24, did you run a block transfer speed test similar speed_test.py in my patch?

I hadn't.  I see that your speed_test.py is already setup to work with the KL25Z so I ran the test on that device.

These are the results when running the code from the current master branch:
```
------ TEST READ / WRITE SPEED ------
Writing 16384 byte took 1.28158807755 seconds: 12784.1389032 B/s
Reading 16384 byte took 1.28282189369 seconds: 12771.8431378 B/s
TEST PASSED
Reading 131072 byte took 10.253330946 seconds: 12783.357983 B/s
```

And these are the results with your change:
```
------ TEST READ / WRITE SPEED ------
Writing 16384 byte took 1.2824640274 seconds: 12775.4070679 B/s
Reading 16384 byte took 1.28190517426 seconds: 12780.9765722 B/s
TEST PASSED
Reading 131072 byte took 10.2539038658 seconds: 12782.6437341 B/s
```

On Ubuntu 12.04 this change doesn't appear to work.  It doesn't increase the read/write rate and it looks like it will result in a hang if it can't read out the results faster than the CMSIS-DAP firmware sends them back.

0xc0170:
@c1728p9 There has not be any answer, is there any progress? Shall we close this PR?

c1728p9:
I was able to reproduce @adamgreen s problems using pyusb on windows but do not have a fix yet. I was out of town last weekend, and this weekend I was busy so I didn't get to work on this much.  I plan on investigating this further when I have time.

c1728p9:
@adamgreen can you try this patch again and see if this fixes the problems you were seeing?

c1728p9:
I probably should elaborate further on my findings over the weekend.  I tracked the dropped packets down to the device firmware which on the KL25 is only configured to handle 5 packets - "#define DAP_PACKET_COUNT 5".  There was one additional spot for data in the transmit buffer, which is why 6 packets worked correctly.  The reason 8 packets worked with pyWinUSB is because there was a dedicated reader thread which would read back the data faster then it was sent, preventing all 5 slots from being used.  This is also why the speeds for pyWinUSB were higher.

Fortunately, there is a command to query the packet count.  This patch uses that value to determine how many packets can be queued for a given board.  This prevents the debugger firmware from dropping packets which in turn prevents the pc software from getting stuck.

0xc0170:
I tested this PR on windows machine with LPC812, KL25Z and Nordic, run the basic test and speed test you pointed to, for KL25Z. Did see the speed increase in the speed test. I dont have set up mac at the moment.

Are you planning to expand that speed test to other targets? Or at least add it to pyOCD tests, so we can expand it by other targets?

c1728p9:
Good to hear.  Do you remember what speed you saw when you ran the test?

I think a speed test would be good to have.  Do you want me to add it to basic_test.py or create a dedicated script for it?

0xc0170:
probably own dedicated script, which can be generic like basic_test.py to be able to run on any pyOCD device.

c1728p9:
Added a speed test script which supports the KL25Z, K22F and KL46Z.  If you can supply me with the ram/rom info on the other processors I can add them to this and the flash script in my other pull request.  I'll need someone else to verify those targets though, since I only own the above 3 boards.

adamgreen:
Sorry I took so long to look at this pull request.  I tested it out tonight on my KL25Z, LPC1768, LPC11U24, LPC4330, and NRF51822 boards.

I only hit one issue with this pull request.  The NRF51822 board uses EP0 for writes so I had to modify that code path to also release the semaphore:
```diff
--- a/pyOCD/interface/pyusb_backend.py
+++ b/pyOCD/interface/pyusb_backend.py
@@ -152,16 +152,16 @@ class PyUSB(Interface):
             bmRequest     = 0x09              #Set_REPORT (HID class-specific request for transferring data over EP0)
             wValue        = 0x200             #Issuing an OUT report
             wIndex        = self.intf_number  #mBed Board interface number for HID
+            self.read_sem.release()
             self.dev.ctrl_transfer(bmRequestType,bmRequest,wValue,wIndex,data)
             return
-            #raise ValueError('EP_OUT endpoint is NULL')
-
+
         self.read_sem.release()
         self.ep_out.write(data)
         #logging.debug('sent: %s', data)
         return


     def read(self):
         """
         read data on the IN endpoint associated to the HID interface
```

I made a few updates to speed_test.py to support a few additional boards:
```diff
--- a/test/speed_test.py
+++ b/test/speed_test.py
@@ -1,6 +1,6 @@
 """
  mbed CMSIS-DAP debugger
- Copyright (c) 2006-2013 ARM Limited
+ Copyright (c) 2015 ARM Limited

  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
@@ -46,8 +46,28 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board:
         ram_size = 0x20000
         rom_start = 0x00000000
         rom_size = 0x80000
+    elif target_type == "lpc11u24":
+        ram_start = 0x10000000
+        ram_size = 0x2000
+        rom_start = 0x00000000
+        rom_size = 0x8000
+    elif target_type == "lpc1768":
+        ram_start = 0x10000000
+        ram_size = 0x8000
+        rom_start = 0x00000000
+        rom_size = 0x80000
+    elif target_type == "lpc4330":
+        ram_start = 0x10000000
+        ram_size = 0x20000
+        rom_start = 0x14000000
+        rom_size = 0x100000
+    elif target_type == "nrf51822":
+        ram_start = 0x20000000
+        ram_size = 0x4000
+        rom_start = 0x00000000
+        rom_size = 0x40000
     else:
-        raise Exception("A board is not supported by this test script.")
+        raise Exception("The board is not supported by this test script.")

     target = board.target
     transport = board.transport
```

**NRF51822 Note**: This board doesn't work with a 10MHz SWD clock.  I switched it back to 1MHz when running on this board.  Maybe speed_test should allow the user to provide a frequency override on the command line or change the SWD clock rate once it knows which board is connected.

0xc0170:
@adamgreen what OS did you use? Thanks for sharing those diffs.

adamgreen:
@0xc0170 Ubuntu 12.04

test/speed_test.py
```diff
@@ -0,0 +1,91 @@
+"""
+ mbed CMSIS-DAP debugger
+ Copyright (c) 2006-2013 ARM Limited
```
0xc0170: Please correct the year, I recall Adam highlighted this in his diff

c1728p9:
Thanks for the testing and the code.  I updated the patch to address the feedback.  Let me know if I missed anything.

adamgreen:
Latest patch looks good to me! :+1: 

I retested on NRF51822 and LPC1768.

PR #100, opened 2015-02-05
Add LPC4330 support

adamgreen:
Add support for the LPC4330 based Micromint Bambino 210/210E boards to pyOCD.

My previous pull request for FPU register support was leading up to this pull request since the LPC43xx device has a Cortex-M4F core.

Most of the changes in this pull request are isolated to this specific device and don't have an impact on existing devices.  The few changes which touch code related to other devices include:
* In Flash::flashBinary(), I switched the flashPtr local variable to be an optional parameter instead.  This parameter defaults to a value of 0 which is what the local variable used to be initialized to before.  This allows tests for the LPC4330 to use a non-zero base address when using this method.  This was required since the external SPIFI FLASH is based at 0x14000000.
* I updated basic_test.py to add LPC4330 tests.  These tests take advantage of the new flashPtr parameter in Flash::flashBinary() noted above.

I have been dogfooding these LPC4330 changes for a few weeks now as I have been developing code for this device.

I ran basic_test.py on the LPC4330, the NRF51822, and the KL25Z.  I also tried it on the LPC1768 but it failed in the erasePage() tests since that device's flashing code doesn't currently support that operation and it also fails without my changes.

See individual commit comments for more information.


pyOCD/board/mbed_board.py
```diff
@@ -38,6 +38,8 @@
                 "1070": "nrf51822",
                 "1080": "stm32f103rc",
                 "1090": "stm32f051",
+                "1600": "lpc4330",
```
0xc0170: why there are two ids for one board?

adamgreen: They use two different IDs.  One for the Bambino 210 and another for the Bambino 210E.  This should support them both.
https://github.com/micromint/CMSIS-DAP-lpc43xx/blob/master/interface/Common/src/board.c#L77
https://github.com/micromint/CMSIS-DAP-lpc43xx/blob/master/interface/Common/src/board.c#L82

PR #104, opened 2015-02-22
Speed up flash programming

c1728p9:
Add smarter flash programming.  This patch creates a single function
for flash programming called flashBlock.  This function tries to flash
in the fastest manner possible by only programming pages that have
new data.  This behavior can be overwritten by setting the parameter
smart_flash to False.  This patch also adds flash_test.py to test the
different flashing modes.


0xc0170:
It uses erasePage which is not yet implemented for all targets, which we should add.I'll look at changes and run some boards test.

c1728p9:
Added an extra test case (offset write) and added the KL46 target.  I verified that flash_test.py passes on the KL46, KL25 and K22F.  This patch should be ready to go (once erasePage is fixed).

adamgreen:
What kind of performance improvement do you think that this PR will get for a user in the typical case?  

I skimmed through this change a bit tonight but didn't actually try it out.  This is a pretty big change.  My personal preference is that when I program a device, it just erase the device and programs in the bits that I have given it with no care about what was placed in it before.

c1728p9:
Here are some rough performance stats:
1. Reloading same or similar code: 2 to 3x speed increase (on my comp with same image 12.6KB/s before to 43.9KB/s after)
2. Reloading code padded with 0xFF: Proportional to the number of sectors that are not erased - if 50% of code is padding then flashing will be 2x faster.
3. Loading a completely new image with no erased pages: About the same time as before - Analysis is fast and takes less than 1s estimate contents of targets flash and to determine the number of erased pages in a 512KB image.

There are parameters to the flashing functions so the smart flashing can be turned off.  If you don't want this behavior set smart_flash=False.  This will force it to always program everything.  Without any parameters specified it will try and determine the fastest way on its own (which typically adds less than 1s in analysis time).

Yeah, I agree, this is a big change.  That is one of the reasons I added a dedicated script to test it - I want to make sure everything is working properly, and as expected.  If you have a chance you should try the patch out, I think you'll like the performance.

c1728p9:
Updated the pull request to support more boards.  @adamgreen if you are brave enough, can you see if flash_test.py works with the lpc4330?  I don't blame you if you don't want to.  I haven't tested flashing any processors with ROM starting at a non zero address. I can't even simulate it with the Kinetis boards I have because erasing address 0x400 locks them.

adamgreen:
@c1728p9 I tried it out on my LPC4330 board tonight.  I had to make some mods to the test code to not hardcode the beginning of FLASH to 0x00000000.

```diff
--- a/pyOCD/flash/flash.py
+++ b/pyOCD/flash/flash.py
@@ -372,7 +372,7 @@ class Flash(object):
             if chip_erase is None:
                 chip_erase = False
             elif chip_erase is True:
-                logging.warning('Chip erase used when flash address is not the same as flash start', addr)
+                logging.warning('Chip erase used when flash address 0x%x is not the same as flash start', addr)

         # Simple flash
         if not smart_flash:


--- a/test/flash_test.py
+++ b/test/flash_test.py
@@ -158,11 +158,11 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     data = struct.unpack("%iB" % len(data), data)
     unused = rom_size - len(data)
     
-    addr = 0
+    addr = rom_start
     size = len(data)
     
     print "\r\n\r\n------ Test Basic Page Erase ------"
-    flash.flashBlock(addr, data, False, False, progress_cb = print_progress)
+    flash.flashBlock(addr, data, False, False, progress_cb = print_progress, flash_start = rom_start)
     data_flashed = target.readBlockMemoryUnaligned8(addr, size)
     if same(data_flashed, data):
         print("TEST PASSED")
@@ -170,9 +170,9 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     else:
         print("TEST FAILED")
     test_count += 1
-    
+
     print "\r\n\r\n------ Test Basic Chip Erase ------"
-    flash.flashBlock(addr, data, False, True, progress_cb = print_progress)
+    flash.flashBlock(addr, data, False, True, progress_cb = print_progress, flash_start = rom_start)
     data_flashed = target.readBlockMemoryUnaligned8(addr, size)
     if same(data_flashed, data):
         print("TEST PASSED")
@@ -180,9 +180,9 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     else:
         print("TEST FAILED")
     test_count += 1
-    
+
     print "\r\n\r\n------ Test Smart Page Erase ------"
-    flash.flashBlock(addr, data, True, False, progress_cb = print_progress)
+    flash.flashBlock(addr, data, True, False, progress_cb = print_progress, flash_start = rom_start)
     data_flashed = target.readBlockMemoryUnaligned8(addr, size)
     if same(data_flashed, data):
         print("TEST PASSED")
@@ -192,7 +192,7 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     test_count += 1
 
     print "\r\n\r\n------ Test Smart Chip Erase ------"
-    flash.flashBlock(addr, data, False, True, progress_cb = print_progress)
+    flash.flashBlock(addr, data, False, True, progress_cb = print_progress, flash_start = rom_start)
     data_flashed = target.readBlockMemoryUnaligned8(addr, size)
     if same(data_flashed, data):
         print("TEST PASSED")
@@ -200,11 +200,11 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     else:
         print("TEST FAILED")
     test_count += 1
-    
+
     print "\r\n\r\n------ Test Offset Write ------"
     new_data = [0x55] * board.flash.page_size * 2
     addr = rom_start + rom_size / 2
-    operation = flash.flashBlock(addr, new_data, progress_cb = print_progress)
+    operation = flash.flashBlock(addr, new_data, progress_cb = print_progress, flash_start = rom_start)
     data_flashed = target.readBlockMemoryUnaligned8(addr, len(new_data))
     if same(data_flashed, new_data) and operation in (FLASH_PAGE_ERASE_BASIC, FLASH_PAGE_ERASE):
         print("TEST PASSED")
@@ -212,14 +212,14 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     else:
         print("TEST FAILED")
     test_count += 1
-    
+
     # Note - The decision based tests below are order dependent since they
     # depend on the previous state of the flash
-    
+
     print "\r\n\r\n------ Test Chip Erase Decision ------"
     new_data = list(data)
     new_data.extend([0xff] * unused) # Pad with 0xFF
-    operation = flash.flashBlock(0, new_data, progress_cb = print_progress)
+    operation = flash.flashBlock(rom_start, new_data, progress_cb = print_progress, flash_start = rom_start)
     if operation == FLASH_CHIP_ERASE:
         print("TEST PASSED")
         test_pass_count += 1
@@ -230,7 +230,7 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     print "\r\n\r\n------ Test Chip Erase Decision 2 ------"
     new_data = list(data)
     new_data.extend([0x00] * unused) # Pad with 0x00
-    operation = flash.flashBlock(0, new_data, progress_cb = print_progress)
+    operation = flash.flashBlock(rom_start, new_data, progress_cb = print_progress, flash_start = rom_start)
     if operation == FLASH_CHIP_ERASE:
         print("TEST PASSED")
         test_pass_count += 1
@@ -241,22 +241,22 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
     print "\r\n\r\n------ Test Page Erase Decision ------"
     new_data = list(data)
     new_data.extend([0x00] * unused) # Pad with 0x00
-    operation = flash.flashBlock(0, new_data, progress_cb = print_progress)
+    operation = flash.flashBlock(rom_start, new_data, progress_cb = print_progress, flash_start = rom_start)
     if operation == FLASH_PAGE_ERASE:
         print("TEST PASSED")
         test_pass_count += 1
     else:
         print("TEST FAILED")
     test_count += 1
-    
+
     print "\r\n\r\n------ Test Page Erase Decision 2 ------"
     # Note: This test can cause progress to go backwards when more accurate estimates are available.
     new_data = list(data)
-    size_same = unused * 5 / 6 
+    size_same = unused * 5 / 6
     size_differ = unused - size_same
     new_data.extend([0x00] * size_same) # Pad 5/6 with 0x00 and 1/6 with 0xFF
     new_data.extend([0x55] * size_differ)
-    operation = flash.flashBlock(0, new_data, progress_cb = print_progress)
+    operation = flash.flashBlock(rom_start, new_data, progress_cb = print_progress, flash_start = rom_start)
     if operation == FLASH_PAGE_ERASE:
         print("TEST PASSED")
         test_pass_count += 1
@@ -264,12 +264,12 @@ with MbedBoard.chooseBoard(frequency = 10000000) as board: #TODO - change to 1MH
         print("TEST FAILED")
     test_count += 1
 
-        
+
     print("\r\n\r\nTest Summary:")
     print("Pass count %i of %i tests" % (test_pass_count, test_count))
     if test_pass_count == test_count:
         print("FLASH TEST SCRIPT PASSED")
     else:
         print("FLASH TEST SCRIPT FAILED")
-    
+
     target.reset()
```

c1728p9:
Thanks for testing and the diff. I updated my patch so it should work with your board now.

c1728p9:
Rebased on top of master and did a few todos.  LPC targets still need to be fixed before this can be merged.

c1728p9:
Added support for K64F and rebased on top of master.  All LPC targets now support page erase thanks to #109.  This patch should be ready to go unless anyone has additional feedback.

Ran basic test on targets KL46, K64F, K22F, LPC810 and LPC11U24 and they all passed.  Also ran flash_test.py on the same targets.  All passed except LPC810 and LPC11U24 due to flash remapping on reset.  I created #111 and verified this fixes the flash tests on the LPC11U24.

I still see problems flashing the LPC810, even with basic_test.py on older patches.  The board runs without problems, but if the data is read back, addresses 0x1c-0x1f do not match the code that was flashed.  This occurs on basic test even without this patch.  I verified that after a page program operation the data buffer in the targets ram is correct, just the data flashed does not match.  This might be a problem with my specific board.

adamgreen:
There is a 32-bit interrupt vector checksum that is stored at address 0x1C for most LPC parts.  The flashing algorithm calculates the correct value and inserts it during programming.  That would account for the mismatch that you are seeing.

c1728p9:
@adamgreen  should the build tools automatically account for this checksum and add it to the binary?  I'm thinking that one solution would be to modify the binary to have the checksum built in.  That way flash_test.py would succeed because this value would match.  Let me know if this sounds like a good solution.

c1728p9:
Overhauled this pull request to add additional flexibility and features:
-Differing page sizes supported (LPC1768)
-Programming progress never goes backwards now
-Updated flash_tool.py and added hex file support
-Moved most of the flashing logic from flash.py to flash_builder.py and updated scoping (added an underscore for private functions)
-Relaxed programming requirements for GDB - data chuncks no longer need to be sent over in order

This pull request should be ready for merging unless anyone has suggestions for improvements.

Locally rebased #115 on top of this and ran automated_test.py.  All tests passed:
```
------ TEST RESULTS ------
Target         Test           Result

kl46z          Basic Test     Pass
kl46z          Speed Test     Pass
kl46z          Flash Test     Pass
k22f           Basic Test     Pass
k22f           Speed Test     Pass
k22f           Flash Test     Pass
lpc1768        Basic Test     Pass
lpc1768        Speed Test     Pass
lpc1768        Flash Test     Pass
k64f           Basic Test     Pass
k64f           Speed Test     Pass
k64f           Flash Test     Pass
lpc800         Basic Test     Pass
lpc800         Speed Test     Pass
lpc800         Flash Test     Pass
lpc11u24       Basic Test     Pass
lpc11u24       Speed Test     Pass
lpc11u24       Flash Test     Pass
```

c1728p9:
Also ran some code coverage metrics and found 98% of the code in flash_builder.py gets tested from running flash_test.py on a single target.  The 3 untested lines were error or warning cases that probably don't warrant a test in flash_test.py (but could easily be added).

0xc0170:
@c1728p9 very good progress here  :+1: , I will look at the latest additional commits.

c1728p9:
Added more advanced flash analysis to quickly determine flash contents before programming based on the CRC32 of each page.  Also other improvements:
-Added CRC based analysis to speed up flash programming
-Added full support for basic chip erase / page erase (program all pages even if data is the same)
-Rebased on master
-Added automated test support to flash_test.py

```
------ TEST RESULTS ------
Target         Test           Result

kl46z          Basic Test     Pass
kl46z          Speed Test     Pass
kl46z          Flash Test     Pass
k22f           Basic Test     Pass
k22f           Speed Test     Pass
k22f           Flash Test     Pass
lpc1768        Basic Test     Pass
lpc1768        Speed Test     Pass
lpc1768        Flash Test     Pass
k64f           Basic Test     Pass
k64f           Speed Test     Pass
k64f           Flash Test     Pass
lpc800         Basic Test     Pass
lpc800         Speed Test     Pass
lpc800         Flash Test     Pass
lpc11u24       Basic Test     Pass
lpc11u24       Speed Test     Pass
lpc11u24       Flash Test     Pass
```

0xc0170:
Shall we share somehow the analyzer ? How people can get source for that blob?

0xc0170:
- os.remove(log_file) in the automated_test.py fails if the file is not present, please add there a check
- using python 2.7.6 - automated_test.py stops working when flashing new binary or later in ROM test speed (testes so far on windows), automated test file is empty then. Using K64F with Version: 0203 Build: May 30 2014. lpc1768 does not seem to have any problems. I should probably update the K64F :-) The report here just in case.

```
Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	python.exe
  Application Version:	0.0.0.0
  Application Timestamp:	527fcf61
  Fault Module Name:	ntdll.dll
  Fault Module Version:	6.1.7601.18247
  Fault Module Timestamp:	521eaf24
  Exception Code:	c0000005
  Exception Offset:	0000000000053290
```

- chip erase decision 2 takes really long on K64F (just a statement)

c1728p9:
-Added check to make sure the log file exists before deleting it
-Added source and build directions for analyzer blob
-Rebuilt blob using GCC 4.9 (saved a couple of bytes)


```
------ TEST RESULTS ------
Target         Test           Result

k64f           Basic Test     Pass  
k64f           Speed Test     Pass  
k64f           Flash Test     Pass  
kl46z          Basic Test     Pass  
kl46z          Speed Test     Pass  
kl46z          Flash Test     Pass  
k22f           Basic Test     Pass  
k22f           Speed Test     Pass  
k22f           Flash Test     Pass  
lpc1768        Basic Test     Pass  
lpc1768        Speed Test     Pass  
lpc1768        Flash Test     Pass  
lpc800         Basic Test     Pass  
lpc800         Speed Test     Pass  
lpc800         Flash Test     Pass  
lpc11u24       Basic Test     Pass  
lpc11u24       Speed Test     Pass  
lpc11u24       Flash Test     Pass  

All tests passed
```

c1728p9:
@0xc0170 I have seen my K64F get into a state where it doesn't talk over CMSIS-DAP and all the tests fail (firmware 0203).  Pressing the reset button fixes this problem.  I haven't seen python crash because of it. What interface were you using when you saw this crash?

As for the chip erase decision 2, that is a worst case scenario where all the flash needs to be programmed.  How long did it take to run?  On my machine it takes ~83 seconds (to program 1MB K64F) which corresponds to ~12KB/s.

pyOCD/flash/analyzer/build.bat
```diff
@@ -0,0 +1,4 @@
+set path=%path%;C:\Program Files (x86)\GNU Tools ARM Embedded\4.9 2014q4\bin
+arm-none-eabi-gcc.exe -O3 -ffunction-sections -fdata-sections -Wall -mcpu=cortex-m0 -mthumb -mfloat-abi=soft -g3 -gdwarf-2 -gstrict-dwarf -T"linker_script.ld" -Wl,-Map,main.map,--gc-sections,-ecompute_crc -n -mcpu=cortex-m0 -mthumb -mfloat-abi=soft -g3 -nostdlib -fpic main.c -o main.elf
+arm-none-eabi-objcopy.exe --output-target binary main.elf main.bin
+python.exe generate_python.py
```
0xc0170: Can you remove .exe extension?

pyOCD/flash/flash_klxx.py
```diff
@@ -74,14 +77,21 @@
...
+               'analyzer_supported' : False     # Not enough space on KL02 and KL05
               };
 
 # @brief Flash algorithm for Kinetis L-series devices.
 class Flash_klxx(Flash_Kinetis):
 
     def __init__(self, target):
-        super(Flash_klxx, self).__init__(target, flash_algo)
+        algo = flash_algo
+        if isinstance(target, KL25Z) or isinstance(target, KL26Z) or isinstance(target, KL46Z):
```
0xc0170: What is this condition handling?

0xc0170:
I left some comments.

I am going to merge this.

pyOCD/flash/analyzer/build.bat
c1728p9: For such a small program I think it is ok to leave it using a simple script.  Multi-platform support could be an additional improvement in the future if the algorithm needs to be rebuilt often.

pyOCD/flash/flash_klxx.py
c1728p9: The KL02 and KL05 didn't have enough space to support the analyzer so this is missing from the default flash_algo.  If the target is a KL25Z, KL26Z or KL46Z then analyzer support is added to flash_algo.

c1728p9:
@0xc0170 you mentioned that the K64F crashed earlier when testing.  I was searching over the Freedom board schematic and it looks like this might be caused by EzPort.  If it is the first time you are connecting USB, or if SW3 is pressed on usb connect or reset the K64F will boot into EzPort mode and disable the SWD interface.  If you have time could you test to see if this is the cause of your crash? You should be able to get your device into EzPort mode by pressing the reset button while the button SW3 is held down.  Then you could try running a script and see if you get this same crash.

jeremybrodt:
The MAXWSNENV and MAX32600MBED platforms were not updated for this change and are now broken. It appears adding "'analyzer_supported' : False" to the flash_<platform>.py file is sufficient to make basic_test.py work again. What needs to be done to make 'analyzer_supported' : True?

c1728p9:
Sorry I missed those boards, I started on this patch a while back (before those boards were added) and missed them when rebasing.  To support the analyzer you need to:
1. In memory reserve 0x600 for code & table
2. Make sure data buffer is big enough to hold 4 bytes for each page that could be checked (ie.  >= num pages * 4)

I'll have a pull request up shortly to add this.

jeremybrodt:
Thanks for the quick response. So something like this?

               'analyzer_supported' : True,
               'analyzer_address' : 0x20001000  # Analyzer 0x20001000..0x20001600

How do I test it?

c1728p9:
I created #130 to fix this.  If you run flash_test.py it should verify that everything is working correctly.  You might need to add ram/rom start and size for the MAX targets to the test to get it to run.

0xc0170:
Thanks @c1728p9 for instant fix :+1:

jeremybrodt:
Thanks. I will add these boards to flash_test.py and test it.

PR #105, opened 2015-02-28
Misc improvements

c1728p9:
This pull request changes the target reset command to default to a software reset instead of a hardware reset.  It also modifies the GDB server to exit when the client detaches, unless the --presist option is specified.


c1728p9:
Verified basic_test.py works with a KL25, KL46 and K22F.

pyOCD/target/cortex_m.py
```diff
@@ -712,17 +712,21 @@ def step(self, disable_interrupts = True):
     def clearDebugCauseBits(self):
         self.writeMemory(DFSR, DFSR_DWTTRAP | DFSR_BKPT | DFSR_HALTED)
 
-    def reset(self, software_reset = False):
+    def reset(self, software_reset = None):
```
0xc0170: Why did you change the default value from False to None?

c1728p9: Internally reset converts None to True.  I set it as None to indicate that no value was specified.  This way calling reset(None) has the same effect as reset().  This allows resetStopOnReset to propagate an unspecified reset value down to reset.

adamgreen:
I ran basic_test.py on KL25Z, NRF51822, LPC4330, LPC11U24, and LPC1768.  It ran successfully on all of them except for the LPC1768.  It doesn't appear to like the software reset.  Here is the output from the test when it fails on the LPC1768:
```
------ TEST RESET ------
ERROR:root:exception during uninit
  File "test/basic_test.py", line 209, in <module>
    target.reset()
  File "/media/psf/depots/pyOCD/pyOCD/target/target_lpc1768.py", line 40, in reset
    CortexM.reset(self, software_reset)
  File "/media/psf/depots/pyOCD/pyOCD/target/cortex_m.py", line 725, in reset
    self.writeMemory(NVIC_AIRCR, NVIC_AIRCR_VECTKEY | NVIC_AIRCR_SYSRESETREQ)
  File "/media/psf/depots/pyOCD/pyOCD/target/cortex_m.py", line 491, in writeMemory
    self.transport.writeMem(addr, value, transfer_size)
  File "/media/psf/depots/pyOCD/pyOCD/transport/cmsis_dap.py", line 133, in writeMem
    [addr, data])
  File "/media/psf/depots/pyOCD/pyOCD/transport/cmsis_dap_core.py", line 215, in dapTransfer
    raise ValueError('SWD Fault')
ValueError: SWD Fault
```
I switched that test's call to target.reset() to target.reset(False) and it got past this failure point (it then fails later during the reset in the subsequent programming test.)

c1728p9:
Do you have any thoughts on why this is crashing, especially when the other LPC targets are working with software reset?  Could it be related to the remapping of the flash - self.writeMemory(0x400FC040, 1)?  I think it is the only target that does this in the reset call.  Or maybe the LPC1768 needs a hardware reset to function properly, in which case I need to revert my changes in this file.

0xc0170:
I can confirm it fails for LPC1768. It returns value 7 as resp[2], the response 0 and 1 are correct. I  can't locate any issue why SYSRESETREQ  would fail for LPC1768

c1728p9:
I updated my patch to revert the changes to the LPC1768 and configure it to always use a hardware reset.  It seems like it might not support software resets.  I believe the LPC1768 is a cortex-m3, which implements the ARMv7m architecture, but digging through the ARM reference manual, I was unable to determine if the AIRCR used for software resets is present on all cortex-m3 devices.  Do you know if this is a guaranteed feature, or if it is implementation defined?

I also checked the LPC's datasheet, and it sounds like software resets are not supported.  In the document UM10360 (LPC176x/5x User manual) in sections 3.4 Reset, it says:
"Reset has 4 sources on the LPC176x/5x: the RESET pin, Watchdog Reset, Power On
Reset (POR), and Brown Out Detect (BOD)."

0xc0170:
@c1728p9 http://sourceforge.net/p/openocd/mailman/message/31873092/, a patched which switched to sw reset for lpc1768, tested on mbed platform.

The rev3 says that SYSRESETREQ is supported.

c1728p9:
@0xc0170 If you get a chance, could you try my patch with self.writeMemory(0x400FC040, 1) removed from the lpc1768 and see if reset works as expected?  I'm wondering if this will fix the problem, since the lpc800 and lpc11u24 work without this.

0xc0170:
I did previously, while debugging it. Did not have any impact.

0xc0170:
I have just tested the uvision mbed project with LPC1768, the flashing fails with SYSRESET, but fine with VECTRESET or HWRESET. This indicates something. So fine for now that target set to not use sys reset.

c1728p9:
I updated this pull request to include a fix for the OpenOCD eclipse plugin.  I re-ran basic_test.py on a K22F an there were no problems.  I think this patch is ready to be merged unless there is additional feedback anyone has.  Thanks for help with this PR with all the research and testing.

c1728p9:
Found a bug with dapTransferBlock so I updated the pull request to include and additional commit.

c1728p9:
Updated to include LPC sector programming fixes.  Verified basic_test.py works on a LPC800 and a LPC11U24.

0xc0170:
Thanks for this patches. How did you find out about the page size difference?

I tested on lpc1768, it does not crash but erase sector fails, which we can fix later. I'll merge this and release new version of pyOCD 0.3.1.

0xc0170:
:thumbsup: for this work

c1728p9:
What kind of failure are you seeing on sector erase?  Is the return value non-zero?

0xc0170:
data != expected. Seeing data as a list of 0xFF values. I recall (not 100 percent certain) I was seeing the same when I regenerated the flash algo for lpc1768.

adamgreen:
I have reproduced the problem on the LPC1768 and I am working on a fix.  I hope to have something by tomorrow.

adamgreen:
I have the start of a fix for the LPC1768 issue over [here](https://github.com/mbedmicro/pyOCD/compare/master...adamgreen:lpc1768PageProgramFix?expand=1).  I will test more on Tuesday evening and then issue a PR.

0xc0170:
@adamgreen great, keep us reported

c1728p9:
@adamgreen Thanks for looking at this!  I was wondering, instead of removing page_size from all targets, would it be easier to override programPage and erasePage for the lpc1768 and make it look like the sector size is 32KB?  That would allow #104 to still work since it relies on all sectors being the same size.

0xc0170:
@c1728p9 how would 104 break from that change , adding a new method getPageSize(addr)  I don't fully understand "it relies on all sectors being the same size." from the above comment

c1728p9:
The pull request #104 uses a fixed page size when it reads out sector data to determine which ones need to be re-programmed.  It also only supports time estimates based on a fixed size sector erase and sector program.

adamgreen:
@c1728p9 I have made the requested changes.  I will finish testing later today and issue PR.

adamgreen:
I have issued PR #109 with the latest fixes for the LPC1768.

PR #107, opened 2015-03-10
Jtagonly

jeremybrodt:
1) Currently, pyOCD does not work for targets that only support JTAG (no SWD). This patch adds support for JTAG only devices.

2) If the JTAG tap is externally reset (e.g. target power cycle), pyOCD throws an exception and stops workings. This patch, adds handling of this exception and reconnects to the target.

### DIFF
--- a/pyOCD/target/cortex_m.py
+++ b/pyOCD/target/cortex_m.py
@@ -488,7 +488,10 @@ class CortexM(Target):
         write a memory location.
         By default the transfer size is a word
         """
-        self.transport.writeMem(addr, value, transfer_size)
+        try:
+            self.transport.writeMem(addr, value, transfer_size)
+        except:
+            self.handleMemoryError()
         return
 
     def write32(self, addr, value):
@@ -514,7 +517,11 @@ class CortexM(Target):
         read a memory location. By default, a word will
         be read
         """
-        return self.transport.readMem(addr, transfer_size)
+        try:
+            return self.transport.readMem(addr, transfer_size)
+        except:
+            self.handleMemoryError()
+            return 0
 
     def read32(self, addr):
         """
@@ -643,7 +650,10 @@ class CortexM(Target):
             n = self.auto_increment_page_size - (addr & (self.auto_increment_page_size - 1))
             if size*4 < n:
                 n = (size*4) & 0xfffffffc
-            self.transport.writeBlock32(addr, data[:n/4])
+            try:
+                self.transport.writeBlock32(addr, data[:n/4])
+            except:
+                self.handleMemoryError()
             data = data[n/4:]
             size -= n/4
             addr += n
@@ -659,7 +669,10 @@ class CortexM(Target):
             n = self.auto_increment_page_size - (addr & (self.auto_increment_page_size - 1))
             if size*4 < n:
                 n = (size*4) & 0xfffffffc
-            resp += self.transport.readBlock32(addr, n/4)
+            try:
+                resp += self.transport.readBlock32(addr, n/4)
+            except:
+                self.handleMemoryError()
             size -= n/4
             addr += n
         return resp
@@ -1155,3 +1168,7 @@ class CortexM(Target):
             return '0' + val
         else:
             return val
+
+    def handleMemoryError(self):
+            self.transport.reinit()
+            self.init()
--- a/pyOCD/transport/cmsis_dap.py
+++ b/pyOCD/transport/cmsis_dap.py
@@ -123,6 +123,22 @@ class CMSIS_DAP(Transport):
             self.writeDP(DP_REG['CTRL_STAT'], STICKYERR | STICKCMP | STICKORUN)
         return
 
+    def reinit(self):
+        # clear internal state
+        self.csw = -1
+        self.dp_select = -1
+        if (self.mode == DAP_MODE_SWD):
+            # switch from jtag to swd
+            JTAG2SWD(self.interface)
+            # clear abort err
+            dapWriteAbort(self.interface, 0x1e);
+        elif (self.mode == DAP_MODE_JTAG):
+            # Test logic reset, run test idle
+            dapSWJSequence(self.interface, [0x1F])
+            # clear errors
+            self.writeDP(DP_REG['CTRL_STAT'], STICKYERR | STICKCMP | STICKORUN)
+        return
+
     def uninit(self):
         dapDisconnect(self.interface)
         return
--- a/pyOCD/transport/cmsis_dap.py
+++ b/pyOCD/transport/cmsis_dap.py
@@ -65,12 +65,14 @@ TRANSFER_SIZE = {8: CSW_SIZE8,
                  32: CSW_SIZE32
                  }
 
+# Response values to DAP_Connect command
 DAP_MODE_SWD = 1
 DAP_MODE_JTAG = 2
 
-STICKORUN = 0x00000002
-STICKCMP = 0x00000010
-STICKYERR = 0x00000020
+# DP Control / Status Register bit definitions
+CTRLSTAT_STICKORUN = 0x00000002
+CTRLSTAT_STICKYCMP = 0x00000010
+CTRLSTAT_STICKYERR = 0x00000020
 
 def JTAG2SWD(interface):
     data = [0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff]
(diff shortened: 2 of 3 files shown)


pyOCD/target/cortex_m.py
```diff
@@ -488,7 +488,10 @@ def writeMemory(self, addr, value, transfer_size = 32):
         write a memory location.
         By default the transfer size is a word
         """
-        self.transport.writeMem(addr, value, transfer_size)
+        try:
+            self.transport.writeMem(addr, value, transfer_size)
+        except:
```
c1728p9: You might want to just catch TransferError exceptions to prevent more serious errors from getting masked.

pyOCD/target/cortex_m.py
```diff
@@ -488,7 +488,10 @@ def writeMemory(self, addr, value, transfer_size = 32):
...
+            self.handleMemoryError()
```
c1728p9: As is, the code calling writeMemory has no way to tell that the function failed.  It might be good to re-raise the exception, or raise a another.  That way the calling code would know the write failed, and that it should retry the write, or sequence of writes.

c1728p9:
Hi Jeremy, thanks for sharing your change.  I tried it on a K22F Freedom board, which supports SWD, and it worked without problems.  I was wondering, what board are you testing on that uses JTAG rather than SWD?

pyOCD/transport/cmsis_dap.py
```diff
@@ -65,6 +65,12 @@
                  32: CSW_SIZE32
                  }
 
+DAP_MODE_SWD = 1
+DAP_MODE_JTAG = 2
+
+STICKORUN = 0x00000002
```
0xc0170: Shall we add prefix, this is a group of errors. Something like above DAP_MODE_xxxx

jeremybrodt:
Hi @c1728p9. We are in the process of adding a platform based on the MAX32600 processor. It does not implement SWD. Thank you for testing this on the K22F board to confirm I did not break anything.

pyOCD/transport/cmsis_dap.py
jeremybrodt: I'm not sure I understand what you are recommending here @0xc0170. The first two values are response values to the DAP_Connect command specified in the DAP protocol. Can you make a small code snippet of what you think it should look like? Thanks!

pyOCD/target/cortex_m.py
jeremybrodt: It is actually the ValueError exception that is being thrown. The HDK responds that it did not complete any transactions (transaction count = '0'). Perhaps the error checking order in dapTransfer should be changed so that TransferError is thrown first. I will look into this today.

pyOCD/target/cortex_m.py
jeremybrodt: I agree. However, writeMemory is called from many different places. It may take some time to catch and properly handle all of these failures. If possible, I would prefer to implement this in a separate patch.

pyOCD/transport/cmsis_dap.py
0xc0170: Ah sorry, should have added a snippet. If those 3 are a set of errors for one register, if we can group them via prefix CTRL_ . we got something similar for CSW_ in the lines below. A comment above would be also helpful. I had to open the reference manual to find those values and their meanings

jeremybrodt: Ah, I see. Yes. I agree. I will add this.

pyOCD/target/cortex_m.py
c1728p9: Would it make sense to create a new exception to indicate a tap reset occurred?

pyOCD/target/cortex_m.py
c1728p9: When this exception occurs, does only the debug logic get reset and not the processor?  Maybe writeMemory could recover from the error by performing the write again.  That way the calling code could continue what is it doing, confident that the write succeed.

c1728p9:
Out of curiosity, what causes a tap reset to occur?  Is there any way to prevent this when debugging, or is this expected to occur?

jeremybrodt:
In general, it would occur anytime a target experienced a power cycle. If the HDK is powered separately from the target, this TAP reset condition would occur.

On the MAX32600, in addition to the traditional system reset, there is a power-on-reset input that resets everything on the chip, including the debug logic. If the user presses this button while debugging, the TAP is reset.

pyOCD/target/cortex_m.py
jeremybrodt: Perhaps. Confirming that a TAP reset occurred vs some other error would take further inspection. The DP CTRL/STAT register would need to be inspected. Where do you think it makes sense to do this?

pyOCD/target/cortex_m.py
c1728p9: Since the target will be reset when a tap reset occurs, maybe it would make more sense to re-throw the exception and catch it at an even higher level.  If a reset occurs while running most scripts, the script will likely fail even if individual write functions don't throw an exception.  Scripts that need to be able to recover from a tap reset could be written like below:

```
flashing_done = False
while not flashing_done:
    try:
        flash.flashBinary(binary_file)
        flashing_done = True
    except ValueError: # or maybe TapResetError
        Pass
```

pyOCD/target/cortex_m.py
c1728p9: You would want to check at most or every location that throws a ValueError.  That's probably outside the scope this patch, so it might be easiest to just catch ValueError (This will catch TransferErrors as well since it inherits from ValueError).

adamgreen:
This breaks the existing SWD handling of memory errors.  If I try to read past the end of valid RAM (this RAM block ends at 0x10007FFF) on my LPC1768 in GDB using a command like:
```
x/128bx 0x10007ff0
```
I get the following exception:
```
Exception in thread Thread-2:
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/gdbserver/gdbserver.py", line 152, in run
    [resp, ack, detach] = self.handleMsg(data)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/gdbserver/gdbserver.py", line 210, in handleMsg
    return self.getMemory(msg[2:]), 1, 0
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/gdbserver/gdbserver.py", line 442, in getMemory
    mem = self.target.readBlockMemoryUnaligned8(addr, length)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/cortex_m.py", line 585, in readBlockMemoryUnaligned8
    mem = self.readMemory(addr, 8)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/cortex_m.py", line 523, in readMemory
    self.handleMemoryError()
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/cortex_m.py", line 1174, in handleMemoryError
    self.init()
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/target/cortex_m.py", line 383, in init
    self.transport.writeDP(DP_REG['SELECT'], 0)
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/transport/cmsis_dap.py", line 245, in writeDP
    dapTransfer(self.interface, 1, [WRITE | DP_ACC | (addr & 0x0c)], [data])
  File "/usr/local/lib/python2.7/dist-packages/pyOCD/transport/cmsis_dap_core.py", line 215, in dapTransfer
    raise ValueError('Transfer not completed')
ValueError: Transfer not completed

ERROR:root:exception during unknit
```

The master branch works as expected.  This trace indicates that the memory error was caught in the CortexM layer and it incorrectly attempted to re-init where before it was caught in the GDBServer layer and an appropriate error sent to GDB.

jeremybrodt:
Thanks for the feedback @adamgreen I'll try this on my platform and see if I can find a better way to handle it.

c1728p9:
This sounds like a bug fixed recently by this patch: https://github.com/c1728p9/pyOCD/commit/790f32c6d628a2b2353a1fab2e4c46428be1df33

If you rebase the pull request do you still see a crash?

adamgreen:
@c1728p9 I tried rebasing to latest master, verified that this PR branch now included the 790f32c commit, and it still hit the same crash.

jeremybrodt:
I'm going to split this up and do a separate pull request for the JTAG only support and continue working on the exception handling in a different branch.

PR #108, opened 2015-03-16
New release package v0.3.1

0xc0170:
Merge this and release a new package version


c1728p9:
If you wait a day or so I can update #104 so it can be included in the new version as well assuming there is a quick fix for the lpc1768.

0xc0170:
We should release the new one with flash algo logic, incrementing minor number.

0xc0170:
@c1728p9 can you send a separate PR for lpc1768 if you manage to fix it

PR #109, opened 2015-03-17
LPC1768 has both 4k and 32k pages

adamgreen:
I updated the flashing code to handle the LPC1768 which has sectors of multiple sizes: 
* the first 64k of FLASH uses 4k pages.
* the rest of FLASH uses 32k pages.

This commit includes these updates:
* I updated both the Flash_lpc11u24 and Flash_lpc1768 classes to no
  longer have code in the programPage() method to override the
  page_size field.  This does cause the program_page flashing algorithm
  routine to be passed the current page size (4k or 32k) instead of the
  fixed 1k value.  This is ok as those functions ignore that parameter
  anyway since that is the cause of needing this programPage()
  workaround in the first plae.
* For the LPC1768, I made the pc_erase_sector address odd to be
  consistent with the rest of the addresses in this map. It will work
  either way but this is consistent with the rest.
* For the LPC1768, I moved begin_data up into the upper AHB RAM banks
  so that a whole 32k is available once program_page is able to take
  a whole page at once and not just a fixed 1k.
* For the LPC1768, I updated the memoryMapXML to return the correct
  FLASH block sizes, including the switch from 4k to 32k pages.
* For the LPC1768, I bumped the page size up to 32k and then override
  the erasePage() and programPage() methods to loop the necessary
  amount of times to properly erase and program 32k.

I ran the basic_test.py tests on my LPC1768, LPC11U24, KL25Z, NRF51822, and LPC4330 boards.  I also used gdb_server.py to upload a 90,472 byte image to my LPC1768 to include a mix of 4k and 32k pages.

### DIFF
--- a/pyOCD/flash/flash_lpc11u24.py
+++ b/pyOCD/flash/flash_lpc11u24.py
@@ -1,6 +1,6 @@
 """
  mbed CMSIS-DAP debugger
- Copyright (c) 2006-2013 ARM Limited
+ Copyright (c) 2006-2015 ARM Limited
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


c1728p9:
Thanks again for putting together this pull request.  I don't think I would have been able to fix this without a board.

adamgreen:
@c1728p9 No problem.  Thanks for the recent updates to pyOCD.

0xc0170:
Thanks @adamgreen

PR #110, opened 2015-03-18
Added support for JTAG only targets.

jeremybrodt:
Currently, pyOCD does not work for targets that only support JTAG (no SWD). This patch adds support for JTAG only devices.

### DIFF
--- a/pyOCD/transport/cmsis_dap.py
+++ b/pyOCD/transport/cmsis_dap.py
@@ -15,7 +15,7 @@
  limitations under the License.
 """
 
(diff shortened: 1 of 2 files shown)


adamgreen:
I tested existing SWD support on my LPC1768, LPC11U24, LPC4330, NRF51822, and KL25Z boards using basic_test.py.  I also connected GDB to my mbed-LPC1768 board using this version of pyOCD and didn't notice any regressions there either.  This PR allows accessing invalid memory locations from GDB without resulting in a pyOCD crash.

I don't have a board that allows exercising the new JTAG functionality so it is untested by me.

0xc0170:
Thanks @jeremybrodt for adding JTAG support.

PR #111, opened 2015-03-20
Remap flash to user code after reset on LPC

c1728p9:
On the LPC11u24 and LPC800 the vector table is mapped to the boot
loader flash on reset rather than user flash.  This patch remaps the
flash back to user code and sets the SP and PC when resetStopOnReset
is called.


c1728p9:
Moving to #104

PR #118, opened 2015-04-10
Python 3 fix: encode/decode bytes/string as needed.

dbrobins:


c1728p9:
Thanks for taking the time to fix this on python 3.  I gave it a try on python 2, but it didn't work for me (just hung, didn't crash).  Do the decode calls fix python 3 by converting the data from 'str' to 'unicode'?  Python 2 is expecting type 'str' and the 'unicode' type seems to cause problems.

0xc0170:
I am closing this PR for the moment, as @dbrobins pointed out, this can be used as a guidance, this should be fixed. Can you create an issue describing all issues you had when using this module with python 3? Thanks

PR #121, opened 2015-04-16
Fix hidapi back end

c1728p9:
In HidApiUSB call the parent class's constructor so all the required
fields are present.  In particular the addition of the member
'packet_count'  fixes crashes when running hidapi.

The HID API interface was broken by the patch:
d2b5c06d4cde08c2ed8a1121c055d3e6e56fb756 -
"Improve dapTransfer speed by queuing packets"


autopulated:
works for me now, thanks! :+1:

PR #123, opened 2015-04-21
Refactoring for PyUSB backend

Pithikos:
I did some refactoring while playing with PyOCD. Most of the refactoring is just making the code more pythonic (removing some redundant things).

Some problematic value-checking is fixed as well. Namely in some cases `is` is used for checking of values. This might work now but since `is` does identity checking rather than value checking, some bugs might slip into the code.

### DIFF
--- a/pyOCD/interface/pyusb_backend.py
+++ b/pyOCD/interface/pyusb_backend.py
@@ -88,9 +88,9 @@ class PyUSB(Interface):
             
             # iterate on all interfaces:
             #    - if we found a HID interface -> CMSIS-DAP
-            for interface in config:
-                if interface.bInterfaceClass == 0x03:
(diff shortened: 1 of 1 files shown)


0xc0170:
Can you provide better commit messages ? If you can, squash those 2 into one and provide there a descriptive message , like you did above? 

For example, without looking at how endpoints are implemented for interface, we could get the change from the commit message why those 10 lines were changed to : 

```
ep_in, ep_out = interface.endpoints()
```

Thanks

Pithikos:
Ok, I made a new pull request [here](https://github.com/mbedmicro/pyOCD/pull/124)

0xc0170:
Thanks, closing this one. 
You could use rebase or soft reset, then force push, this PR would get updated.

Issue #113, opened 2015-03-22
Test binary for each development board rather than each processor


c1728p9:
Currently when pyOCD finishes basic_test.py it will program a new image to flash.  Right now this image is determined by the processor type.  The problem I'm concerned with is that several development kits can have the same processor, but have pins routed to different hardware.  Because of this, an image that caused an led to blink on one device could sink current into a button an another. Would it be a change for the better to modify pyOCD to use a test binary for each development board rather than each processor?

0xc0170:
board id to target defines targets (which is the board), kl25z is frdm-kl25z.
```
BOARD_ID_TO_TARGET = {
                "0200": "kl25z",
```

And the binary is chosen based on the target (each target has own specific number).

What do you propose? To update targets name to reflect the real names, kl25z -> frdm-kl25z ?

c1728p9:
Right now several development boards used the same binary:
```
"1070" -> "nrf51822" -> l1_nrf51822.bin
"9009" -> "nrf51822" -> l1_nrf51822.bin
"9012" -> "nrf51822" -> l1_nrf51822.bin
```

It might make more sense to be structured like this:
```
"1070" -> "Nordic_nRF51-Dongle.bin"
"9009" -> "Nordic_nRF51-DK.bin"
"9012" -> "Seeed_Tiny_BLE.bin"
```

That way each board will have dedicated firmware for its hardware.  Even with this change the target and flash names would stay the same.  There would just be a dedicated binary for each board.

c1728p9:
Done in #129

PR #129, opened 2015-04-24
Map boards to board info rather than to target

c1728p9:
Map each board ID to a board info object rather than only to a target.
This make it possible to have a different test binary for each board.
Dedicated test binaries are important on boards that use the same
processor, but have buttons, lights or other hardware on different
pins.

Let me know what you think of this patch.  It's not finished yet, but I figured I would push it up so I could get some feedback.

### DIFF
--- a/pyOCD/board/mbed_board.py
+++ b/pyOCD/board/mbed_board.py
@@ -22,28 +22,35 @@ from time import sleep
 from board import Board
 from pyOCD.interface import INTERFACE, usb_backend
 
-BOARD_ID_TO_TARGET = {
-                "0200": "kl25z",
-                "0210": "kl05z",
-                "0220": "kl46z",
-                "0230": "k20d50m",
-                "0231": "k22f",
-                "0240": "k64f",
-                "0250": "kl02z",
-                "0260": "kl26z",
-                "1010": "lpc1768",
-                "9004": "lpc1768",
-                "1040": "lpc11u24",
-                "1050": "lpc800",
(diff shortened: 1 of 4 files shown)


0xc0170:
:+1: , we will need to add the names of boards, can you create an issue for tracking this.

c1728p9:
I have #113 open for this

c1728p9:
@adamgreen 
What should the names for the board IDs 1600 and 1605 be?

@jeremybrodt 
What should the names for board IDs 0400 and 0405 be?  Same as the target - maxwsnenv and max32600mbed?

adamgreen:
* 1600 is Bambino 210
* 1604 is Bambino 210E

c1728p9:
Updated and ran automated test:
```
------ TEST RESULTS ------
Target         Test           Result

kl46z          Basic Test     Pass
kl46z          Speed Test     Pass
kl46z          Flash Test     Pass
k22f           Basic Test     Pass
k22f           Speed Test     Pass
k22f           Flash Test     Pass
lpc1768        Basic Test     Pass
lpc1768        Speed Test     Pass
lpc1768        Flash Test     Pass
k64f           Basic Test     Pass
k64f           Speed Test     Pass
k64f           Flash Test     Pass
lpc11u24       Basic Test     Pass
lpc11u24       Speed Test     Pass
lpc11u24       Flash Test     Pass

All tests passed
```

0xc0170:
Thanks, merging

PR #131, opened 2015-04-24
Added MAXWSNENV and MAX32600MBED to flash_test.py

jeremybrodt:
Added MAXWSNENV and MAX32600MBED to flash_test.py

### DIFF
--- a/test/flash_test.py
+++ b/test/flash_test.py
@@ -116,6 +116,16 @@ def flash_test(board_id):
             rom_size = 0x40000
             # Override clock since 10MHz is too fast
             test_clock = 1000000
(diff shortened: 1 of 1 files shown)


c1728p9:
You might also want to add these boards to the remaining tests - basic_test.py and speed_test.py.  If you do that, then you can use automated_test.py to verify all tests work and get performance metrics.

jeremybrodt:
They are already supported in those tests. I just some how overlooked the flash one.

c1728p9:
Excellent, I see that now.  The script flash_test.py was brought in in #104, so it probably didn't exist when your brought your targets in.

PR #135, opened 2015-05-17
New device support and command line script.

flit:
Added a new `pyocd.py` script that lets you perform low level operations like read and write memory, reset, and so on from the command line. This is an early version and not all features work, and a lot of useful features are missing. The command line interface is likely to change. `pyocd.py` supports printing the disassembly of data you read using the Capstone library. If Capstone is not present, you'll just get a warning if you try to use the --disasm option.

Install Capstone via `pip install capstone`.

Also added new pre-launch KL28Z device support. And some refactoring and minor improvements.


c1728p9:
Thanks for the patch!  Do I sense a a dual core Kinetis L device about to be released soon? :smile:

pyOCD/target/target_kinetis.py
```diff
@@ -141,3 +164,20 @@ def massErase(self):
...
+    #
+    # This will prevent the device from being secured after the next reset.
+    def unlockFlash(self):
+        logging.info("Unlocking chip...")
+
+        # Init flash algorithm.
+        self.flash.init()
+
+        logging.info("Performing mass erase")
+        self.massErase()
```
c1728p9: Mass erase will take care of writing 0xFE to the FCF to keep the part unlocked, so the other flash operations init and pageProgram shouldn't be needed.

pyOCD/flash/__init__.py
```diff
@@ -32,10 +33,12 @@
 
 FLASH = {
          'cortex_m' : Flash_cortex_m,
+         'kinetis' : Flash_cortex_m,
```
c1728p9: What is the purpose of mapping 'kinetis' to Flash_cortex_m?

flit: To have a generic Kinetis device, similar to the generic Cortex-M. Since Kinetis devices have different and incompatible flash configurations (and therefore drivers), we have to disable flash support for it.

pyOCD/target/target_kinetis.py
flit: Mass erase actually leaves the FCF blank. Internally, mass erase (however triggered) uses the Erase All Blocks flash command.

From the FTFA chapter's description of the Erase All Blocks command:
If the flash memory module verifies that all flash memories were properly erased, access
control is disabled and security is released by setting the FSEC[SEC] field _[this means the FSEC register in FTFA]_ to the unsecure state. … The security byte and all other contents of the flash configuration field (see Flash Configuration Field Description) are erased by the Erase All Blocks command.

So security is only temporarily disabled until the next reset.

On the KL43 and follow-on L-series devices we added a new Erase All Unsecure flash command that _does_ program the FSEC field of FCF after erasing everything. But this command is not used by mass erase triggered from the MDM-AP.

c1728p9: In AN4835 - "Production Flash Programming Best Practices for Kinetis K- and L-series MCUs" in 4.2.1 - "Mass erase," it states that as one of the steps of mass erase is, "5. The security byte in the flash configuration field is also programmed to 0xFE (unsecure). On subsequent resets, the processor will boot in unsecure mode unless the flash configuration field is modified."

This is also consistent with the behavior I have seen.  On flash init setTargetState is called, which calls resetStopOnReset.  This would re-lock the processor if mass erase didn't write to the FSEC register.  I verified that Kinetis devices can be reliably unlocked with the script blank_test.py.

pyOCD/flash/flash_kl28z.py
```diff
@@ -0,0 +1,87 @@
...
+    0xd20042ab, 0x5d63461d, 0x18e3005b, 0x4718bc30, 0x00040002, 0x00080000, 0x00100000, 0x00200000,
+    0x00400000, 0x00000080, 0x00000000, 0x00800000, 0x40020004, 0x00000000,
+                                ],
+               'pc_init' : 0x20000021,
+               'pc_eraseAll' : 0x2000004D,
+               'pc_erase_sector' : 0x20000071,
+               'pc_program_page' : 0x2000009F,
+               'begin_stack' : 0x20000800,
+               'begin_data' : 0x20000a00,       # Analyzer uses a max of 1 KB data (256 pages * 4 bytes / page)
+                                                # Note: 128 pages on KL25 and KL26, 256 pages on KL46
```
c1728p9: Old comments.  This should read:
# Analyzer uses a max of 2 KB data (512 pages * 4 bytes / page)

tools/pyocd.py
```diff
@@ -0,0 +1,276 @@
...
+                target.unlockFlash()
+            elif args.action == ACTION_GO:
+                target.resume()
+
+        except ToolError, e:
+            print "Error:", e
+            exitCode = 1
+        finally:
+            if board != None:
+                board.uninit()
```
c1728p9: By default board.uninit() calls target.resume().  Because of this --halt does not work as expected.  You could fix this by adding changing the line to board.uninit(False)

c1728p9:
Good start on the new pyocd.py tool!  I did a more thorough review of your code so this should be my last set of comments.

What progress bar problem do you see when flush is missing?  It might also need to be fixed in flash_tool.py line ~77 and gdbserver.py line ~368.

I noticed you moved flash_tool.py and flash_nrf51822.py to the tools directory.  It might also be good to move gdb_server.py to this directory since it is a useful tool as well.

Finally, I ran automated_test.py on your patch and all tests passed.
Targets tested: kl46z, k22f, lpc1768, k64f, lpc800, lpc11u24

tools/pyocd.py
flit: Thanks, I was wondering how to fix that!

pyOCD/flash/flash_kl28z.py
flit: Ok, thanks.

flit:
On OS X at least, the progress bar was not being updated for every change, only when the stdout buffer was flushed.

I agree that gdb_server.py should probably be moved to tools/. I can go ahead and make that change.

And thanks for testing!

0xc0170:
Great work guys !

c1728p9:
Since a missing flush causes problems you might also want to add it to flash_tool.py line ~77 and gdbserver.py line ~368.

0xc0170:
@flit is this feature complete?

flit:
Yes, for now. There will be a lot of stuff to add to pyocd.py in the future, but the basic read/write works and is a good starting point.

0xc0170:
@flit Can you rebase this PR? There are multiple conflicts as I merged a minute ago one pull request

flit:
Will do.

PR #140, opened 2015-05-22
New device support and command line script. (Rebased)

flit:
Added a new pyocd.py script that lets you perform low level operations like read and write memory, reset, and so on from the command line. This is an early version and not all features work, and a lot of useful features are missing. The command line interface is likely to change. pyocd.py supports printing the disassembly of data you read using the Capstone library. If Capstone is not present, you'll just get a warning if you try to use the --disasm option.

Install Capstone via `pip install capstone`.

Also added new pre-launch KL28Z device support. And some refactoring and minor improvements.

This branch is a rebased version of #135.


0xc0170:
thanks

pyOCD/target/target_kinetis.py
```diff
@@ -141,3 +164,18 @@ def massErase(self):
...
+    def unlockFlash(self):
+        logging.info("Unlocking chip...")
+
+        logging.info("Performing mass erase")
+        self.massErase()
+
+        # Write FCF
+        logging.info("Writing FCF = %s" % repr(fcf))
+        self.flash.init()
+        self.flash.programPage(FCF_ADDR, fcf)
```
c1728p9: Do you plan on keeping this?  According to AN4835, a mass erase triggered by the MDM-AP will program the FSEC to unsecure (0xFE), so this is unnecessary.  Just wanted to make sure you got my last comment on #135.

flit: Oh, right. I should remove this, thanks for the reminder. I did speak with the NVM team and they confirmed that you were right. ;) I was getting mass erase triggered by MDM-AP confused with the Erase All Blocks command. The Erase All Unsecure command we added was so that were is a command that works exactly like the MDM-AP mass erase.

PR #141, opened 2015-05-25
GDB server improvements

flit:
Added 'p' and 'M' commands to GDB server. This enables using the server with IAR's GDB server debug plugin.

Fixed a bug in the 'm' (getMemory) command handler in cases where the length was a single hex digit.

Includes a performance increase when using HID API as the USB backend, by enabling multiple packets.

Also includes some minor refactoring and gdb server logging enhancements.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -18,7 +18,7 @@
 import logging, threading, socket
 from ..target.target import TARGET_HALTED, WATCHPOINT_READ, WATCHPOINT_WRITE, WATCHPOINT_READ_WRITE
 from ..transport import TransferError
-from ..utility.conversion import hexStringToBinary
+from ..utility.conversion import hexStringToIntList, hexEncode, hexDecode
 from struct import unpack
 from time import sleep, time
 import sys
@@ -454,7 +454,7 @@ class GDBServer(threading.Thread):
         length = int(split[0], 16)
 
         split = split[1].split('#')
-        data = hexStringToBinary(split[0])
+        data = hexStringToIntList(split[0])
 
         try:
             if length > 0:
@@ -562,7 +562,7 @@ class GDBServer(threading.Thread):
             return self.createRSPPacket(resp)
 
         elif query[0].startswith('Rcmd,'):
-            cmd = self.hexDecode(query[0][5:].split('#')[0])
+            cmd = hexDecode(query[0][5:].split('#')[0])
             logging.debug('Remote command: %s', cmd)
 
             safecmd = {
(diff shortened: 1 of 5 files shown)


pyOCD/gdbserver/gdbserver.py
```diff
@@ -436,12 +456,38 @@ def getMemory(self, data):
...
+
+        split = split[1].split('#')
+        data = hexStringToIntList(split[0])
+
+        if LOG_MEM:
+            logging.debug("GDB writeMemHex: addr=%x len=%x", addr, length)
+
+        try:
+            if length > 0:
+                self.target.writeBlockMemoryUnaligned8(addr, data)
```
c1728p9: With deferred transfers enabled you'll also want to call flush to make sure data gets written out immediately (and any exceptions get thrown while you are still in the try block)

pyOCD/interface/hidapi_backend.py
```diff
@@ -107,3 +107,7 @@ def close(self):
         """
         logging.debug("closing interface")
         self.device.close()
+
+    def setPacketCount(self, count):
+        # No interface level restrictions on count
+        self.packet_count = count
```
c1728p9: This probably won't increase speed much with just this.  As is, the hidapi backend only reads or writes, but doesn't do both at the same time.  If a reader thread was created then this would likely double speed.  An easy way to check the effect on performance is to run speed_test.py.

c1728p9:
Thanks for the patch.  I somehow overlooked this pull request so sorry for the late comments.

pyOCD/interface/hidapi_backend.py
flit: Actually, it had a pretty big impact, not quite doubling the speed for actual memory access.
<pre>
Target      Write Speed     Read Speed      USB speed       USB overlap speed

without     24.929752 KB/s  25.037943 KB/s  31.970114 KB/s  31.939022 KB/s  
with        44.660883 KB/s  45.513174 KB/s  31.969902 KB/s  63.973134 KB/s  
</pre>

This is on OS X. The hidapi implementation on the Mac uses a reader thread internally. For Windows it may be very different.

pyOCD/gdbserver/gdbserver.py
flit: Ok, thanks for pointing this out. I'll fix it shortly.

pyOCD/interface/hidapi_backend.py
c1728p9: Ah that explains it.  I tried it on windows and it gives similar performance improvements.

0xc0170: We are getting better numbers day by day :-) Awesome work guys !

PR #146, opened 2015-05-28
Double buffering for flash programming

flit:
Improve flash programming performance by using double buffering. Data for the next page is loaded into RAM while the previous page is still being programmed. Double buffering is enabled by adding the 'page_buffers' value to the `flash_algo` dictionary.

Only Freescale Kinetis devices are currently supported. I need someone to help enable and test on other devices.

On the K64F, these are the flash_test.py results before and after:

<pre>
Target           Chip Erase          Page Erase          Page Erase (Same data)

k64f single      14.796891 KB/s      14.272724 KB/s      35.775076 KB/s      
k64f double      24.274857 KB/s      22.992638 KB/s      35.759061 KB/s      
</pre>


test/speed_test.py
```diff
@@ -234,5 +237,13 @@ def speed_test(board_id):
...
-    speed_test(None)
+    parser = argparse.ArgumentParser(description='pyOCD speed test')
+    parser.add_argument('-d', '--debug', action="store_true", help='Enable debug logging')
+    args = parser.parse_args()
+    level = logging.DEBUG if args.debug else logging.INFO
+    logging.basicConfig(level=level)
+    board = pyOCD.board.mbed_board.MbedBoard.getAllConnectedBoards(close = True)[0]
+    test = SpeedTest()
+    result = [test.run(board)]
+    print result
```
c1728p9: Why is result printed?

pyOCD/flash/flash_builder.py
```diff
@@ -422,3 +504,106 @@ def _page_erase_program(self, progress_cb = _stub_progress):
...
+                # Wait for the program to complete.
+                result = self.flash.waitForCompletion()
+
+                # check the return code
+                if result != 0:
+                    logging.error('programPage(0x%x) error: %i', current_addr, result)
+                    error_count += 1
+                    if error_count > self.max_errors:
+                        logging.error("Too many page programming errors, aborting program operation")
+                        break
```
c1728p9: Were you seeing errors when programming?  Seem like you might want to abort after even 1 error

c1728p9:
Excellent idea!  Patch looks great.  I'll test with the targets I have but I might not be able to get to it  until the weekend.

test/speed_test.py
0xc0170: From  the commit message for this change: 

Also added printing of exception tracebacks for any exceptions that happen while running the tests.

flit: Ooops, that was a debug print that was accidentally committed. I'll remove it. Thanks!

pyOCD/flash/flash_builder.py
flit: I only saw errors when I did something wrong while working on the code. :) I figured it wouldn't hurt to be a little forgiving. Though we should probably be returning a status code to the caller.

flit:
Thanks for reviewing. If this pull request is not merged in by the time you are able to test other targets, you can send me a patch for additional device support and I'll push it to this branch for inclusion in the pull request. Or we can just add other devices as separate pull requests.

pyOCD/flash/flash_builder.py
```diff
@@ -422,3 +504,106 @@ def _page_erase_program(self, progress_cb = _stub_progress):
...
+                data = self.flash.target.readBlockMemoryUnaligned8(page.addr, len(page.data))
+                page.same = _same(page.data, data)
+                progress += page.getVerifyWeight()
+                count += 1
+                if page.same:
+                    same_count += 1
+
+                # Update progress
+                progress_cb(float(progress) / float(self.page_erase_weight))
+        logging.debug("scanned %d pages for equality (%d same)", count, same_count)
```
c1728p9: If debug logging is on and only a few pages needed to be reprogrammed this will print after the progress bar making the progress bar appear twice:

```
DEBUG:root:Chip erase weight 37.320000, Page erase weight 15.065200
DEBUG:root:Using double buffer page erase program
[================    ]  83%DEBUG:root:scanned 245 pages for equality (245 same)
[====================] 100%
DEBUG:root:Estimated page erase count: 11
DEBUG:root:Actual page erase count: 11
```

This is pretty minor so it might not be worth fixing, but I figured I would mention it.  The estimated vs real page same count is printed later, so this statement might not be needed.

c1728p9:
No problem.  Thanks for the patch, I have been using it today and it is excellent :smiley:.  If you want to merge this patch feel free to.  I'll open up a new pull request as you suggested if this gets merged before my changes.

PR #147, opened 2015-05-30
fixed read/write BlockMemoryUnaligned8 functions

molejar:
Added while loop for processing all data in list. It has impact to Flash > programPage function. For me its working now correct.

### DIFF
--- a/pyOCD/target/cortex_m.py
+++ b/pyOCD/target/cortex_m.py
@@ -540,45 +540,46 @@ class CortexM(Target):
         """
         res = []
 
-        # try to read 8bits data
-        if (size > 0) and (addr & 0x01):
-            mem = self.readMemory(addr, 8)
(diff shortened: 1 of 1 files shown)


c1728p9:
Hi @molejar , thanks for the pull request.  What problem were you seeing before you created this pull request?  Were not enough bytes getting written/read?  The way writeBlockMemoryUnaligned8 and readBlockMemoryUnaligned8 are written there should be no need for a loop.  In those functions the first and last 2 checks transfer all the unaligned data and the check in the middle transfers all the other data.

Were you using python 3 when you saw problems?  If so it might have something to do with changes to integer division in python 3 which handles rounding differently when using '/'.

molejar:
Hi @c1728p9, sorry for wrong pull request. You have right, my changes has no impact to functionality. Please, ignore it. Thanks.

PR #148, opened 2015-05-31
GDBServer no ack mode support.

flit:
Small change to support no ack mode in the gdb server. Since we only support TCP/IP, there's really no benefit to the acks.

This change also saved the set of supported features sent by gdb for later use.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -57,6 +57,9 @@ class GDBServer(threading.Thread):
         self.hide_programming_progress = options.get('hide_programming_progress', False)
         self.fast_program = options.get('fast_program', False)
         self.packet_size = 2048
+        self.send_acks = True
+        self.clear_send_acks = False
+        self.gdb_features = []
         self.flashBuilder = None
         self.conn = None
         self.lock = threading.Lock()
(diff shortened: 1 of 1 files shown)


c1728p9:
Thanks for the patch.  I agree, the acks shouldn't be necessary.  There might also be a slight speed increase.

I gave this a try, but I can't seem to halt the processor anymore from GDB (ctrl-c doesn't work).  If I revert the patch then I can halt correctly.

flit:
Thanks for testing. I'll try to reproduce the ctrl-c problem tonight.

0xc0170:
:+1:

adamgreen:
I got blocked by the CTRL+C issue while doing some development today.  The following diff contains the changes I made to unblock myself but are not thoroughly tested.  Summary of fixes:
* Actually stop sending ACKs once GDB has made that request.  GDB was getting confused by the ACK after the continue command was sent so it never bothered to send the 0x03 (CTRL+C) byte to pyOCD.
* Fix bugs in code that are then exposed once ACK transmission is turned off.

```diff
diff --git a/pyOCD/gdbserver/gdbserver.py b/pyOCD/gdbserver/gdbserver.py
index dc943a3..12d7b91 100644
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -162,6 +162,10 @@ class GDBServer(threading.Thread):
                 if len(data) != 0:
                     # decode and prepare resp
                     [resp, ack, detach] = self.handleMsg(data)
+
+                    # Prepare to parse next command which can exist if not in ack mode
+                    if not self.send_acks:
+                        data = data[data.index("#"):]
             
                     if resp is not None:
                         # ack
@@ -680,7 +684,7 @@ class GDBServer(threading.Thread):
 
         if feature == 'StartNoAckMode':
             # Disable acks after the reply and ack.
-            self.clear_send_acks = False
+            self.clear_send_acks = True
             return self.createRSPPacket("OK")
         else:
             return self.createRSPPacket("")
@@ -733,6 +737,7 @@ class GDBServer(threading.Thread):
         return resp
     
     def ack(self):
-        self.abstract_socket.write("+")
+        if self.send_acks:
+            self.abstract_socket.write("+")
```

adamgreen:
I see that @c1728p9 has already issued a PR with almost the exact same patches.  I was so heads down trying to get unblocked I didn't realize that there was already a PR on GitHub for it :)

PR #149, opened 2015-06-01
Simplify Installation

posborne:
These changes push the burden of determining and installing package dependencies into the Python packaging system where those concerns belong.  Since these concerns are now handled automatically, you should be able to run the following commands on any system to get a GDB server up-and-running that is attached to hardware...

```Shell
$ sudo pip install --pre -U https://github.com/posborne/pyOCD/archive/simplify-installation.zip
$ pyocd
...
```

I have been able to test on Linux (Ubuntu 14.04) and Windows and this approach works well.  I believe most users will already have pip installed.  If they do not, they should.

When the next release is pushed to pypi, the first step will become even easier...

```Shell
$ pip install pyocd
```


posborne:
I tested on mac as well now.  I had to install `Cython` (easy enough via pip).  Removing this should be possible but only if the build for hidapi was changed to include the Cython-generated C files in the distribution packages.

setup.py
```diff
@@ -24,11 +39,18 @@
...
     license="Apache 2.0",
-    classifiers = [
+    classifiers=[
         "Development Status :: 4 - Beta",
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
     ],
+    entry_points={
+        'console_scripts': [
+            'pyocd = pyOCD.gdbserver.runner:main',
```
c1728p9: Excellent idea!  It might also be useful to add flash_tool.py here as well

README.md
```diff
@@ -25,49 +25,40 @@ You can use the following interfaces:
...
-$ git clone https://github.com/trezor/cython-hidapi.git
-$ cd cython-hidapi
-$ sudo python setup.py install
-```
-
-### Install pyOCD
-Clone pyOCD somewhere then run the install script.
-```Shell
-$ cd /path-to-pyOCD/
-$ python setup.py install
```
c1728p9: It might not hurt to keep some instructions for installing from the git rep

c1728p9:
Thanks for the patch.  I tried it out on my windows machine and didn't have any problems installing.

setup.py
posborne: Will do; are there any other scripts worth adding?

README.md
posborne: Sure.  I can do that.

posborne:
Rebased and reworked with the following changes:

1. There now are entry points defined for all 3 of the tools that were present before.  All are prefixed with `pyocd`.  I am not too familiar with how the `pyocd` tool was used, so please provide suggest better names if they might exist.  I prefixed them all with `pyocd` in order to prevent namespace confllicts/pollution as most users will install pyocd globally.
2. Updated documentation and tested pyinstaller.  See https://github.com/posborne/pyOCD/commit/e38cf5ad07dcc505768db5793a0cfa713bdce8ed for more details.
3. Other documentation updates including some additional guidance encourage developers to use virtualenv in keeping with modern python best practices.

0xc0170:
Thanks @posborne, well documented. This should simplify usage. We can start using pyocd-gdbserver :+1:

PR #152, opened 2015-06-06
Fix halting target in gdb_server.py

c1728p9:
The gdb server reports it supports "no ack" mode, but still sends acks
after switching modes.  This causes the gdb client to hang when sending
the halt command (ctrl-c).

This patch prevents gdb server from sending acks after it has switched
modes.  It also explicitly clears out the data buffer after calling
handleMsg, since this was done implicitly by the code responsible for
reading acks.

The problems halting the target were introduced in the commit:
29e552501686908fc6c05abdf98c9fadf0446b76 -
"GDBServer no ack mode support."

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -162,6 +162,9 @@ class GDBServer(threading.Thread):
                 if len(data) != 0:
                     # decode and prepare resp
                     [resp, ack, detach] = self.handleMsg(data)
+
+                    # Clear out data
+                    data = ""
             
                     if resp is not None:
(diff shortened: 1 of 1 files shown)


c1728p9:
@flit ,I figured you were probably busy with FTF, so I decided to track down this problem.  This fixes the strange behavior I was seeing.

I also verified that with this patch neither the gdb server, nor gdb client sent acknowledge messages after the initial communication.

flit:
Thanks for fixing this for me! Unfortunately, I was busy all week and didn't get a chance to look at it before you.

adamgreen:
:+1:

PR #154, opened 2015-06-08
Prevent uninit timeout in pyusb_backend

geky:
The PyUSB backend in a virtual machine can't seem to keep up with 
multiple packets being sent at the same time. Even with no timeout, 
the speed_test script hangs as soon as the overlapped test starts.

There may be a more efficient solution than just limiting the backend 
to a single packet, but this at least prevents PyUSB from crashing.
Should PyUSB be able to handle multiple packets on slow connections?

Tested on the K64F and nRF51822 in Ubuntu 14.04.

### DIFF
--- a/pyOCD/interface/pyusb_backend.py
+++ b/pyOCD/interface/pyusb_backend.py
@@ -165,10 +165,6 @@ class PyUSB(Interface):
             pass
         return self.rcv_data.pop(0)
 
-    def setPacketCount(self, count):
-        # No interface level restrictions on count
-        self.packet_count = count
-
     def close(self):
         """
         close the interface


c1728p9:
Hi @geky , can you provide any more details on your setup, or the problem you are experiencing?  Is your virtual machine passing through USB traffic to your PC's hardware?  Could the root cause of this be a virtual machine driver problem?  Also, do you know how many overlapped packets get sent before pyOCD gets stuck?  PyUSB should be able to handle this even on slow connections.

geky:
Ah yes, It's an instance of the [Oracle VirtualBox](https://www.virtualbox.org/) and USB is being passed through their VirtualBox USB driver to the Windows 7 host machine.

It does send roughly 900 packets before timing out. It wouldn't be too surprising if the problem is just a flaw in the virtual machine's driver.

Alternatively, do you think it would be reasonable to add a packet-limiting flag to PyOCD?

posborne:
I'm pretty sure there is definitely some bugs in the VirtualBox USB drivers.  For me, when I run pyOCD in a VM and attempt to perform a flash, my entire VM crashes/freezes.  Could be https://www.virtualbox.org/ticket/9786.

geky:
Ack, yep, looks like I'm most likely running into this issue: https://www.virtualbox.org/ticket/13579.

I guess it's not really worth working around then, thanks for looking into it.

geky:
I put together a quick fork with a packet_limit flag in case anyone else runs into this issue and needs a workaround: [/geky/pyOCD/tree/packet_limit](/geky/pyOCD/tree/packet_limit)

PR #155, opened 2015-06-09
Modernize setup

mindw:
- add ``__version__`` to package.
- use setuptools-scm to auto generate version from git tag
- add install dependency on intelhex
- fill in some missing metadata


setup.py
```diff
@@ -14,31 +14,30 @@
...
 setup(
     name="pyOCD",
-    version="0.4.2",
+    use_scm_version={
+        'local_scheme': 'dirty-tag',
+        'write_to': 'pyOCD/_version.py'
+    },
+    setup_requires=['setuptools_scm'],
     description="CMSIS-DAP debugger for Python",
+    long_description=open('README.md', 'Ur').read(),
```
posborne: `long_description` actually needs to be restructured text.  I have done something like this on other projects in order to work around this: https://github.com/digidotcom/python-wvalib/blob/master/setup.py#L27

pyOCD/tools/pyocd.py
```diff
@@ -22,8 +22,9 @@
 
 import os
 import pyOCD
-from pyOCD.board import MbedBoard
-from pyOCD.target import target_kinetis
+from ..board import MbedBoard
+from ..target import target_kinetis
+from .. import __version__
```
posborne: You should just use absolute imports here.  The base namespace of `pyOCD` is going to be more stable than the relative position of this file to the base namespace.  See https://www.python.org/dev/peps/pep-0008/#imports

0xc0170: :+1:

setup.py
0xc0170: Thanks for sharing, I see that issue to support md is still open for pypi :/

mindw: I've converted the README to restructured text.

pyOCD/tools/pyocd.py
mindw: I've reverted the change and tried my best to fix import according to pep8 with absolute imports.

I failed to find the relevant entry in the pep which support the use of absolute imports in the case above.

0xc0170:
LGTM

@posborne Updated, what about now?

posborne:
Overall, looks good.  The translation to RST looks like it has some problems.  You can view it rendered here: https://github.com/mindw/pyOCD/tree/modernize_setup.  Is there a major problem with converting from markdown just-in-time?  E.g. https://github.com/posborne/cmsis-svd/blob/master/python/setup.py#L34

This requires pandoc, but this should be easy to satisfy so long as you are building the sdist from a linux box.  In general, I think more people are familiar with markdown than restructured text.

posborne:
There is also  this: https://pypi.python.org/pypi/setuptools-markdown/0.2 which works similarly.  I did not use that most recently as it chokes if pandoc is not present rather than failing gracefully.  I suppose that should be an issue against that project.

posborne:
Alright, issue filed against setuptools-markdown: https://github.com/msabramo/setuptools-markdown/issues/6

mindw:
Yes, I've used both pypandoc and looked at setuptools-markdown. The output you have now is after using pandoc manually and fixing the output (rst2html used for preview). It looks like github output is uses a different engine different (https://github.com/github/markup). pypandoc or pandoc cannot be expected to be available on Windows machines (where I do most of my development).
I'll try to fix the text issues.

mindw:
I hope version is more acceptable.

posborne:
:+1:

c1728p9:
When I run gdb_server.py with this patch I get the below error.  If I revert this pull request everthing works correctly.

```
  File "E:\Projects\mbed\pyOCD\pyOCD\pyOCD\tools\gdb_server.py"
, line 23, in <module>
    from pyOCD import __version__
  File "build\bdist.win32\egg\pyOCD\__init__.py", line 26, in <module>

ImportError: No module named _version
```

posborne:
@c1728p9, are you running in a virtualenv.  I believe the file should be generated when you install dev-requirements.txt.

```
$ pip install -r dev-requirements.txt
```

This should trigger `setuptools_scm` in setup.py to create the missing file.

mindw:
Please build or install pyOCD prior to usage - any of the following will
generate _version.py:
1. Install it: pip install.
2. Build it: python setup.py build.

-g

-gabi

0xc0170:
@mindw The build command generates an error:

```
writing entry points to pyOCD.egg-info\entry_points.txt
File Finder Failed for .git = setuptools_scm.git:FILES_COMMAND
  File "build\bdist.win-amd64\egg\setuptools_scm\integration.py", line 31, in find_files
    return do(ep.load(), path).splitlines()
  File "build\bdist.win-amd64\egg\setuptools_scm\utils.py", line 57, in do
    out, err, ret = do_ex(cmd, cwd)
  File "build\bdist.win-amd64\egg\setuptools_scm\utils.py", line 42, in do_ex
    HGPLAIN='1',
  File "C:\Python27\lib\subprocess.py", line 710, in __init__
    errread, errwrite)
  File "C:\Python27\lib\subprocess.py", line 958, in _execute_child
    startupinfo)
TypeError: environment can only contain strings
```

mindw:
#162 Should solve this issue.
To manually bypass this - do the following while in pyOCD source root:
1. delete .egg directory
2. pip install setuptools-scm==1.5.2
3. run setup.py again.

PR #157, opened 2015-06-10
pyocd-tool and flash-tool improvements

flit:
Change pyocd-tool to focus on an interactive command mode.

Example interactive session:
<pre>
creed@sapo:pyOCD-flit$ pyocd-tool 
Connected to K64F [Halted]: 0240020245C50E13B838F3AB
>>> g
Successfully resumed device
>>> stat
Security:       Unlocked
MDM-AP Control: 0x00000000
MDM-AP Status:  0x0000003b
Core status:    Running
>>> h
Successfully halted device
>>> reg
     r0: 0x1fff0414       r6: 0x0000047e      r12: 0xcccccccc 
     r1: 0x00000000       r7: 0x77777777       sp: 0x1fff0688 
     r2: 0x0000002f       r8: 0x88888888       lr: 0x00000bf3 
     r3: 0x00000000       r9: 0x99999999       pc: 0x00000446 
     r4: 0x00000000      r10: 0xaaaaaaaa     xpsr: 0x61000000 
     r5: 0x0000044f      r11: 0xbbbbbbbb  primask: 0x00000000 
>>> r4
r4 = 0x00000000 (0)
>>> pc
pc = 0x00000446 (1094)
>>> d -c pc 32          
pc = 0x00000446
0x00000436:  856a      ldr     r5, [r0, #0x28]
0x00000438:  2e00      movs    r6, r5
0x0000043a:  02e0      b       #0x442
0x0000043c:  491e      subs    r1, r1, #1
0x0000043e:  0c19      adds    r4, r1, r4
0x00000440:  0600      movs    r6, r0
0x00000442:  dff80004  ldr.w   r0, [pc, #0x400]
0x00000446:* c069      ldr     r0, [r0, #0x1c]
0x00000448:  0028      cmp     r0, #0
0x0000044a:  2dd0      beq     #0x4a8
0x0000044c:  dff8f403  ldr.w   r0, [pc, #0x3f4]
0x00000450:  d0f81c80  ldr.w   r8, [r0, #0x1c]
0x00000454:  0027      movs    r7, #0
</pre>

Also added a few minor improvements to pyocd-flashtool such as being able to perform chip and sector erases without writing anything, i.e. `pyocd-flashtool -se -a0x40000 -n16` will erase 16 sectors starting at 0x40000. It also can use the file's extension instead of requiring the type to be provided on the command line.


pyOCD/target/cortex_m.py
```diff
@@ -371,31 +374,34 @@ def init(self, initial_setup=True, bus_accessible=True):
...
+            self.register_list.append(reg)
+            SubElement(xml_regs_general, 'reg', **reg.gdb_xml_attrib)
+        # Check if target has ARMv7 registers
+        if self.core_type in  (ARM_CortexM3, ARM_CortexM4):
+            for reg in self.regs_system_armv7_only:
+                self.register_list.append(reg)
+                SubElement(xml_regs_general, 'reg',  **reg.gdb_xml_attrib)
+        # Check if target has FPU registers
+        if self.has_fpu:
+            #xml_regs_fpu = SubElement(xml_root, "feature", name="org.gnu.gdb.arm.vfp")
```
c1728p9: I accidentally left this in.  Feel free to remove it if you want.

pyOCD/target/target.py
```diff
@@ -91,13 +95,13 @@ def reset(self):
...
-    
+
     # GDB functions
     def getTargetXML(self):
         return ''
-    
+
     def getMemoryMapXML(self):
-        return self.memoryMapXML
+            return self.memoryMapXML
```
c1728p9: Why the extra space?

pyOCD/target/target_kinetis.py
```diff
@@ -93,26 +93,27 @@ def init(self):
...
-        # release MDM halt once it has taken effect in the DHCSR
-        self.transport.writeAP(MDM_CTRL, 0)
-        
-        # sanity check that the target is still halted
-        if self.getState() == TARGET_RUNNING:
-            raise Exception("Target failed to stay halted during init sequence")
+            while self.transport.readAP(MDM_CTRL) & (MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET) != (MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET):
+                self.transport.writeAP(MDM_CTRL, MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET)
+            # Enable debug
+            self.writeMemory(DHCSR, DBGKEY | C_DEBUGEN)
```
c1728p9: I think setting C_DEBUGEN is required in the init sequence.

pyOCD/target/target_kinetis.py
```diff
@@ -93,26 +93,27 @@ def init(self):
         if isLocked:
             return
 
-        # Prevent the target from resetting if it has invalid code
-        self.transport.writeAP(MDM_CTRL, MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET)
-        while self.transport.readAP(MDM_CTRL) & (MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET) != (MDM_CTRL_DEBUG_REQUEST | MDM_CTRL_CORE_HOLD_RESET):
+        if self.halt_on_connect:
```
c1728p9: I like the ability to attach to a running target, but I'm not sure of the stability implications it will have.  If the target is running there are a number of things it can do that could cause a TransferError - like performing a software reset or entering a sleep mode.  I think this should work 99% of the time if there is valid code loaded though.

pyOCD/tools/flash_tool.py
```diff
@@ -144,12 +154,42 @@ def main():
             if args.hide_progress:
                 progress = None
 
+            # Convert address and count to int.
+            if args.address:
+                args.address = int(args.address, base=0)
```
c1728p9: Instead of converting address to an integer here you could add type checking to the args above.  That way argparse could print its own error message.  Something like "type=lambda val:int(val, 0)" in add_argument.  Or if that is too long you could define the conversion function explicitly.  It would probably also be good to use this same function for args, count and skip

pyOCD/tools/flash_tool.py
```diff
@@ -144,12 +154,42 @@ def main():
...
             elif args.sector_erase:
                 chip_erase = False
 
+            if not has_file:
+                if chip_erase:
+                    print("Erasing chip...")
+                    flash.init()
+                    flash.eraseAll()
+                    print("Done")
+                elif args.sector_erase and args.address:
```
c1728p9: args.address could be supplied as 0 in which case this will evaluate as false.  Maybe check for None instead.

pyOCD/tools/flash_tool.py
```diff
@@ -144,12 +154,42 @@ def main():
...
             if args.chip_erase:
                 chip_erase = True
             elif args.sector_erase:
                 chip_erase = False
 
+            if not has_file:
+                if chip_erase:
+                    print("Erasing chip...")
+                    flash.init()
+                    flash.eraseAll()
```
c1728p9: Is there any concern about this locking kinetis devices?

pyOCD/tools/flash_tool.py
```diff
@@ -77,7 +86,8 @@
 group.add_argument("-se", "--sector_erase", action="store_true", help="Use sector erase when programming.")
```
c1728p9: Originally sector_erase and chip_erase were meant to direct the method for program flash more than specifying and actual chip erase or page erase.  Right now page erase programming won't actually erase any pages if the data hasn't changed.  Maybe a new parameter -ft --flash-type <chiperase, pageerase, auto> could be used for programming instead of these.  This would remove some ambiguity.

pyOCD/tools/flash_tool.py
```diff
@@ -165,7 +205,7 @@ def main():
                                  fast_verify=args.fast_program)
 
             # Intel hex file format
-            if args.format == 'hex':
+            elif args.format == 'hex':
                 hex = IntelHex(args.file)
                 addresses = hex.addresses()
                 addresses.sort()
```
c1728p9: Good idea on the auto file type detection.  You might also want to have it print a warning if the file type could not be determined.

pyOCD/tools/flash_tool.py
```diff
@@ -144,12 +154,42 @@ def main():
...
+                    print("Erasing chip...")
+                    flash.init()
+                    flash.eraseAll()
+                    print("Done")
+                elif args.sector_erase and args.address:
+                    flash.init()
+                    page_addr = args.address
+                    for _ in range(args.count):
+                        page_info = flash.getPageInfo(page_addr)
+                        print("Erasing sector 0x%08x" % page_addr)
```
c1728p9: You might want to force address to be aligned to a page boundary, and possibly print a warning if it isn't.

pyOCD/tools/flash_tool.py
c1728p9: Just realized you already did this.  Ignore my comment.

c1728p9:
Will continue my review tomorrow.  Automated test passed.

```
------ TEST RESULTS ------
Target         Test           Result         Time

nrf51          Basic Test     Pass           4.71600008011
nrf51          Speed Test     Pass           40.0249998569
nrf51          Cortex Test    Pass           6.48300004005
nrf51          Flash Test     Pass           117.736999989
kl46z          Basic Test     Pass           3.82500004768
kl46z          Speed Test     Pass           13.7679998875
kl46z          Cortex Test    Pass           3.28200006485
kl46z          Flash Test     Pass           55.7750000954
k22f           Basic Test     Pass           3.6779999733
k22f           Speed Test     Pass           21.7300000191
k22f           Cortex Test    Pass           4.36999988556
k22f           Flash Test     Pass           73.6289999485
lpc1768        Basic Test     Pass           30.6550002098
lpc1768        Speed Test     Pass           33.5979998112
lpc1768        Cortex Test    Pass           5.80900001526
lpc1768        Flash Test     Pass           158.503000021
k64f           Basic Test     Pass           5.46799993515
k64f           Speed Test     Pass           49.3450000286
k64f           Cortex Test    Pass           5.90900015831
k64f           Flash Test     Pass           160.674000025
lpc800         Basic Test     Pass           5.09699988365
lpc800         Speed Test     Pass           10.5230000019
lpc800         Cortex Test    Pass           4.65300011635
lpc800         Flash Test     Pass           18.7649998665
lpc11u24       Basic Test     Pass           7.31299996376
lpc11u24       Speed Test     Pass           8.99100017548
lpc11u24       Cortex Test    Pass           4.29299998283
lpc11u24       Flash Test     Pass           20.7319998741

Test Time: 879.345999956
All tests passed
```

pyOCD/target/target.py
flit: Ah, that's a merge error. I'll fix it. Thanks!

pyOCD/target/target_kinetis.py
flit: It does still have issues if there is no valid code. We can continue to improve the capabilities.

Though to be honest, the main reason I added the no halt on connect support was for the older non-interactive version of pyocd-tool. In that version it was required in order for the go, halt, etc commands to work.

But I think this could be a generally useful feature. We'll just need to improve it over time.

pyOCD/tools/flash_tool.py
flit: Good idea.

pyOCD/tools/flash_tool.py
flit: Done

pyOCD/tools/flash_tool.py
flit: Yes, I like separating the action from the programming algorithm. But I'll leave that for another time. :)

pyOCD/tools/flash_tool.py
flit: Yes, this will certainly lock Kinetis devices. That's actually what I wanted, because it is useful for testing. But perhaps for most users it would be a bad thing. Do you think it should perform a mass erase instead?

pyOCD/tools/flash_tool.py
flit: Done

pyOCD/target/target_kinetis.py
flit: Doesn't seem to be necessary for no halt on connect, because `CortexM.halt()` will set it.

pyOCD/target/cortex_m.py
flit: Sorry, I don't understand. Can you explain more?

c1728p9: I accidentally left this commented out line here.  It can be removed, but don't feel obligated to.  The line is:
 #xml_regs_fpu = SubElement(xml_root, "feature", name="org.gnu.gdb.arm.vfp")

pyOCD/tools/flash_tool.py
c1728p9: I think it is fine to keep it like this.  Just wanted to make sure it wouldn't go by unnoticed.

pyOCD/utility/conversion.py
```diff
@@ -37,6 +38,22 @@ def word2byte(data):
         res.append((x >> 24) & 0xff)
     return res
 
+## @brief Convert a byte array into a halfword array.
+def byte2half(data):
+    byteData = []
+    for i in data:
+        byteData.extend([i & 0xff, (i >> 8) & 0xff])
```
c1728p9: This should be data[i] getting added to byteData, rather than just 'i'.  Also this function should swapped with half2byte.

pyOCD/utility/conversion.py
```diff
@@ -37,6 +38,22 @@ def word2byte(data):
...
+    return byteData
+
+## @brief Convert a halfword array into a byte array
+def half2byte(byteData):
```
c1728p9: This function should be swapped with byte2half

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                self.handle_list([])
+                return 0
+
+            if self.args.clock != DEFAULT_CLOCK_FREQ_KHZ:
+                print "Setting SWD clock to %d kHz" % self.args.clock
+
+            # Connect to board.
+            self.board = MbedBoard.chooseBoard(board_id=self.args.board, target_override=self.args.target, init_board=False, frequency=(self.args.clock * 1000))
+            self.board.target.setAutoUnlock(False)
+            self.board.target.setHaltOnConnect(False)
```
c1728p9: A prototype for these functions should probably be added in target.py, along with a stub in cortex_m.py.  That way targets without this will be able to use pyocd tool.

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                'w32' :     self.handle_write32,
+                'go' :      self.handle_go,
+                'g' :       self.handle_go,
+                'step' :    self.handle_step,
+                's' :       self.handle_step,
+                'halt' :    self.handle_halt,
+                'h' :       self.handle_halt,
+                'disasm' :  self.handle_disasm,
+                'd' :       self.handle_disasm,
+                'map' :     self.handle_memory_map,
```
c1728p9: This command crashes pyocd.  I think this will be fixed with your memory map patch.

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                'status' :  self.handle_status,
+                'stat' :    self.handle_status,
+                'reg' :     self.handle_reg,
+                'reset' :   self.handle_reset,
+                'read' :    self.handle_read8,
+                'read8' :   self.handle_read8,
+                'read16' :  self.handle_read16,
+                'read32' :  self.handle_read32,
+                'r' :       self.handle_read8,
+                'r8' :      self.handle_read8,
```
c1728p9: This command can't be run because it is interpreted as the register 'r8' instead

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                'list' :    self.handle_list,
+                'erase' :   self.handle_erase,
+                'unlock' :  self.handle_unlock,
+                'info' :    self.handle_info,
+                'i' :       self.handle_info,
+                'status' :  self.handle_status,
+                'stat' :    self.handle_status,
+                'reg' :     self.handle_reg,
+                'reset' :   self.handle_reset,
+                'read' :    self.handle_read8,
```
c1728p9: reading a register also reads the address the register is pointing to, often causing a transfer error
```
>>> r r2
r2 = 0x400ff0cc
400ff0cc:  00 00 00 00
```

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                'read32' :  self.handle_read32,
+                'r' :       self.handle_read8,
+                'r8' :      self.handle_read8,
+                'r16' :     self.handle_read16,
+                'r32' :     self.handle_read32,
+                'write' :   self.handle_write8,
+                'write8' :  self.handle_write8,
+                'write16' : self.handle_write16,
+                'write32' : self.handle_write32,
+                'w' :       self.handle_write8,
```
c1728p9: I'm not sure if this is a supported feature, but I can't seem to write to registers with this command

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                try:
+                    line = raw_input(self.PROMPT)
+                    line = line.strip()
+                    if line:
+                        self.process_command_line(line)
+                        self.last_command = line
+                    elif self.last_command:
+                        self.process_command(self.last_command)
+                except KeyboardInterrupt:
+                    print
```
c1728p9: What is the purpose of the empty print here?

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+                'g' :       self.handle_go,
+                'step' :    self.handle_step,
+                's' :       self.handle_step,
+                'halt' :    self.handle_halt,
+                'h' :       self.handle_halt,
+                'disasm' :  self.handle_disasm,
+                'd' :       self.handle_disasm,
+                'map' :     self.handle_memory_map,
+                'log' :     self.handle_log,
+                'clock' :   self.handle_clock
```
c1728p9: You might want to consider adding an 'exit' command.

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +76,490 @@ def dumpHexData(data, startAddress=0, width=8):
...
+
+            # Run command.
+            handler = self.tool.command_list[cmd]
+            handler(args)
+        except ValueError:
+            print "Error: invalid argument"
+            traceback.print_exc()
+        except pyOCD.transport.transport.TransferError:
+            print "Error: transfer failed"
+        except ToolError, e:
```
c1728p9: optional - pep8 recommends using 'as' rather than a comma.  This would be "except ToolError as e:".

c1728p9:
Great tool and thanks for all the new features.  Being able to run commands directly is really nice.  At this rate we will soon be able to bypass gdb all together and have our own :smile:.  I had a lot of minor comments, but the structure of everything looks good.  I'm glad you kept the command like functionality still, which allows people to run commands from something simple like a batch file.

Also I was unable to install the capstone library.  Does anything special need to be done to get it installed?  If I install it from a local archive I don't get any errors, but it pyocd still can't find it.  If I install it from pip I get the following error:
```
running install_data

error: can't copy 'src\build\capstone.dll': doesn't exist or not a regular file
```

pyOCD/tools/pyocd.py
posborne: The `as` syntax is also required for python 3.

pyOCD/utility/conversion.py
flit: Ugh, must have been asleep. Fixed this and the one above.

pyOCD/tools/pyocd.py
flit: setHaltOnConnect() is implemented in Target, and sets a halt_on_connect attribute.

Added setAutoUnlock() stub to Target.

pyOCD/tools/pyocd.py
flit: Done. Also removed map command.

pyOCD/tools/pyocd.py
flit: No, that's not supported. I'll add a 'wreg' command.

pyOCD/tools/pyocd.py
flit: Oops, that should've been obvious. I'll just remove the 'r8' and 'w8' aliases in favour of 'r' and 'w'.

pyOCD/tools/pyocd.py
flit: That's intentional. If you provide a register name for an address (or length), it will use that register's value. The most useful case is 'disasm pc'. But also useful for looking at structures that are pointed to by a register.

To just read the register's value use 'reg r2' or 'r2'.

pyOCD/tools/pyocd.py
flit: Thanks, I wasn't aware of that.

pyOCD/tools/pyocd.py
flit: When you type Ctrl-D to quit on a Posix system, the shell's prompt would appear on the same line as the last pyocd-tool prompt. This print prevents that. I added a comment and check so it doesn't print on Windows.

flit:
Thanks!

For running commands from a batch or shell file, we should add support for processing commands on either stdin or a specified text file.

I also had issues installing Capstone on Windows, with the same error from pip. I hadn't yet tried using a local archive.

Really, we should probably have our own disassembly library. Capstone is good because it is so simple, but it doesn't/can't hook into the ELF and DWARF info. So it won't tell you that some address is really PC-relative data, it just disassembles anyway and you get garbage instructions.

There's another Python disassembler library called amoco that is pretty amazing. But it's focus is on reconstructing the actions of the instruction sequence for security and related analysis.

pyOCD/utility/conversion.py
```diff
@@ -37,6 +38,20 @@ def word2byte(data):
...
+def half2byte(data):
+    byteData = []
+    for i in data:
+        byteData.extend([data[i] & 0xff, (data[i] >> 8) & 0xff])
+    return byteData
+
+## @brief Convert a byte array into a halfword array.
+def byte2half(byteData):
+    data = []
+    while i in range(0, len(byteData), 2):
```
c1728p9: I think this is meant to be a 'for' loop rather than a 'while'.  When I run it 'i' is not defined and causes an exception.

pyOCD/tools/pyocd.py
c1728p9: Looks like isLocked also needs to be added to target.

pyOCD/utility/conversion.py
flit: Oh, the original code was correct (though not the function names!), it just had a funny name for the **for** loop variable.

pyOCD/tools/pyocd.py
```diff
@@ -33,17 +36,135 @@
...
+        pyOCD.target.cortex_m.TARGET_HALTED : "Halted",
+        pyOCD.target.cortex_m.TARGET_RUNNING : "Running"
+        }
+
+## Default SWD clock in kHz.
+DEFAULT_CLOCK_FREQ_KHZ = 1000
+
+## Command info and help.
+COMMAND_INFO = {
+        'list' : {
```
c1728p9: Might want to remove this since it is not supported in interactive mode

flit: Done. I was originally thinking this help would also be used for non-interactive mode, but it's easier to just remove 'list' for now.

pyOCD/tools/pyocd.py
flit: Done.

pyOCD/tools/pyocd.py
c1728p9: Maybe it could be popped from help at the same time it is popped from command_list

flit: Funny, didn't even think of that since COMMAND_INFO was marked as a constant in my mind. Done.

c1728p9: You could also make a copy of it and pop from that if you wanted to keep COMMAND_INFO constant.

pyOCD/utility/conversion.py
c1728p9: Good catch.  I saw 'i' and just assumed it was an index.

c1728p9:
Looks good to go with all the fixes :+1: .

One further thought to consider:  A command line based tool like this is something that would be fairly straight forward to test with a script.  You could have a batch file (or the linux equivalent) to pipe in a series of commands and then pipe the output to a log.  When a change is made you could diff a log from the previous patch with a log from the current patch to see if anything broke.  You could also use the python coverage package to automatically generate code coverage metrics.

A test batch file like this:
```
coverage run pyocd.py < test_commands.txt > output.txt
coverage html
```

And a test script like this:
```
clock
clock 1000
clock 10000
d
d 0
d 0 10
log warning
log blarg
r16
r16 0
r16 0 10
r32
r32 0
r32 0 10
reg
help
g
h
regs
reset
h
r0
r1
r2
r3
r4
r5
r6
r7
r8
psp
r 3 100
r 0x3 100
r16 0x0
r32 0x0
s
step
stat
status
i
info
r 0
read 0
...
```

pyOCD/tools/pyocd.py
```diff
@@ -69,207 +190,514 @@ def dumpHexData(data, startAddress=0, width=8):
...
+    def handle_exit(self, args):
+        raise ToolExitException()
+
+    ## @brief Convert an argument to a 32-bit integer.
+    #
+    # Handles the usual decimal, binary, and hex numbers with the appropriate prefix.
+    # Also recognizes register names and address dereferencing. Dereferencing using the
+    # ARM assembler syntax. To dereference, put the value in brackets, i.e. '[r0]' or
+    # '[0x1040]'. You can also use put an offset in the brackets after a comma, such as
+    # '[r3,8]'. The offset can be positive or negative, and any supported base.
```
c1728p9: Are you using doxygen syntax here?  Python has built in support for documentation strings that might better to use instead.  More info on this can be found in pep-0257.  Probably not worth changing now, but I figured I would mention it for future reference.

0xc0170: +1

0xc0170:
@flit Can you please rebase?

flit:
@0xc0170, yes, I'll merge in the latest from master as soon as I get a chance.

pyOCD/tools/pyocd.py
flit: Yup, that's doxygen syntax. Sam and I had talked probably more than a year ago about using doxygen comments for pyOCD since they are much more expressive and you can generate much nicer documentation than docstrings. Doxygen does recognize docstrings, but they are treated as preformatted text and no doxygen commands are parsed, so the output looks ugly.

flit:
@0xc0170 , there were no merge differences when I merged master into dev_tools, so git didn't create a merge commit.

flit:
@0xc0170 Merged in master.

PR #159, opened 2015-06-10
Memory map objects

flit:
Replaced fixed GDB memory map XML with `MemoryMap` and `MemoryRegion` objects. The XML is generated at runtime from the objects when requested by GDB. The memory map is also used in `Flash` to look up page sizes and the ROM start address.

Haven't updated LPC devices with multiple page sizes to remove the `getPageInfo()` and/or `getFlashInfo()` methods of the `Flash` subclass, as I don't have hardware to test with.


pyOCD/flash/flash.py
```diff
@@ -211,16 +211,40 @@ def loadPageBuffer(self, bufferNumber, flashPtr, bytes):
...
+        region = self.target.getMemoryMap().getRegionForAddress(addr)
+
         info = PageInfo()
         info.erase_weight = DEFAULT_PAGE_ERASE_WEIGHT
         info.program_weight = DEFAULT_PAGE_PROGRAM_WEIGHT
-        info.size = self.flash_algo['page_size']
+        if region.isFlash and region.blocksize > 0:
+            info.size = region.blocksize
+        else:
+            inf.size = self.page_size
```
c1728p9: If addr is a valid flash address should blocksize ever be 0?  Maybe the else case could be removed and instead there could be a flag to indicate that this is a valid page or not.  When valid the size attribute would be present.

pyOCD/target/memory_map.py
```diff
@@ -0,0 +1,165 @@
...
+        return self._type == 'ram'
+
+    @property
+    def isRom(self):
+        return self._type == 'rom'
+
+    @property
+    def isBootMemory(self):
+        return self._is_boot_mem
+
```
c1728p9: Another property that would be good to have is isPoweredOnBoot.  That way the debugger knows to avoid things like external ram/flash that hasn't been setup yet but is in the memory map.  It would also allow test scripts to verify all of the powered regions can be read from.

pyOCD/target/target_lpc11u24.py
```diff
@@ -16,19 +16,17 @@
...
-    memoryMapXML =  """<?xml version="1.0"?>
-<!DOCTYPE memory-map PUBLIC "+//IDN gnu.org//DTD GDB Memory Map V1.0//EN" "http://sourceware.org/gdb/gdb-memory-map.dtd">
-<memory-map>
-    <memory type="flash" start="0x0" length="0x8000"> <property name="blocksize">0x400</property></memory>
-    <memory type="ram" start="0x10000000" length="0x1000"> </memory>
-</memory-map>
-"""
-    
+    memoryMap = MemoryMap(
+        FlashRegion(    start=0,           length=0x8000,       blocksize=0x400, isBootMemory=True),
```
c1728p9: The block size should be 0x1000.  That is the actual page_size used by flash_lpc11u24.py.

pyOCD/flash/flash.py
```diff
@@ -211,16 +211,40 @@ def loadPageBuffer(self, bufferNumber, flashPtr, bytes):
...
+        """
+
+        # prevent security settings from locking the device
+        bytes = self.overrideSecurityBits(flashPtr, bytes)
+
+        # first transfer in RAM
+        self.target.writeBlockMemoryUnaligned8(self.begin_data, bytes)
+
+        # update core register to execute the program_page subroutine
+        result = self.callFunction(self.flash_algo['pc_program_page'], flashPtr, len(bytes), self.begin_data)
```
c1728p9: Should this be callFunctionAndWait instead of callFunction, so when the call returns you know the processor is done writing?  Also is this the same as programPage?

pyOCD/tools/pyocd.py
```diff
@@ -218,8 +218,13 @@ def run(self):
                 else:
                     dumpHexData(data, args.read, width=args.width)
             elif args.action == ACTION_WRITE:
+                is_flash = self.isFlashWrite(args)
                 if args.width == 8:
-                    target.writeBlockMemoryUnaligned8(args.write, args.data)
+                    if is_flash:
+                        target.flash.init()
+                        target.flash.programPhrase(args.write, args.data)
```
c1728p9: Good idea!  We'll definitely want this feature.  I have a couple of comments that will probably need to be addressed in the future (or now).
-You might want to check for existing data before writing.  I know on Kinetis devices if non 0xFF data is programmed to the same spot multiple times without an erase then that address can get into a state where reads cause a fault.
-Not all processors support the same sized write.  Maybe a flash attribute for minimum flash write size could be added.  Or maybe a new mode could be added to flash builder which reads the page, and then writes it back with the new data.  That way individual bytes could be programmed.
-If the user had a program running this would mess up the state of that program, since the PC would now be pointing to a breakpoint in ram.  Probably out of the scope of this pull request, but if the state of the ram and registers were recorded then the execution context could be restored.  Something like this will be necessary for flash breakpoints in the future.

c1728p9:
This will be a very good change to have.  I ran automated_test.py and it passed on all the targets aside from the lpc11u24 because the block size is wrong.  I left a comment on the PR for this.  If it is updated then all the tests should pass.

pyOCD/flash/flash.py
flit: You're right, the blocksize should never be 0. That check was mostly added for error checking. Really, we could probably just remove the 'page_size' key from the flash algo dict, and related attribute in Flash.

Just noticed a typo on line 246 because of your comment! Thanks :)

pyOCD/target/memory_map.py
flit: Good idea.

pyOCD/target/target_lpc11u24.py
flit: Done.

pyOCD/flash/flash.py
flit: Good catch. The difference form programPage is that this will write only the number of bytes you pass in. programPage always writes the full page size as determined by page info. I guess not all devices support sub-page programming.

Maybe we should have a flash algo dict entry for that feature, as well as the minimum program length (phrase size in Freescale NVM terminology). Will add this.

pyOCD/tools/pyocd.py
flit: I've taken notes on these ideas. And implemented the minimum program length.

Unfortunately, you can't write individual bytes, at least on Kinetis. If you try to write to a phrase that has been written previously, even if you are only writing to bytes that are 0xFF and previously programmed bytes in that phrase are set to the same value, the flash gets unhappy.

c1728p9: I was thinking as a future improvement you could simulate single byte writes by copying the current page, erasing it, and writing it back.  This will be needed for flash breakpoints.

pyOCD/flash/flash.py
c1728p9: I would be for removing 'page_size'.  Any script that uses 'page_size' directly will not be compatible with all targets.

pyOCD/tools/pyocd.py
flit: Ah, I understand now. Yes, that would be nice, and is indeed needed for flash breakpoints.

pyOCD/flash/flash.py
flit: Ok, I'll go ahead and remove it.

c1728p9:
Turns out page_size was being used by some of the tests.  I created this branch if you want to pull some simple test fixes into your pr:
https://github.com/c1728p9/pyOCD/tree/fix_tests

pyOCD/target/memory_map.py
```diff
@@ -0,0 +1,170 @@
...
+        for r in self._regions:
+            if r.containsAddress(address):
+                return r
+        return None
+
+    def isValidAddress(self, address):
+        return self.regionForAddress(address) is not None
+
+    ## @brief Generate GDB memory map XML.
+    def getXML(self):
```
geky: Out of curiousity, why not move XML generation into a function that acts on MemoryMap objects in the gdbserver?

It seems like the XML is a very gdb specific thing and this would be the perfect pull request to move it out of the target directory completely.

flit:
Merged in Russ' test fixes.

pyOCD/target/memory_map.py
flit: Good idea. In general, I'd like to move the gdb-specific stuff to its own class and out from Target and CortexM. This would go along with that. We can do all of this in another pull request.

PR #160, opened 2015-06-11
rebase on top of the master and then added target for W7500 of WIZnet

kaizen8501:


pyOCD/target/target_w7500.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+from cortex_m import CortexM
+
+class W7500(CortexM):
+
+    memoryMapXML =  """<?xml version="1.0"?>
+<!DOCTYPE memory-map PUBLIC "+//IDN gnu.org//DTD GDB Memory Map V1.0//EN" "http://sourceware.org/gdb/gdb-memory-map.dtd">
+<memory-map>
+    <memory type="flash" start="0x0" length="0x8000"> <property name="blocksize">0x400</property></memory>
```
c1728p9: You have block size as 0x400 here, but 256 in flash_w7500.py.  These should be set to the same value.

pyOCD/target/target_w7500.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def resetStopOnReset(self, software_reset = None, map_to_user = True):
+        CortexM.resetStopOnReset(self, software_reset)
+
+        # Remap to use flash and set SP and SP accordingly
+        if map_to_user:
+            sp = self.readMemory(0x0)
+            pc = self.readMemory(0x4)
+            self.writeCoreRegisterRaw('sp', sp)
+            self.writeCoreRegisterRaw('pc', pc)
```
c1728p9: When the w7500 is reset do the pc and sp not get initialized properly?  If not, then this code shouldn't be needed - similar to target_k20d50m.py, or other non LPC targets.

pyOCD/target/target_w7500.py
```diff
@@ -0,0 +1,41 @@
...
+
+from cortex_m import CortexM
+
+class W7500(CortexM):
+
+    memoryMapXML =  """<?xml version="1.0"?>
+<!DOCTYPE memory-map PUBLIC "+//IDN gnu.org//DTD GDB Memory Map V1.0//EN" "http://sourceware.org/gdb/gdb-memory-map.dtd">
+<memory-map>
+    <memory type="flash" start="0x0" length="0x8000"> <property name="blocksize">0x400</property></memory>
+    <memory type="ram" start="0x10000000" length="0x1000"> </memory>
```
c1728p9: I think the ram and rom info needs to be updated as well.

pyOCD/flash/flash_w7500.py
```diff
@@ -0,0 +1,45 @@
...
+               'pc_init'          : 0x2000002B,
+               'pc_eraseAll'      : 0x20000033,
+               'pc_erase_sector'  : 0x20000045,
+               'pc_program_page'  : 0x20000057,
+               'static_base'      : 0x2000006C,               
+               'begin_data'       : 0x20001000, # Analyzer uses a max of 256 B data (64 pages * 4 bytes / page)
+               'begin_stack'      : 0x20004000,
+               'page_size'        : 256,
+               'analyzer_supported' : False,
+               'analyzer_address' : 0x20001000 # Analyzer 0x20001000..0x20001600
```
c1728p9: Right now the analyzer address is overlapping with 'begin_data', so it will get corrupted before it can run.  You might be able to use it if it is loaded to a free address, such as 0x20002000

0xc0170: We need to document these flash_algo. I vote for docs folder :smile:

0xc0170:
@c1728p9 thanks for reviewing :-)

0xc0170:
Thanks for updating, LGTM

Issue #143, opened 2015-05-26
Error loading .ELF via gdb_server on FRDM-K64F


thom-nic:
I built some examples from the Freescale KSDK 1.2 and I'm trying to load one in particular (httpsrv) but it crashes the GDB server when I do.  Here's the output from gdbserver:

```
± python test/gdb_server.py
Welcome to the PyOCD GDB Server Beta Version 
INFO:root:new board id detected: 02400221C3163E593EC8C3E1
INFO:root:board allows 5 concurrent packets
INFO:root:DAP SWD MODE initialised
INFO:root:IDCODE: 0x2BA01477
INFO:root:K64F not in secure state
INFO:root:6 hardware breakpoints, 4 literal comparators
INFO:root:CPU core is Cortex-M4
INFO:root:FPU present
INFO:root:4 hardware watchpoints
INFO:root:GDB server started at port:3333
INFO:root:One client connected!
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/Users/thom/dev/sandbox/kinetis/pyOCD/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 152, in run
    [resp, ack, detach] = self.handleMsg(data)
  File "/Users/thom/dev/sandbox/kinetis/pyOCD/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 219, in handleMsg
    return self.flashOp(msg[2:]), 1, 0
  File "/Users/thom/dev/sandbox/kinetis/pyOCD/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 360, in flashOp
    self.flashBuilder.addData(write_addr, self.unescape(data[idx_begin:len(data) - 3]))
  File "/Users/thom/dev/sandbox/kinetis/pyOCD/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 412, in unescape
    data[data_idx] = data[data_idx] ^ 0x20
IndexError: list index out of range
```

And on the gdb client side:

```
$ $ARMGCC_DIR/bin/arm-none-eabi-gdb httpsrv.elf         
GNU gdb (GNU Tools for ARM Embedded Processors) 7.6.0.20140731-cvs

(...... GDB header output ......)

(gdb) target remote localhost:3333
[...]

Issue #134, opened 2015-05-12
"IndexError: list index out of range" while loading executable from GDB to FRDM-K64F


autopulated:
I can't reproduce this, but thought it might be useful to report anyway:

```
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 153, in run
    [resp, ack, detach] = self.handleMsg(data)
  File "/usr/local/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 220, in handleMsg
    return self.flashOp(msg[2:]), 1, 0
  File "/usr/local/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 369, in flashOp
    self.flashData += self.unescape(data[idx_begin:len(data) - 3])
  File "/usr/local/lib/python2.7/site-packages/pyOCD/gdbserver/gdbserver.py", line 437, in unescape
    data[data_idx] = data[data_idx] ^ 0x20
IndexError: list index out of range
```

**Version: pyOCD 0.3.2**

c1728p9:
Thanks for the report.  Judging by the location of the crash, it looks like a gdb flashwrite packet got corrupted, causing the data to end with the escape character, 0x7d.  How often do you see this?  Also is your socket connection from GDB client to host on the same PC, or is it connected via the internet?

autopulated:
I've only seen this once. I'm connecting to localhost.

autopulated:
This same error just happened to me again (still on localhost). Restarting pyOCD and re-loading the same program worked fine.

c1728p9:
Closing due to lack of activity. Please re-open if you see this again.

PR #167, opened 2015-06-20
GDB server packet I/O thread 

flit:
Refactored all RSP packet I/O in `GDBServer` into a new `GDBServerPacketIOThread` class. The I/O class handles everything related to communication with GDB, including automatic ack'ing of valid packets and receiving ctrl-C interrupts. A received packet queue is provided that `GDBServer` accesses through the `receive()` method. The `send()` method writes to the socket immediately.

Improvements:
- Packet checksums are now verified.
- When acks are enabled, a nack will be sent for packets with invalid checksums.
- Packets are retransmitted if gdb sends a nack.
- Socket `send()` usage was corrected so the entire packet is guaranteed to be sent.
- Fixed potential bugs that could cause data loss on packet receipt. It looked like The buffer was being cleared in some cases where part of another packet could have been received. Maybe this was related to #143 and #134?
- Closed connections are handled correctly.
- No sleeping between socket reads. The I/O thread continuously reads from the socket. Writes happen asynchronously as soon `send()` is called.

Aside from just being a nice refactorization, these changes are required for semihosting and gdb syscall support.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -148,11 +148,8 @@ class GDBServer(threading.Thread):
                     self.buffer = ""
 
                     if resp is not None:
-                        # ack
-                        if ack and self.send_acks:
-                            resp = "+" + resp
                         # send resp
-                        self.send_packet(resp)
+                        self.send_packet(resp, ack)
 
                     if detach:
                         self.abstract_socket.close()
@@ -161,8 +158,6 @@ class GDBServer(threading.Thread):
                             break
                         else:
                             return
-
-                    self.timeOfLastPacket = time()
 
                 self.lock.release()
 
@@ -183,24 +178,30 @@ class GDBServer(threading.Thread):
             try:
                 if self.shutdown_event.isSet() or self.detach_event.isSet():
                     break
-                self.buffer += self.abstract_socket.read()
+                data = self.abstract_socket.read()
 
                 if LOG_PACKETS:
                     logging.debug('-->>>>>>>>>>>> GDB rsp packet: %s', data)
 
+                self.buffer += data
                 if self.buffer.index("$") >= 0 and self.buffer.index("#") >= 0:
+                    self.timeOfLastPacket = time()
                     break
             except (ValueError, socket.error):
                 pass
 
         self.abstract_socket.setBlocking(1)
 
-    def send_packet(self, packet):
+    def send_packet(self, packet, ack=True):
+        # ack
+        if ack and self.send_acks:
+            packet = "+" + packet
 
         if LOG_PACKETS:
             logging.debug('--<<<<<<<<<<<< GDB rsp packet: %s', packet)
 
         self.abstract_socket.write(packet)
+        self.timeOfLastPacket = time()
 
         if self.send_acks:
             # wait a '+' from the client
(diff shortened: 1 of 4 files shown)


posborne:
Initially, this looks like a good change.  I may take a shot at adding some unittest coverage, since none exists for pyOCD currently and I believe that unit tests are quite important for the overall health of a project in the long term.

posborne:
Ok, I decided it actually makes more sense to add the first set of unit tests against the main tree on something easy to test.  Hopefully this gets the ball rolling.

flit:
I had also been thinking about writing unit tests for the GDB server code. But you're right, this is some of the more difficult code to test in the project.

0xc0170:
@posborne the ball will keep rolling :smile: 


@flit "Aside from just being a nice refactorization, these changes are required for semihosting and gdb syscall support" more changes are coming !

flit:
@0xc0170 , yup, got a lot more changes coming :)

pyOCD/gdbserver/gdbserver.py
```diff
@@ -94,183 +268,146 @@ def setBoard(self, board, stop = True):
         self.flash = board.flash
         self.lock.release()
         return
-        
+
     def run(self):
         self.timeOfLastPacket = time()
```
c1728p9: You should be able to remove timeOfLastPacket since packet_io.receive() is blocking

pyOCD/gdbserver/gdbserver.py
```diff
@@ -24,10 +24,185 @@
...
+        super(GDBServerPacketIOThread, self).__init__(name="gdb-packet-thread")
+        self._abstract_socket = abstract_socket
+        self._receive_queue = Queue.Queue()
+        self._shutdown_event = threading.Event()
+        self.interrupt_event = threading.Event()
+        self.send_acks = True
+        self._clear_send_acks = False
+        self._buffer = ''
+        self._expecting_ack = False
+        self.time_of_last_packet = 0
```
c1728p9: It looks like time_of_last_packet is only set and never read.  Are there plans for it, or should it be removed?

pyOCD/gdbserver/gdbserver.py
```diff
@@ -332,21 +466,20 @@ def resume(self):
         self.timeOfLastPacket = time()
```
c1728p9: This can be changed to a local variable.  Not directly related to your patch, but figured I would mention it.

c1728p9:
The patch looks good.  A refactor like this has been needed for a while.  If you can wait a day I'll give this a more thorough testing tomorrow.

pyOCD/gdbserver/gdbserver.py
flit: I changed resume() to use time_of_last_packet from the packet I/O object, and removed both cases of setting timeOfLastPacket in GDBServer.

flit:
Please! I would appreciate further testing :)

Something I considered but didn't follow through with (yet) is changing the code so only packet contents, excluding the initial '$' and checksum, are put in the received queue and provided to send(). No reason for the command handlers to have to deal with that. This would also remove the need for createRSPPacket(), or move it into the packet thread. But it would require a lot of changes throughout the command handlers, so I decided to wait for another time.

pyOCD/gdbserver/gdbserver.py
flit: Actually, changed my mind. I remove the time of last packet attributes. Also changed resume() to call wait() on the interrupt_event instead of sleeping.

c1728p9:
I finally got around to trying this out and didn't see any problems.  The patch looks good to me.

flit:
@c1728p9, thanks for testing.

Once this pull request is merged, I can create a pull request for semihosting.

PR #168, opened 2015-06-20
tests: add initial unittests and related infrastructure

posborne:
These changes add tests for some functions for the conversion
utilities.  Since some of the names for these functions were
not very good, those were also updated to be more meaningful.
Finally,  the implementation for intToHex8 was modified to be
much simpler (after writing a test that covered the code with
the existing implementation).

Signed-off-by: Paul Osborne <osbpau@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,6 +101,15 @@ dependencies for the current platform by doing the following:
 On Windows, the virtualenv would be activated by executing
 ``env\Scripts\activate``.
 
+To run the unittests, you can execute the following.  Because of how
+nose searches for tests, specifying the directory is important as it
+will otherwise attempt to run non-unit tests as well (which will
+hang).
+
+.. code:: console
+
+    $ nosetest pyOCD/tests
+
 Examples
 --------
 
--- a/README.rst
+++ b/README.rst
@@ -108,7 +108,14 @@ hang).
 
 .. code:: console
 
-    $ nosetest pyOCD/tests
+    $ nosetests pyOCD/tests
+
+To get code coverage results, do the following:
+
+.. code:: console
+
+    $ nosetests --with-coverage --cover-html --cover-package=pyOCD pyOCD/tests
+    $ firefox cover/index.html
 
 Examples
 --------
--- a/pyOCD/test/test_utility/test_conversion.py
+++ b/pyOCD/test/test_utility/test_conversion.py
@@ -13,15 +13,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import unittest
-from pyOCD.utility.conversion import byteListToLittleEndianU32List, u32leListToBytelist, u16leListToByteList, \
-    byteListToU16leList, int2float, float2int, intToHex8, hex8ToInt, intToHex2, hexStringToIntList, hexDecode, hexEncode
+from pyOCD.utility.conversion import byteListToU32leList, u32leListToByteList, u16leListToByteList, \
+    byteListToU16leList, u32BEToFloat32BE, float32beToU32be, u32beToHex8le, hex8leToU32be, byteToHex2, hexToByteList, \
+    hexDecode, hexEncode
 
 
 class TestConversionUtilities(unittest.TestCase):
-
-    def test_byteListToLittleEndianU32List(self):
+    def test_byteListToU32leList(self):
         data = range(32)
-        self.assertEqual(byteListToLittleEndianU32List(data), [
+        self.assertEqual(byteListToU32leList(data), [
             0x03020100,
             0x07060504,
             0x0B0A0908,
@@ -43,9 +43,9 @@ class TestConversionUtilities(unittest.TestCase):
             0x1B1A1918,
             0x1F1E1D1C,
         ]
-        self.assertEqual(u32leListToBytelist(data), range(32))
+        self.assertEqual(u32leListToByteList(data), range(32))
 
-    def test_byteListToNibbleList(self):
+    def test_u16leListToByteList(self):
         data = [0x3412, 0xFEAB]
         self.assertEqual(u16leListToByteList(data), [
             0x12,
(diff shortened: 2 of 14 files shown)


pyOCD/utility/conversion.py
```diff
@@ -18,18 +18,18 @@
 import struct
 import binascii
 
-## @brief Convert a byte array into a word array.
-def byte2word(data):
+## @brief Convert a list of bytes to a list of 32-bit integers (little endian)
+def byteListToLittleEndianU32List(data):
```
flit: Any reason this function has "LittleEndian" spelled out, whereas the others use "le"?

pyOCD/utility/conversion.py
```diff
@@ -18,18 +18,18 @@
...
+    for i in range(len(data) / 4):
+        res.append(data[i * 4 + 0] |
+                   data[i * 4 + 1] << 8 |
+                   data[i * 4 + 2] << 16 |
+                   data[i * 4 + 3] << 24)
     return res
 
 ## @brief Convert a word array into a byte array.
-def word2byte(data):
+def u32leListToBytelist(data):
```
flit: "list" after "Byte" is not capitalized as in the related functions.

pyOCD/utility/conversion.py
```diff
@@ -64,18 +64,7 @@ def float2int(data):
...
-    for i in range(8-size):
-        r += '0'
-    r += str(val)
-
-    resp = ''
-    for i in range(4):
-        resp += r[8 - 2*i - 2: 8 - 2*i]
-
-    return resp
+    return "%08X" % int(val)
```
flit: I think this is incorrect. When I use the original implementation, I see this:
<pre>
In [3]: intToHex8(0x01020304)
Out[3]: '04030201'
</pre>

It's reversing the byte order of the output string.

While we're fixing names here, we should probably rename these to `intToHex8LE()` and similar for the related functions.

pyOCD/test/test_utility/test_conversion.py
```diff
@@ -0,0 +1,89 @@
...
+            0x0B0A0908,
+            0x0F0E0D0C,
+            0x13121110,
+            0x17161514,
+            0x1B1A1918,
+            0x1F1E1D1C,
+        ]
+        self.assertEqual(u32leListToBytelist(data), range(32))
+
+    def test_byteListToNibbleList(self):
```
flit: This test method name seems to not match the test case.

pyOCD/utility/conversion.py
posborne: Will change to be `le`.  Was playing with the names a bit.

pyOCD/utility/conversion.py
posborne: Will fix.

pyOCD/utility/conversion.py
posborne: Good catch.  Reverting the code makes the test case fail, so that's no good.  I'll revert the implementation or provide an implementation that passes this test case:

```

    def test_intToHex8(self):
        self.assertEqual(intToHex8(0x0102ABCD), "cdab0201")
```

README.rst
```diff
@@ -101,6 +101,15 @@ dependencies for the current platform by doing the following:
...
 
+To run the unittests, you can execute the following.  Because of how
+nose searches for tests, specifying the directory is important as it
+will otherwise attempt to run non-unit tests as well (which will
+hang).
+
+.. code:: console
+
+    $ nosetest pyOCD/tests
+
```
posborne: I'm going to add an example showing capturing code coverage since that is very useful.  You can capture and generate HTML coverage results by doing the following.  To enable this, I will also add the `coverage` package to the `dev-requirements.txt`:

```
nosetests --with-coverage --cover-html --cover-package=pyOCD pyOCD/test
```

c1728p9:
Hi @posborne , good idea.  It will be very helpful to have some formal unit testing.

I haven't used nose before, so I gave it a try with this pull request.  When I run "nosetests pyOCD/test" from the project directory I get an import error from board.py.  I found by chance that if I run "nosetests pyOCD.test" from the same directory that the tests work.  Any idea what is happening?

nosetests pyOCD/test:
```
======================================================================
ERROR: Failure: ImportError (No module named target)
----------------------------------------------------------------------
  File "C:\Python27\lib\site-packages\nose\loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "C:\Python27\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "C:\Python27\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "E:\Projects\mbed\pyOCD\pyOCD\pyOCD\__init__.py", line 18, in <module>
    import board
  File "E:\Projects\mbed\pyOCD\pyOCD\pyOCD\board\__init__.py",line 18, in <module>
    import mbed_board
  File "E:\Projects\mbed\pyOCD\pyOCD\pyOCD\board\mbed_board.py", line 22, in <module>
    from board import Board
  File "E:\Projects\mbed\pyOCD\pyOCD\pyOCD\board\board.py", line 18, in <module>
    from pyOCD.target import TARGET
ImportError: No module named target

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (errors=1)
```




nosetests pyOCD.test:
```
............
----------------------------------------------------------------------
Ran 12 tests in 0.002s

OK
```

pyOCD/tools/pyocd.py
```diff
@@ -545,7 +545,7 @@ def do_read(self, args):
         addr = self.convert_value(args[0])
         if len(args) < 2:
             count = 4
-                else:
+        else:
```
posborne: I think there may be some other errors in pyocd.py from another recent change to `pyocd-tool`.

pyOCD/utility/conversion.py
```diff
@@ -38,64 +40,63 @@ def word2byte(data):
...
-
-    return resp
-
-## @brief Build 32-bit register value from little-endian 8-digit hexadecimal string.
-def hex8ToInt(data):
+
+def u32BEToFloat32BE(data):
+    """Convert a 32-bit int to an IEEE754 float"""
+    d = struct.pack(">I", data)
+    return struct.unpack(">f", d)[0]
```
posborne: Ok, I could use some confirmation that this behavior is actually correct.  Previously this used `@` in the struct format string which is native byte order.  In my experience, using native is almost never actually  what is wanted in python as you don't really want to assume anything about the machine you are running on when serializing and deserializing.  Since most host PCs are big endian, I changed it to be that.

posborne:
@c1728p9, I'm not sure why that is happening for you and not for me.  The ImportError appears to be the result of a circular import (for this reason, I try to avoid putting much of anything in `__init__.py` because it gets imported implicitly in a number of situations.

This is a long shot, but does clearing out pyc files do anything for you?

posborne:
As a follow-on, once this is merged, we can probably go ahead and add travisci support so that the unittests are run automatically and gate merging PRs if they are failing.

c1728p9:
@posborne ,that long shot was dead on :smile:.  After I deleted those nosetests ran without problems.  Any idea why the pyc files would cause problems?

posborne:
@c1728p9 , nose has a protocol for discovering tests.  As part of this, it will import various modules to search for tests.  It doesn't know not to import `.pyc` files, so it must have imported one that was sitting around that had bad contents and choked.

0xc0170:
+1 for unit tests !

flit:
Does py.test deal with `.pyc` files properly? The Kinetis ROM team tried out both nose and py.test extensively, and found that for the basic stuff they were nearly identical. But py.test has a lot more advanced features and seemed to be more actively developed. Not sure if that's changed in the past year and a half.

pyOCD/tools/pyocd.py
flit: Thanks for noticing. I had the "compress whitespace" option turned on in Apple's FileMerge that I use for conflict resolution, and I think it was causing these indent issues on some merges. I'm working on a pull request to fix this and add a test script for pyocd-tool.

flit: Fixed in #169.

posborne:
@flit I don't think nose is doing any wrong with regards to pyc.  Its behavior is reasonable.  I'm fine with either but have had no problems using nose on large projects (>1000 unit tests).  As long as tests are written using something like the `unittest` framework, discovering them should be possible and a generally reasonable thing.  If we wanted to try to run the automated non-unit tests via a runner, py.test might have some nice features.

PR #175, opened 2015-07-20
gdb server non-stop mode support

flit:
This PR adds support for [non-stop mode](https://sourceware.org/gdb/current/onlinedocs/gdb/Remote-Non_002dStop.html#Remote-Non_002dStop). This feature allows for accessing memory while the target is running.

These changes were tested with command-line arm-none-eabi-gdb version 7.8.0.20150604 from GNU ARM Embedded 2015q2. To enable non-stop mode, issue the `set non-stop on` command before connecting to the gdb server. Note that you need to use `continue&` to run the target and still have gdb command access.

Eclipse supposedly has [support for non-stop mode](http://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.cdt.doc.user%2Freference%2Fcdt_u_dsfgdb.htm). However, it doesn't seem to work at all. Setting this preference doesn't actually cause non-stop mode to be requested to the gdb server.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -16,7 +16,7 @@
 """
 
 import logging, threading, socket
-from ..target.target import (TARGET_HALTED, BREAKPOINT_HW, BREAKPOINT_SW,
+from ..target.target import (TARGET_HALTED, TARGET_RUNNING, BREAKPOINT_HW, BREAKPOINT_SW,
     WATCHPOINT_READ, WATCHPOINT_WRITE, WATCHPOINT_READ_WRITE)
 from ..transport import TransferError
 from ..utility.conversion import hexToByteList, hexEncode, hexDecode
@@ -239,7 +239,7 @@ class GDBServer(threading.Thread):
         self.packet_io = None
         self.gdb_features = []
         self.non_stop = False
-        self.is_target_running = False
+        self.is_target_running = (self.target.getState() == TARGET_RUNNING)
         self.flashBuilder = None
         self.lock = threading.Lock()
         self.shutdown_event = threading.Event()
@@ -333,6 +333,7 @@ class GDBServer(threading.Thread):
                 if self.packet_io.interrupt_event.isSet():
                     if self.non_stop:
                         self.target.halt()
+                        self.is_target_running = False
                         self.sendStopNotification()
                     else:
                         logging.debug("Got unexpected ctrl-c, ignoring")
@@ -342,9 +343,11 @@ class GDBServer(threading.Thread):
                     try:
                         if self.target.getState() == TARGET_HALTED:
                             logging.debug("state halted")
+                            self.is_target_running = False
                             self.sendStopNotification()
-                    except:
-                        logging.debug('Target is unavailable temporary.')
+                    except Exception as e:
+                        logging.debug("Unexpected exception: %s", e)
+                        traceback.print_exc()
 
                 # read command
                 try:
@@ -393,7 +396,7 @@ class GDBServer(threading.Thread):
 
         # query command
         if msg[1] == '?':
-            return self.createRSPPacket(self.target.getTResponse()), 0
+            return self.stopReasonQuery(), 0
 
         # we don't send immediately the response for C and S commands
         elif msg[1] == 'C' or msg[1] == 'c':
@@ -510,6 +513,13 @@ class GDBServer(threading.Thread):
             self.target.removeWatchpoint(addr, size, watchpoint_type)
         return self.createRSPPacket("OK")
 
+    def stopReasonQuery(self):
+        # In non-stop mode, if no threads are stopped we need to reply with OK.
+        if self.non_stop and self.is_target_running:
+            return self.createRSPPacket("OK")
+
+        return self.createRSPPacket(self.target.getTResponse())
+
     def _get_resume_step_addr(self, data):
         if data is None:
             return None
@@ -529,10 +539,6 @@ class GDBServer(threading.Thread):
         self.target.resume()
         logging.debug("target resumed")
 
-        if self.non_stop:
-            self.is_target_running = True
-            return self.createRSPPacket("OK")
-
         val = ''
 
         while True:
@@ -545,7 +551,7 @@ class GDBServer(threading.Thread):
                 logging.debug("receive CTRL-C")
                 self.packet_io.interrupt_event.clear()
                 self.target.halt()
-                val = self.target.getTResponse(True)
+                val = self.target.getTResponse(forceSignal=signals.SIGINT)
                 break
 
             try:
@@ -583,8 +589,8 @@ class GDBServer(threading.Thread):
         self.target.halt()
         return self.createRSPPacket(self.target.getTResponse())
 
-    def sendStopNotification(self):
-        data = self.target.getTResponse()
+    def sendStopNotification(self, forceSignal=None):
+        data = self.target.getTResponse(forceSignal=forceSignal)
         packet = '%Stop:' + data + '#' + checksum(data)
         self.packet_io.send(packet)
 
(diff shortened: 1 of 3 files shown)


flit:
Rebased onto latest master.

pyOCD/gdbserver/gdbserver.py
```diff
@@ -329,12 +338,27 @@ def run(self):
                     break
 
                 if self.packet_io.interrupt_event.isSet():
-                    logging.debug("Got unexpected ctrl-c, ignoring")
+                    if self.non_stop:
```
c1728p9: Should this halt the target when non_stop mode is turned off?

pyOCD/gdbserver/gdbserver.py
```diff
@@ -329,12 +338,27 @@ def run(self):
...
+                            self.is_target_running = False
+                            self.sendStopNotification()
+                    except Exception as e:
+                        logging.debug("Unexpected exception: %s", e)
+                        traceback.print_exc()
+
                 # read command
                 try:
-                    packet = self.packet_io.receive()
+                    packet = self.packet_io.receive(block=not self.non_stop)
```
c1728p9: If all the targets are halted in non_stop mode then you could probably block here

pyOCD/gdbserver/gdbserver.py
```diff
@@ -489,7 +524,29 @@ def breakpoint(self, data):
...
+        # c[;addr]
+        if data[0] in ('c', 's'):
+            addr = int(data[2:], base=16)
+        # Csig[;addr]
+        elif data[0] in ('C', 'S'):
+            addr = int(data[1:].split(';')[1], base=16)
+        return addr
+
+    def resume(self, data):
+        addr = self._get_resume_step_addr(data)
```
c1728p9: It doesn't look like anything is being done with addr here and in step.  Do you have future plans for this?

pyOCD/gdbserver/gdbserver.py
```diff
@@ -363,11 +391,15 @@ def run(self):
                         if self.persist:
                             break
                         else:
-                            self._cleanup()
+                            self.shutdown_event.set()
                             return
 
                 self.lock.release()
 
+            except Exception as e:
```
c1728p9: If an exception is thrown here then the current command won't be able to return a response of failure.  Would it make sense to respond with a generic error here?  Or maybe an exception could be handled at the individual command level.

pyOCD/gdbserver/gdbserver.py
```diff
@@ -498,14 +555,14 @@ def resume(self):
...
-                return self.createRSPPacket(val), 0
+                return self.createRSPPacket(val)
 
             # Wait for a ctrl-c to be received.
             if self.packet_io.interrupt_event.wait(0.01):
                 logging.debug("receive CTRL-C")
                 self.packet_io.interrupt_event.clear()
                 self.target.halt()
-                val = self.target.getTResponse(True)
+                val = self.target.getTResponse(forceSignal=signals.SIGINT)
```
c1728p9: I get the exception "NameError: global name 'signals' is not defined" when I run this in stop mode

c1728p9:
What commands can be run while the target is running?  All of them?  If so, it seems like this could cause some undefined behavior and exceptions.

pyOCD/gdbserver/gdbserver.py
flit: Probably not. In all-stop mode, the only time the target should be running is in `resume()`, where we already have a loop that checks for ctrl-c.

But now I'm wondering if all-stop should be more integrated into the main packet receive loop, like non-stop is. `resume()` would just resume the target and set the target-running flag. Then the main loop would check for ctrl-c as with non-stop. Originally, I was trying to add non-stop with as little change to the rest of the code as possible, but maybe that's not the right way to do it. Or maybe we just clean this up later.

pyOCD/gdbserver/gdbserver.py
flit: If I block here, then we can't check for ctrl-c events. This could probably be reorganized to pass a timeout into `receive()` instead of setting non-blocking and then calling `sleep()` a few lines down.  `GDBServer.syscall()` also uses the same pattern and could be similarly changed. What is your preference?

pyOCD/gdbserver/gdbserver.py
flit: Yes, I was going to support the optional address parameter on the step and continue commands. Not sure that gdb actually uses that parameter, though.

pyOCD/gdbserver/gdbserver.py
flit: I think exceptions _should_ be handled at the command level. These new exception handlers were added just to ensure we don't exit the connection loop while gdb still thinks we're still connected, or let an unhandled exception terminate the thread.

If wanted to return a generic error response, we could add an exception handler in `handleMsg()`. In these two exception handlers in `run()` and `_run_connection()`, we don't necessarily know that gdb is waiting for a response.

pyOCD/gdbserver/gdbserver.py
flit: Thanks, I'll fix it.

flit:
Being able to access memory while the target is running is a major feature of the Cortex-M debug architecture. So not being able to make use of this feature from gdb is an issue. Erich Styger has said that many people are asking for it in KDS, since it was available in CodeWarrior. Other debug vendors (including Segger) have said they are looking into adding non-stop mode support.

I'm not completely sure what all commands can be run. I know you can't do things like get back traces or view stack frames and local variables of a running thread. The most important thing is that you _can_ read global variables or other memory.

The gdb docs for non-stop mode are [here](https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode).

pyOCD/gdbserver/gdbserver.py
c1728p9: Oh gotcha, that makes sense.  For some reason I was thinking that the reception of any data on the socket, including ctrl-c, would cause receive to return.  Now that you mention it, it is obvious why this is not the case.  I think the way you have it now is good.

pyOCD/gdbserver/gdbserver.py
c1728p9: Yeah, that is a good point that all-stop mode would have already halted the target.   I see no downside to the way you have it now, so I wouldn't worry about changing it.  Out of curiosity, how should ctrl-c behave when there are multiple threads?  Should it stop all the threads, or just the current one?  Or should another command such as vCont be used instead of ctrl-c in non-stop mode for well defined behavior among all threads?

c1728p9:
I agree, having a non-stop mode in KDS would be incredibly useful.  I use a Lauterbach debugger sometimes and stop mode is a great feature on it.  Did the eclipse version of CodeWarrior support non-stop mode?

If the memory access commands are the ones that primarily get run then that is good news for stability.  Those functions already have error reporting, and are safe to run when the target is executing code.

Also, good work on this patch.  I kinda wrote off non-stop mode as something that would be infeasible to implement.  It looks like you already have it working though :smile:.

Finally, if this patch can wait a couple days I can try it out at work.  Don't feel like you need to wait on me to merge it though.  I did some preliminary testing with GDB and Eclipse yesterday, and after importing signals, everything worked.

pyOCD/gdbserver/gdbserver.py
flit: The [non-stop mode docs](https://sourceware.org/gdb/current/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode) say that ctrl-c should only apply to the current thread in non-stop mode. But that may just be a UI behaviour, and it might send vCont to the server stub.

pyOCD/gdbserver/gdbserver.py
flit: I may go ahead and change `receive()` to accept a timeout param instead of block.

flit:
I think we should wait until you can try it out! I would really appreciate your help testing.

I had also written off non-stop mode as not very useful until Erich asked about it. I'm pretty sure it was the Eclipse version of CodeWarrior that he was referencing, but I don't think CW uses gdb. It has its own debug infrastructure. (They still use it for the LayerScape chips with the Cortex-A5x cores.)

Please let me know if you can get non-stop mode to work with Eclipse. It works fine with command-line gdb, but Eclipse seems to ignore the non-stop mode preference and gets confused if you force non-stop to be enabled via `set non-stop on`. I'm not sure if it has something to do with the GNU ARM Eclipse plugin.

PR #176, opened 2015-08-03
Fix TabError

devanlai:
Replaced a rogue tab mixed in with spaces in target_stm32f103rc.py. Mixing whitespace is bad in general, but it also produces a TabError when running on Python 3.


jamesadevine:
@0xc0170 Is this pushed on PIP??

0xc0170:
Hello, not yet. I'll look at the changeset since the last release, might publish a new one soon

Issue #177, opened 2015-08-03
pyusb dependency failing


autopulated:
Hi,

I've had some reports that the pyusb dependency is failing to install on Linux (Ubuntu 14.04.2 LTS, pip 1.5.6, python 2.7.6)

```
 Could not find a version that satisfies the requirement pyusb (from pyOCD>=0.3,<1.0->valinor>=0.0.0,<1.0->yotta) (from versions: 1.0.0a2, 1.0.0a2, 1.0.0a3, 1.0.0a3, 1.0.0b1, 1.0.0b2)
  Some externally hosted files were ignored (use --allow-external to allow).
```

It isn't obvious to me why this is. It's possible that pip isn't liking the non-semantic versions that pyusb has. pip does seem to be extremely old on this system...

c1728p9:
Hi @autopulated , I had this same problem.  Even with an updated version of pip, the command "pip install pyusb" fails on Ubuntu.  I was able to get pyusb to install by specifying the version explicitly " pip install pyusb==1.0.0b2".

0xc0170:
This should be resolved in the v0.5.x.

PR #178, opened 2015-08-17
Fix pyusb dependency

c1728p9:
Installation of pyOCD fails at the time of this patch on platforms that
use pyusb, like Linux.  This is because the only released versions
of pyusb are alpha or beta, which pip does not install by default.
This patch sets the minimum version to the latest beta so pyusb
can be installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ import sys
 install_requires = ['intelhex']
 if sys.platform.startswith('linux'):
     install_requires.extend([
-        'pyusb',
+        'pyusb>=1.0.0b2',
     ])
 elif sys.platform.startswith('win'):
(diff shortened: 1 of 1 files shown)


c1728p9:
This should fix #177

0xc0170:
@c1728p9 LGTM. How did you test this?

c1728p9:
I was testing on an Ubuntu virtual machine.  I tested with the following sequence:

1. Uninstalled pyusb "pip uninstall pyusb"
2. Uninstalled pyOCD "pip uninstall pyocd"
3. Verified that pyOCD fails when trying to install current master "pip install -e ."
4. Apply changes verified install works using "pip install -e ."

PR #193, opened 2015-10-06
Removed reference to non-existent "chooseInterface" 

### LINKED ISSUE #192: BUG: `chooseInterface` function does not exist
I noticed in the `Board` class that there is some interface initialization code that refers to a function called `chooseInterface` that is currently not part of the code base. Here is the relevant piece:

```python
class Board(object):
    """
    This class associates a target, a flash, a transport and an interface
    to create a board
    """
    def __init__(self, target, flash, interface, transport="cmsis_dap", frequency=1000000):
        if isinstance(interface, str) == False:
            self.interface = interface
        else:
            self.interface = INTERFACE[interface].chooseInterface(INTERFACE[interface])
```

I'm guessing this function just got taken out of the code at some point and was replaced with something else (`MbedBoard.chooseBoard`?).

tgarc:
It doesn't seem easily possible to auto-initialize an `Interface` (interface initialization parameters are currently set inside the `getAllConnectedInterface` function inside the usb backend) so I decided to simply remove the auto-initialization code for the `interface` inside `Board.__init__`.

Closes #192

### DIFF
--- a/pyOCD/board/board.py
+++ b/pyOCD/board/board.py
@@ -29,10 +29,7 @@ class Board(object):
     to create a board
     """
     def __init__(self, target, flash, interface, transport="cmsis_dap", frequency=1000000):
-        if isinstance(interface, str) == False:
-            self.interface = interface
-        else:
-            self.interface = INTERFACE[interface].chooseInterface(INTERFACE[interface])
+        self.interface = interface
         self.transport = TRANSPORT[transport](self.interface)
         self.target = TARGET[target](self.transport)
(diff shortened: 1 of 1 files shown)


0xc0170:
Can you please sign http://developer.mbed.org/contributor_agreement/ ?

LGTM this changeset

tgarc:
@0xc0170 Done.

c1728p9:
Looks good to me!  Thanks for the PR

PR #194, opened 2015-10-06
tools/pyocd: fixed -H option

tgarc:
The tools/pyocd.py program crashes when passing the -H option since there were some arguments missing from a function call to `handle_halt`. I simply added the `self.args` to the call. I wasn't sure whether I should open an issue first but I figured since this is such a small bug a direct PR would be ok.

### DIFF
--- a/pyOCD/tools/pyocd.py
+++ b/pyOCD/tools/pyocd.py
@@ -390,7 +390,7 @@ class PyOCDTool(object):
 
             # Halt if requested.
             if self.args.halt:
-                self.handle_halt()
+                self.handle_halt([])
 
             # Handle a device with flash security enabled.
             self.didErase = False


pyOCD/tools/pyocd.py
```diff
@@ -390,7 +390,7 @@ def run(self):
 
             # Halt if requested.
             if self.args.halt:
-                self.handle_halt()
+                self.handle_halt(self.args)
```
0xc0170: You could pass empty list there?

tgarc:
Replaced `self.args` with an empty list.

0xc0170:
LGTM

c1728p9:
Thanks for the pull request!  Looks good to me as well.

PR #196, opened 2015-10-08
Add support for the nrf52

c1728p9:
Add support for the nrf52.  The nrf52 shares the same flash algorithm
and has many of the same registers of the nrf51.  Some differences
include rom size, ram size and block size.


0xc0170:
+1

PR #197, opened 2015-10-12
Remove Cython as a setup dependency for Mac

c1728p9:
The package cython was added as a dependency in pyOCD to fix a missing
dependency in hidpi.  This dependency has been properly added to
hidapi so this workaround can be removed.

This patch reverts the following commit:
712ad33d780bc41a02ee67a8b49d4829b4b74ac8 -
"Adding Cython as a setup dependency for Mac"

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@
 from setuptools import setup, find_packages
 import sys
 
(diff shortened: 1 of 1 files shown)


0xc0170:
+1

PR #198, opened 2015-10-20
Fix windows usb problems

c1728p9:
Switch to the latest pywinusb to bring in the following fixes:
-Fix intermittent python.exe crash on some usb operations
-Prevent multiple scripts from accessing the same board, since
    this leads to one of the scripts raising an exception.

Also, raise an exception when the USB backend is missing.  This prevents
the current script from crashing at some future point with a less
understandable error message.


0xc0170:
:+1:

PR #199, opened 2015-10-20
Add return values for continuous integration

c1728p9:
Return -1 on failure to allow continuous integration testing to
correctly detect errors.


test/automated_test.py
```diff
@@ -87,7 +87,9 @@
     print("Test Time: %s" % (stop - start))
     if Test.all_tests_pass(result_list):
         print("All tests passed")
+        exit(0)
```
0xc0170: shouldn't we use exit from sys ? or we do?

PR #205, opened 2015-10-28
protect against re-initing already initd sockets

autopulated:
This fixes:

https://github.com/ARMmbed/valinor/issues/16
https://github.com/ARMmbed/example-mbedos-blinky/issues/9

There also might be an underlying error causing this object to be re-initialised twice in the first place. Based on my understanding of the code it wasn't obvious if/where that is.

cc @sg- @0xc0170

### DIFF
--- a/pyOCD/gdbserver/gdb_socket.py
+++ b/pyOCD/gdbserver/gdb_socket.py
@@ -26,10 +26,11 @@ class GDBSocket(object):
         self.host = ''
 
     def init(self):
-        self.s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        self.s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        self.s.bind((self.host, self.port))
-        self.s.listen(5)
+        if self.s is None:
+            self.s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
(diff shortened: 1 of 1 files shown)


0xc0170:
LGTM

c1728p9:
Looks good here.  Test bot is down through.  If you have tested it locally then it can be merged.

autopulated:
I only tested loading from GDB onto a K64F: pretty sure this hasn't broken anything because the changed code paths would previously have raised exceptions, but if you have hardware to run the tests handy then it'd be worth running them.

c1728p9:
I took a closer look at this bug to understand the root cause.  This is a timing problem, and the root problem is fixed by @autopulated 's patch.  The failure occurs when the call to select in GDBSocket.connect times out.  This timeout causes connect to return None.  This in turn causes the thread TelnetSemihostIOHandler._server to call connect a second time (It calls connect in a loop until something other than None is returned).  Since the socket is already open this causes an exception.  If the timeout is increased from 0.5s to a larger value, like 10s then problem no longer occurs.  The best solution though is to keep the timeout the same and only open the socket once.

For reference here is the line with the timeout:
https://github.com/mbedmicro/pyOCD/pull/205/files#diff-9ad088fa41f36a270925dace8cf5f136R38

0xc0170:
@c1728p9 Thanks for having a closer look

PR #207, opened 2015-11-03
Refactor pyOCD in preperation for pyDAPAccess

c1728p9:
Prepare pyOCD for daplink by using pyDapLink's interface.
Prior to this patch the board class depending on having both a
transport and an interface.  This caused the code to be more
tightly coupled to interfaces than it could be.

This patch combines most of the logic from the interface and transport
class and moves it into the DapLink class.  The DapLink class is an
implementation of the Link class, which provides all the functions
needed to communicate with a board.

More specific changes:
-Raw packet transfer test removed from speed test since pyDapLink
    will now be responsible for testing this
-References to 'transport' changed to 'link'
-Transport interface removed
-Interface code (usb drivers) moved into the transport folder to
    further isolate it from the rest of the codebase
-File cmsis_dap.py temporarily moved into target folder for use by
    cortex_m.py and target_kinetis.py.  In the future cmsis_dap.py
    will be removed and AP/DP drives will be used in its place.


c1728p9:
Any and all feedback on this change is appreciated.  This is a fairly large restructuring.

0xc0170:
I would like to find a moment this week to review this. Why did you decide to do all within one commit?

c1728p9:
There is no need to merge this immediately, so feel free to leave it up this week and look at it when you have time. I kept this all as one commit since it was one logical change and I didn't see a meaningful way to break this into multiple commits.

pyOCD/board/mbed_board.py
```diff
@@ -20,7 +20,7 @@
 
 from time import sleep
 from board import Board
-from pyOCD.interface import INTERFACE, usb_backend
+from pyOCD.transport.daplink import DapLink
```
sg-: DAPLink

c1728p9:
Added an exception hierarchy along with restructuring deferred transfers to use a new API.

pyOCD/gdbserver/gdbserver.py
```diff
@@ -16,7 +16,7 @@
 """
 
 from ..target.target import Target
-from ..transport.transport import Transport
+from pyOCD.pyDAPAccess import DAPAccess
```
0xc0170: we are changing from relative imports to absolute?

pyOCD/pyDAPAccess/__init__.py
```diff
@@ -14,8 +14,7 @@
...
  limitations under the License.
 """
+from .dap_access_api import DAPAccessIntf
+from .dap_access_usb import DapAccessUSB
 
-from cmsis_dap import CMSIS_DAP
-
-TRANSPORT = {'cmsis_dap': CMSIS_DAP
-            }
+DAPAccess = DapAccessUSB
```
0xc0170: What's the intention in this assignment ?

pyOCD/pyDAPAccess/dap_access_usb.py
```diff
@@ -0,0 +1,756 @@
...
+            return reg_read_repeat_cb()
+        else:
+            return reg_read_repeat_cb
+    # ------------------------------------------- #
+    #          Private functions
+    # ------------------------------------------- #
+
+    def _init_deferred_buffers(self):
+        """
+        Initailize or reinitalize all the deferred tranfer buffers
```
0xc0170: cosmetic - typos here

pyOCD/tools/pyocd.py
```diff
@@ -648,7 +645,7 @@ def handle_clock(self, args):
         else:
             nice_freq = "%d Hz" % freq_Hz
 
-        print "Changed %s frequency to %s" % (swd_jtag, nice_freq)
+        print "Changed frequency to %s" % nice_freq
```
0xc0170: Can we start using python 3 compatible prints?

pyOCD/gdbserver/gdbserver.py
c1728p9: Only for pyDAPAccess.  Once pyDAPAccess is its own package a replace of "pyOCD.pyDAPAccess" with "pyDAPAccess" will be done.

pyOCD/pyDAPAccess/__init__.py
c1728p9: DAPAccess is a alias to DAPAccessUSB.  That way modules using it only need to call "from pyDAPAccess import DAPAccess".

pyOCD/pyDAPAccess/dap_access_usb.py
c1728p9: Good catch, this must have been a bulk rename mistake.  Will fix.

pyOCD/tools/pyocd.py
c1728p9: I agree, switching to python 3 prints should be done.  Do you mind if this is done in future PR?  I would like to have one big PR which goes through and makes everything python 3 compatible.

tgarc: I agree @c1728p9, this is the same point I was making about the absolute imports

tgarc:
Just a general note: I think it's great to have a higher level Link class to encapsulate the interaction between transport and interface; it should make the Board code more manageable. E.g., you're implementation of ``getAllConnectedBoards`` makes much more sense and *crucially* removes the board class dependency on using a fixed mbed vid/pid. (pid/vid is an interface level descriptor, it doesn't belong in the board class!).

I'm not sure how I feel yet about the actual implementation of the Link class though. My initial feeling is that the transport and interface classes shouldn't be removed; they should instead be wrapped by the Link class. One of my selfish reasons for wanting that is that I eventually wanted to implement the SWD protocol directly (i.e. as an alternative to using cmsis-dap) in pyOCD. To do that, I would have to add an additional SWD transport subclass and an additional Interface subclass for bit banging the protocol (I'm thinking using an ftdi device in MPSSE mode). Would that addition be possible in your framework? Or are you thinking CMSIS-DAP only?

I think this point deserves discussion because the fact that ``transport`` is a separate module in the current implementation tells me that other transport options **should** be available in the future and the transport class **should not** be removed. OTOH, if ARM/pyOCD core developers feel strongly that pyOCD should be a strictly CMSIS-DAP only library :,(  then it seems like ``transport`` is unnecessary and all the interface/protocol functionality could be integrated into a single module (e.g. DAPLink).

c1728p9:
Hi @tgarc , the Link class (now DAPAccessIntf) is essentially the transport class but with a few more functions (like listing connected boards).  Memory access functions were removed from the link class since they are somewhat processor specific (AHB-AP varies between device families), and there is not really a performance gain from keeping them in the transport layer.  The 4 primary dap access functions (in "DAP Access functions" block) are not specific to CMSIS-DAP, just DAP (SWD/JTAG), so you would need to implement those regardless for FTDI.

This new framework should make it easier to implement a custom Link layer since you only need to implement the functions in the Link class.  From there, you should be able to use pyOCD without any further modifications by passing your implementation of DAPAccessIntf into one of the board list/choose functions as the 'dap_class' parameter.  However, the new pyDAPAccess package will likely only have an implementation of the Link layer for CMSIS-DAP (although the Link interface will be generic as mentioned above).  This will keep the size of pyDAPAccess manageable.  Having said that, I see no scalability issue if the FTDI driver is in its own package.

Let me know if you still have concerns about the Link class or additional feedback.  I really appreciate the feedback you have provided so far.

pyOCD/tools/pyocd.py
0xc0170: Tracking issue for python 3 compatibility - #208

c1728p9:
Last call for comments.  This patch will be merged shortly.

PR #211, opened 2015-11-18
added 'init' and 'reset' functionality to gdbserver

tgarc:
I've **not** tested this with the new pyDAP framework so I'd like to do that before this gets merged but it seems to work fine with the non-pyDAP version.

I also restructured the handleRemoteCommand a little bit: 
+ removed comments that were redundant/already clear from the code
+ reordered commands so ``init`` is now 0x1 and the rest of the cmd codes are shifted by 1
+ logic is slightly different so that ``help`` must be entered first for the help command to succeed - anything else is considered malformed

Question: Why do you force the target to halt after running the commands? Seems to defeat the purpose of having a ``resume`` command.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -948,7 +948,6 @@ class GDBServer(threading.Thread):
         }
 
         resp = 'OK'
-        cmdList = cmd.split()
         if cmd == 'help':
             resp = ''.join(['%s\t%s\n' % (k, v[0]) for k, v in safecmd.items()])
             resp = hexEncode(resp)
@@ -957,6 +956,7 @@ class GDBServer(threading.Thread):
             logging.info("Semihosting %s", ('enabled' if self.enable_semihosting else 'disabled'))
         else:
             resultMask = 0x00
+            cmdList = cmd.split()
             if cmdList[0] == 'help':
                 # a 'help' is only valid as the first cmd, and only
                 # gives info on the second cmd if it is valid
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -940,9 +940,9 @@ class GDBServer(threading.Thread):
 
         safecmd = {
             'init'  : ['Init reset sequence', 0x1],
-            'reset' : ['Reset target', 0x2],
+            'reset' : ['Reset and halt the target', 0x2],
             'halt'  : ['Halt target', 0x4],
-            'resume': ['Resume target', 0x8],
+            # 'resume': ['Resume target', 0x8],
             'help'  : ['Display this help', 0x80],
         }
 
@@ -981,16 +981,13 @@ class GDBServer(threading.Thread):
             if (resultMask & 0x6) == 0x6:
                 self.target.resetStopOnReset()
             elif resultMask & 0x2:
-                self.target.reset()
+                # on 'reset' still do a reset halt
+                self.target.resetStopOnReset()                
+                # self.target.reset()
             elif resultMask & 0x4:
                 self.target.halt()
-            if resultMask & 0x8:
-                self.target.resume()
-
-            if self.target.getState() != Target.TARGET_HALTED:
-                logging.error("Remote command left target running!")
-                logging.error("Forcing target to halt")
-                self.target.halt()
+            # if resultMask & 0x8:
+            #     self.target.resume()
 
         return self.createRSPPacket(resp)
 
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -939,58 +939,54 @@ class GDBServer(threading.Thread):
         logging.debug('Remote command: %s', cmd)
 
         safecmd = {
-            'reset' : ['Reset target', 0x1],
-            'halt'  : ['Halt target', 0x2],
-            'resume': ['Resume target', 0x4],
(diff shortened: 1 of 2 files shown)


flit:
Since you're cleaning this up, I'd prefer that you get rid of the use of masks and instead have a proper command parser that recognizes a command with possible options.

Also, why do we need the 'reg' command? It doesn't do anything useful that you can't already do via gdb's own commands.

flit:
I've also wondered why that force halt was at the end of `handleRemoteCommand()`. Unless someone thinks of a reason not to, I think you can get rid of it.

c1728p9:
I added the explicit halt to prevent strange GDB behavior resulting from leaving the target running.  The behavior was especially bad when using eclipse as a front end.  GDB/Eclipse would indicate that the target is halted at a specific line when it was actually running.  Clicking the single step button would make it appear that the program performed a step but ended up at the same address (when really it was running and not performing a step at all).  Here is the commit that added it: bd2edce86bd4d3ad0922433a1ccc8e7e759cb88e

flit:
Interesting. I wonder what is causing GDB/Eclipse to get confused, and whether the cause is in the debug launch plugin (where we can fix it).

Anyway, I guess the check for halt needs to be made selective based on the command. Or we could remove the 'resume' command, not much value in having it.

c1728p9:
I think it was the reset command causing trouble, since based on the eclipse configuration it can send the reset or reset halt command.  It should be correctable in the plugin - only allow halting reset.  The user will just have to be aware that if they send monitor commands manually it will confuse GDB since it won't know the target has started running.

I think it would be find to remove the halt check, as long as the plain reset (0x2) halts the target.

tgarc:
@flit 

I was about to mention the thing about the reg command. I don't know why it's there and I commented out since it's not actually implemented. I agree it seems like this should be removed.

As for ``init``, I'd like other people's input on this. For me it is useful because it would allow me to reinitialize any non-debug components that have been reset or corrupted during execution. Now is it safe to re-init? I would assume so since the CortexM.init() basically just does a DBG/SYSPWRUPREQ and sets some auto-detected parameters. But I can't say for certain that re-running *any* arbitrary init sequence is safe. Again, I'd like to hear what other people have to say.

c1728p9:
I'm fine with having 'reg' removed.

I agree, being able to call init multiple times would be nice.  That wasn't an expected use case so there are no continuous integration tests for this.  Shouldn't be too hard to add them and fixup the code if there are bugs though.  Glancing over the code it looks like both the breakpoints and watchpoints get re-initialized so that seems safe.  There will probably need to be code to close and re-connect the link on re-init, since calling link.connect twice without closing might lead to unexpected behavior.

tgarc:
@c1728p9 
Good catch, I didn't see that ``link.init()`` that was added in the new code. I haven't dived into the unit tests yet so this is probably a good opportunity for me to do that.

About the issues related to leaving the core running: I think a reset-and-resume command is a reasonable thing to do so if we can somehow "let gdb know* that the target has resumed I think that would be the best option. OTOH, having 'reset' and 'reset halt' both perform a reset halt and removing the resume command would be really simple and would remove the possibility of confusing gdb in the first place. And frankly, I can't think of why resume needs to be implemented as a mon command at all if you're running gdb anyway. Is this to support debugging over telnet?

c1728p9:
I'm really sure why you would need a resume monitor command.  Here's the patch that added it:
https://github.com/mbedmicro/pyOCD/commit/e23e36a29f3d41d178176d54e3bef5067d3c14f4

I would have no problem with it being removed.

0xc0170:
> I'm really sure why you would need a resume monitor command. Here's the patch that added it:

Regarding resume, some details were in the pull request: https://github.com/mbedmicro/pyOCD/pull/62#discussion_r18560354

tgarc:
Ok I removed the reg cmd and removed caching of idcode as suggested by c1728p9 to fix error occurring on re-init.

tgarc:
After some thought, I feel I just want to leave the forced halting action as is. I'm removing ``resume`` because it doesn't actually resume. This means ``resume`` doesn't actually resume, but if there's ever a need for it we can revisit the issue (again, all I can think of is telnet). I decided to force ``reset`` to call ``resetStopOnReset`` since even doing a reset followed by a halt - as opposed to setting the VC_CORERESET bit *before* resetting - was causing GDB to stop functioning for me. (target got stuck in an execution state where I couldn't do anything). I feel like if somebody really wants to reset and let the core run they should just disconnect the debugger :D. Anyway, I think GDB is just not set up to handle this type of reset.

c1728p9:
The patch looks good to me.  Tests also pass on this patch.  @tgarc are you ready for this patch to be merged?

tgarc:
@c1728p9 yessir

tgarc:
@flit 

> Since you're cleaning this up, I'd prefer that you get rid of the use of masks and instead have a proper 
> command parser that recognizes a command with possible options.

This patch is really just to add the ``init`` command; writing a parser that takes options for commands is really outside of this scope. Anyway, I don't see a need to write a more complicated parser unless we actually are ready to implement commands that require additional options.

0xc0170:
@tgarc Can you please sign http://developer.mbed.org/contributor_agreement/ ?

0xc0170:
@anpilog You might be interested in this patch (resume/reset changes)

tgarc:
@0xc0170 I signed this back when I did my first PR

0xc0170:
Thanks,

+1

PR #213, opened 2015-11-23
flash_tool.py: Return if there is no board connected

oliviermartin:
Prevent this tool to block if there is no board connected.

### DIFF
--- a/pyOCD/tools/flash_tool.py
+++ b/pyOCD/tools/flash_tool.py
@@ -146,7 +146,10 @@ def main():
         MbedBoard.listConnectedBoards()
     else:
         board_selected = MbedBoard.chooseBoard(board_id=args.board_id, target_override=args.target_override,
-                                               frequency=args.frequency)
+                                               frequency=args.frequency, blocking=False)
+        if board_selected is None:
+            print("Error: There is no board connected.")
+            sys.exit(1)
         with board_selected as board:
             flash = board.flash
             link = board.link


c1728p9:
Thanks for the pull request.  Instead of making this a command line option, it might be better just to make it always non-blocking.  Not sure if anyone else has preferences on this, but I don't see much of a use in having flash tool block.

0xc0170:
>  it might be better just to make it always non-blocking

What do you think @oliviermartin ? You can always write blocking on top of it.

oliviermartin:
For me making the tool non blocking it is even better because I do not need to change my script that invokes this tool!

0xc0170:
LGTM

pyOCD/tools/flash_tool.py
```diff
@@ -147,6 +147,8 @@ def main():
     else:
         board_selected = MbedBoard.chooseBoard(board_id=args.board_id, target_override=args.target_override,
```
c1728p9: I think chooseBoard is blocking by default, so you'll need to set the 'blocking' parameter to false.

c1728p9:
Hi @oliviermartin , have you had a chance to look at my comment?

pyOCD/tools/flash_tool.py
oliviermartin: You are right, I accidentally removed this change in my previous update. The new version should restore the initial parameter setting.

c1728p9:
test this please

janjongboom:
+1 but I want an error message too

c1728p9:
Hi @oliviermartin , could you rebase this commit on top of master and add an error message?  Once it is rebased I'll run CI testing on it and merge it.

c1728p9:
test this please

oliviermartin:
Tested-By: Olivier Martin <olivier.martin@labapart.com>

With no board connected:

```
$ ./flash_tool.py 
No connected boards
Error: There is no board connected.
```

With KL25Z connected:

```
$ ./flash_tool.py -l
0 => FRDM-KL25Z [kl25z] boardId => 02000203D62D2E6C2BD1D394
```

With KL25Z connected:

```
$ ./flash_tool.py -t kl25z -ce
INFO:root:DAP SWD MODE initialised
INFO:root:KL25Z not in secure state
INFO:root:2 hardware breakpoints, 0 literal comparators
INFO:root:CPU core is Cortex-M0+
INFO:root:2 hardware watchpoints
Erasing chip...
Done
```

With no board connected:

```
$ ./flash_tool.py -t kl25z -ce
No connected boards
Error: There is no board connected.
```

c1728p9:
The phrase "test this please"  triggers the CI system to run tests.  Wanted to let you know so it doesn't seem like I'm being demanding asking you to run tests :smile:.Thanks for the testing though.

c1728p9:
Thanks for the PR @oliviermartin

PR #214, opened 2015-11-25
Deleted use of non-existing RESET register. Only present in nRF51 dev…

David-Garcia-Polo:
As can be seen in the nRF52 specifications (can be found at http://infocenter.nordicsemi.com/index.jsp) POWER.RESET register does not exist in nRF52.


0xc0170:
Reason for closing?

David-Garcia-Polo:
Even though the pull is correct, pin reset in nRF52 and nRF51 is not working properly. I will provide a new pull today with more changes.

PR #215, opened 2015-11-28
Fixed error in MemoryRegion.containsRange() function.

David-Garcia-Polo:
Fixed error. When both end and length parameters to MemoryRegion.containsRange() are not given, None was used in an arithmetic expression.

### DIFF
--- a/pyOCD/target/memory_map.py
+++ b/pyOCD/target/memory_map.py
@@ -89,6 +89,8 @@ class MemoryRegion(object):
 
     def containsRange(self, start, end=None, length=None):
         if end is None:
+            if length is None:
+                raise ValueError("both end and length parameters are None")
             end = start + length - 1
         return self.containsAddress(start) and self.containsAddress(end)
 
--- a/pyOCD/target/memory_map.py
+++ b/pyOCD/target/memory_map.py
@@ -88,9 +88,8 @@ class MemoryRegion(object):
         return (address >= self.start) and (address <= self.end)
 
     def containsRange(self, start, end=None, length=None):
+        assert (end is not None) ^ (length is not None)
         if end is None:
-            if length is None:
-                raise ValueError("both end and length parameters are None")
             end = start + length - 1
         return self.containsAddress(start) and self.containsAddress(end)
 


David-Garcia-Polo:
I am not certain of the appropriate way you take pulls. Do you prefer lots of small pulls, or one big one? Also, do you accept corrections of unnecessary code that is not harmful but it is just useless (example in CortexM class reset() method default parameter as None, since that trick is only needed when the parameter is a mutable object)

c1728p9:
Hi @David-Garcia-Polo , thanks for the pull request!  Code corrections are appreciated even when the code being changed is harmless.  For stylistic changes, try to stick with pep8/pylint.

Pull request size is fairly open ended.  It's a little nicer to have smaller pull requests but it's not a requirement.  It is helpful if each commit within a pull request is a single logical change grouped together.  If you do end up creating several pull requests, it is encouraged to rebase each pull request on top of the previously merged pull request.  That way there is a more linear commit history.

Finally, you mention that the default parameter for reset is set to None.  I think the purpose of this was to allow additional information to be passed to reset, allowing it to determine the type of reset that is required - soft (True), hard (False) or either/unspecified (None).  For example if resetStopOnReset, which calls reset, had software_reset=True by default then reset would see that a software reset is required, even if it hasn't been explicitly specified by the caller of resetStopOnReset.

pyOCD/target/memory_map.py
```diff
@@ -89,6 +89,8 @@ def containsAddress(self, address):
 
     def containsRange(self, start, end=None, length=None):
         if end is None:
+            if length is None:
+                raise ValueError("both end and length parameters are None")
```
c1728p9: Good idea to check for both missing.  It might also be good to check if both are present and if so trigger an error.  An assert should make this easy.  If you want to leave it as a ValueError that works too.  The patch looks good as is.  Just wanted to mention it.
```
assert (end is not None) ^ (length is not None)
```

David-Garcia-Polo:
I agree with you that an assert will be more convenient to check parameter correctness. Updated pull as you suggested.

0xc0170:
@David-Garcia-Polo can you please sign http://developer.mbed.org/contributor_agreement/ ?

David-Garcia-Polo:
You accepted the Contributor Agreement on Mon 30 Nov 2015.

I have accepted with another name, for some reason David-GArcia-Polo was not acceptable. The name is Advenedizo. Will it cause problems?

0xc0170:
+1

c1728p9:
Sorry for the delay.  Nope it won't cause problems.  Patch looks good

PR #216, opened 2015-12-07
Architectural refactoring

flit:
Fairly significant changes to re-architect pyOCD's target and Cortex-M related code.

- Created new pyOCD.coresight submodule. It has DebugPort and AccessPort classes, as well as classes for CortexM, FPB, DWT, breakpoints, and ROM table parsing.
- CortexM class moved to pyOCD.coresight and simplified by using other coresight component classes such as AHB_AP and the breakpoints classes.
- New CoreSightTarget subclass of Target that owns the DebugPort, AccessPort, and CortexM classes. It supports multiple cores. The existing target classes are now subclasses of CoreSightTarget rather than CortexM.
- Factored breakpoints code that was in CortexM into a BreakpointsManager class and related classes.
- CoreSight ROM tables are parsed during init, although they are not yet used to construct objects.
- Additional CPU states are supported: reset, lockup, and sleep.
- Debug resources are disabled and debug is powered down upon disconnect.
- Memory reads are filtered to hide software breakpoints.

This changeset also includes some miscellaneous changes, as well:
- pyocd-tool can execute arbitrary Python expressions via a "$<python code>" syntax.
- Added "core" command to pyocd-tool to select the current core.
- SVD files are associated with targets and the registers can be accessed via pyocd-tool.
- hidapi USB backend was improved to fix issues with multiple connected boards and when another process already has a board open.

PR #219, opened 2015-12-15
Proposal: DAP_Transfer logging

tgarc:
At some point over the last couple of months while trying to get pyOCD working for a chip that is still under development, I wrote a little module that attaches to the `interface` object (which is now under pyDAPAccess), that would do some interpretation of the CMSIS `DAP Transfer` and `DAP Transfer Block` commands. The point was to try and expose what was being sent to the debug port on a lower level without inlining logging messages in the master source code. Here's a snippet of output logging of the board initialization process::

```
DEBUG:root:closing interface
DEBUG:root:receive:dap_info (len=0x1, info=0xa)
DEBUG:root:init board <pyOCD.board.mbed_board.MbedBoard object at 0x7fd3cf6d4f10>
DEBUG:root:receive:dap_swj_clock (resp=DAP_OK)
DEBUG:root:cortexm init
DEBUG:root:receive:dap_connect (resp=SWD)
INFO:root:DAP SWD MODE initialised
DEBUG:root:receive:dap_swj_clock (resp=DAP_OK)
DEBUG:root:receive:dap_transfer_configure
DEBUG:root:receive:dap_swd_configure
DEBUG:root:receive:dap_swj_sequence
DEBUG:root:receive:dap_swj_sequence
DEBUG:root:receive:dap_swj_sequence
DEBUG:root:receive:dap_swj_sequence
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.IDCODE  => 0x2BA01477
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.ABORT   <= 0x00000004
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.SELECT  <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.CTRL/STAT <= 0x50000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.CTRL/STAT => 0xF0000040
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.CTRL/STAT <= 0x50000F00
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.SELECT  <= 0x000000F0
WARNING:root:Attempting to access memory without first setting TAR
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.DRW     => 0x24770011
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DP.SELECT  <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.CSW     <= 0x23000052
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDF0
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DHCSR      <= 0xA05F0003 ; S_RESET_ST=0, S_RETIRE_ST=0, C_HALT=1, C_STEP=0, S_LOCKUP=0, C_DEBUGEN=1, C_MASKINTS=0, S_REGRDY=0, S_HALT=0, C_SNAPSTALL=0, S_SLEEP=0
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002000 => 0x00000260
INFO:root:6 hardware breakpoints, 4 literal comparators
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002000 <= 0x00000002
DEBUG:root:fpb has been disabled
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002008
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002008 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000200C
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE000200C <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002010
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002010 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002014
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002014 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0002018
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0002018 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000201C
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE000201C <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000ED00
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE000ED00 => 0x412FC231
INFO:root:CPU core is Cortex-M3
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DEMCR      => 0x01000400 ; TRCENA=1, MON_REQ=0, MON_STEP=0, VC_CORERESET=0, VC_CHKERR=0, VC_STATERR=0, VC_BUSERR=0, VC_MMERR=0, VC_INTERR=0, MON_EN=0, VC_NOCPERR=0, MON_PEND=0, VC_HARDERR=1
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DEMCR      <= 0x01000400
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0001000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0001000 => 0x40000000
INFO:root:4 hardware watchpoints
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0001028
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0001028 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0001038
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0001038 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0001048
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0001048 <= 0x00000000
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE0001058
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xE0001058 <= 0x00000000
DEBUG:root:testchip init
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDF0
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DHCSR      => 0x02010001 ; S_RESET_ST=1, C_HALT=0, S_LOCKUP=0, S_HALT=0, S_SLEEP=0
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDF0
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DHCSR      <= 0x02010001
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DEMCR      => 0x01000400
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        DEMCR      <= 0x01000401 ; VC_CORERESET=1
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xFFFF1014
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xFFFF1014 => 0x4C1D2607
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xFFFF1014
DEBUG:root:receive:dap_transfer (xfer_cnt=0x01, resp=OK, swderr=0, vmismatch=0)
        0xFFFF1014 <= 0x4C1D2607
DEBUG:root:receive:dap_transfer (xfer_cnt=0x02, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
        DEMCR      => 0x01000401
DEBUG:root:receive:dap_transfer (xfer_cnt=0x04, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
        DEMCR      <= 0x01000401
        AP.TAR     <= 0xE000EDFC
        DEMCR      => 0x01000401
DEBUG:root:receive:dap_transfer (xfer_cnt=0x04, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
        DEMCR      <= 0x01000400 ; VC_CORERESET=0
        AP.TAR     <= 0xE000EDF0
        DHCSR      => 0x02010001
INFO:root:Telnet: server started on port 4444
INFO:root:GDB server started at port:3333    
```

The highpoints:
- Any contiguous lines that are indented are part of a single transfer, with the "header" of the dap command preceding it. 
- When writing to any of the debug registers, there is a comment added to the right of the line preceded by a `;` that shows the value of some of the bit fields. The code is configured right now so that only bits that have changed since the last read/write are displayed.
- For other non-transfer commands, only a header is shown

I even added a cute little SWD translator that would give the equivalent bit-level SWD command :)

```
DEBUG:root:receive:dap_transfer (xfer_cnt=0x04, resp=OK, swderr=0, vmismatch=0)
        AP.TAR     <= 0xE000EDFC
        11010001 100 001111111011011100000000000001111
        DEMCR      <= 0x01000400 ; VC_CORERESET=0
        11011101 100 000000000010000000000000100000000
        AP.TAR     <= 0xE000EDF0
        11010001 100 000011111011011100000000000001111
        DHCSR      => 0x02010001
        11111001 100 100000000000000010000000010000001
```

The nice thing about this code is its non-invasiveness. It reads things that are on the way out to the USB but it never actually writes or reads the interface. And, it only changes (adds) 4 lines to the master source code.

So, does this seem like useful functionality? I'm not married to what the output formatting is right now (I could even exclude all the TAR writes to reduce clutter); so I'm totally open to suggestions.

### DIFF
--- a/pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
+++ b/pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
@@ -1,0 +1,464 @@
+import logging
+from pyOCD.pyDAPAccess.cmsis_dap_core import COMMAND_ID, ID_INFO
+from pyOCD.target.dap import AP_REG, CSW_SIZE, CSW_ADDRINC
+from transport_analyzer import TransportAnalyzer
+from collections import deque
+
+DP_RREG = {  0x00: 'IDCODE'
+           , 0x04: 'CTRL/STAT'
+           , 0x08: 'RESEND'
+           , 0x0C: 'RDBUFF' }
+DP_WREG = {  0x00: 'ABORT'
+           , 0x04: 'CTRL/STAT'
+           , 0x08: 'SELECT' }
+
+DAP_MODE = {  0x0: 'DAP_DEFAULT_PORT'
+            , 0x1: 'SWD'
+            , 0x2: 'JTAG' }
+
+TRANSFER_RESP = {  0b000: 'ACK'
+                 , 0b001: 'OK'
+                 , 0b010: 'WAIT'
+                 , 0b100: 'FAULT'}
+
+CORE_REGS = {  0x00: 'r0'
+             , 0x01: 'r1'
+             , 0x02: 'r2'
+             , 0x03: 'r3'
+             , 0x04: 'r4'
+             , 0x05: 'r5'
+             , 0x06: 'r6'
+             , 0x07: 'r7'
+             , 0x08: 'r8'
+             , 0x09: 'r9'
+             , 0x0a: 'r10'
+             , 0x0b: 'r11'
+             , 0x0c: 'r12'
+             , 0x0d: 'sp'
+             , 0x0e: 'lr'
+             , 0x0f: 'pc'
+             , 0x10: 'xpsr'
+             , 0x11: 'msp'
+             , 0x12: 'psp'
+             , 0x14: 'spmcr'}
+
+REG_MAP = {  0xE000EDFC: 'DEMCR'
+           , 0xE000EDF0: 'DHCSR'
+           , 0xE000EDF4: 'DCRSR'
+           , 0xE000EDF8: 'DCRDR' 
+           , 0xE000ED0C: 'AIRCR'}
+
+MEM_MAP = {   (0xE0040000, 0xE0041000): 'TPIU'
+            , (0xE0041000, 0xE0042000): 'ETM'
+            , (0xE0042000, 0xE00FF000): 'EPPB'
+            , (0xE00FF000, 0xE00FFFFF): 'ROM_TABLE'
+            , (0xE0000000, 0xE0001000): 'ITM'
+            , (0xE0001000, 0xE0002000): 'DWT'
+            , (0xE0002000, 0xE0003000): 'FPB'            
+            , (0xE000E000, 0xE000F000): 'SCS'}
+
+BIT_MAPS = \
+{  
+  'DEMCR': 
+    {   
+          0x01000000: ('TRCENA',       24)
+        , 0x00080000: ('MON_REQ',      19)
+        , 0x00040000: ('MON_STEP',     18)
+        , 0x00020000: ('MON_PEND',     17)
+        , 0x00010000: ('MON_EN',       16) 
+        , 0x00000400: ('VC_HARDERR',   10)
+        , 0x00000200: ('VC_INTERR',     9)
+        , 0x00000100: ('VC_BUSERR',     8)
+        , 0x00000080: ('VC_STATERR',    7)
+        , 0x00000040: ('VC_CHKERR',     6)
+        , 0x00000020: ('VC_NOCPERR',    5)
+        , 0x00000010: ('VC_MMERR',      4)
+        , 0x00000001: ('VC_CORERESET',  0)
+    }
+, 'DFSR':
+    {
+        0x00000010: ('EXTERNAL',      4)
+      , 0x00000008: ('VCATCH',        3)
+      , 0x00000004: ('DWTTRAP',       2)
+      , 0x00000002: ('BKPT',          1)
+      , 0x00000001: ('HALTED',        0)
+    }
+, 'DCRSR': 
+    {   
+        0x00010000: ('REGWnR',       16)
+      , 0x0008001F: ('REGSEL',        0)
+    }
+, 'DCRDR':
+    {
+        0xFFFFFFFF: ('DBGTMP',        0)
+    }
+, 'DHCSR':
+    {
+        0x02000000: ('S_RESET_ST',   25)
+      , 0x01000000: ('S_RETIRE_ST',  24) 
+      , 0x00080000: ('S_LOCKUP',     19)
+      , 0x00040000: ('S_SLEEP',      18)
+      , 0x00020000: ('S_HALT',       17)
+      , 0x00010000: ('S_REGRDY',     16)
+      , 0x00000020: ('C_SNAPSTALL',   5)
+      , 0x00000008: ('C_MASKINTS',    3)
+      , 0x00000004: ('C_STEP',        2)
+      , 0x00000002: ('C_HALT',        1)
+      , 0x00000001: ('C_DEBUGEN',     0)
+    }
+, 'AIRCR': 
+    {   
+        0x00008000: ('ENDIANNESS',   15)
+      , 0x00000700: ('PRIGROUP',      8)
+      , 0x00000004: ('SYSRESETREQ',   2)
+      , 0x00000002: ('VECTCLRACTIVE', 1) 
+      , 0x00000001: ('VECTRESET',     0)
+    }
+}
+
+
+RESP_STATUS = {  0xff: 'DAP_ERROR'
+               , 0x00: 'DAP_OK'}
+CONN_RESP = {0x0: 'FAILED', 0x1: 'SWD', 0x2: 'JTAG'}
+
+invmap = lambda d: {v:k for k,v in d.iteritems()}
+AP_REG = invmap(AP_REG)
+COMMAND_ID = invmap(COMMAND_ID)
+ID_INFO = invmap(ID_INFO)
+
+lookup_code = lambda d,k: d.get(k, "UNKNOWN_CODE(0x%02x)" % k)
+lookup_cmd = lambda cmd: COMMAND_ID.get(cmd, "UNKNOWN_CMD(0x%02x)" % cmd)
+reverse_bits = lambda b,width=0: bin(b)[2:].zfill(width)[::-1]
+
+
+def get_word(data,index=0,pop=False):
+    d1,d2,d3,d4 = data[index:index+4]
+    if pop: del data[index:index+4]
+    return d4 << 24 | d3 << 16 | d2 << 8 | d1
+
+class CMSIS_DAPAnalyzer(TransportAnalyzer):
+    def __init__(self, interface):
+        super(CMSIS_DAPAnalyzer, self).__init__(interface)
+        self.last_rqst = deque()
+        self.last_blktransfer = []
+        self._cache = {}
+
+    def default_trace(self, data):
+        ncols = 16
+        nlines, r = divmod(len(data),ncols)
+        
+        lines = ''
+        if nlines > 0: 
+            lines += '\n\t' + '\n\t'.join([' '.join(['0x{:02x}']*ncols)]*nlines)
+        if r > 0: 
+            lines += '\n\t' + ' '.join(['0x{:02x}']*r)
+
+        return lines.format(*data)
+
+    def get_rqst_translater(self, cmd):
+        return getattr(self, cmd.lower()+'_rqst', None)
+
+    def get_resp_translater(self, cmd):
+        return getattr(self, cmd.lower()+'_resp', None)
+
+    def trace_write(self, data, **kwargs):
+        cmd = lookup_cmd(data[0])
+        if cmd == 'DAP_TRANSFER':
+            (dapidx, tcnt), xfers = data[1:3], data[3:]
+
+            header = " (dap_idx=0x{:02x}, xfer_cnt=0x{:02x})".format(dapidx,tcnt)
+            trace = header + self._xrqst(tcnt, xfers)
+        elif cmd == 'DAP_TRANSFER_BLOCK':
+            dapidx, (tcntL, tcntH), treq, xfers = data[1], data[2:4], data[4], data[5:]
+            tcnt = tcntH << 8 | tcntL
+
+            header = " (dap_idx=0x{:02x}, xfer_cnt=0x{:02x})".format(dapidx,tcnt)
+            trace = header + self._xrqst_block(tcnt, treq, xfers)
+        else:
+            tracer = self.get_rqst_translater(cmd)
+            trace = tracer(data) if tracer is not None else ""
+
+        # logging.debug(("trace_tx:%s" % cmd.lower()) + trace)
+
+    def trace_read(self, data, **kwargs):
+        cmd = lookup_cmd(data[0])
+        if cmd == 'DAP_TRANSFER':
+            tcnt, tresp, xfers = data[1], data[2]&0x1F, data[3:]
+            swderr = (tresp&0b01000) >> 3
+            vmm    = (tresp&0b10000) >> 4
+
+            header = " (xfer_cnt=0x{:02x}, resp={:s}, swderr={:d}, vmismatch={:d})"
+            header = header.format(tcnt,lookup_code(TRANSFER_RESP,tresp),swderr,vmm)
+            trace = header + self._xresp(tcnt, tresp, xfers)
+        elif cmd == 'DAP_TRANSFER_BLOCK':
+            (tcntL,tcntH), tresp, xfers = data[1:3], data[3]&0xF, data[4:]
+            tcnt = tcntH << 8 | tcntL
+            swderr = (tresp&0b01000) >> 3
+
+            header = " (xfer_cnt=0x{:02x}, resp={:s}, swderr={:d})"
+            header = header.format(tcnt,lookup_code(TRANSFER_RESP,tresp),swderr)
+            trace = header + self._xresp_block(tcnt, tresp, xfers)
+        else:
+            tracer = self.get_resp_translater(cmd)
+            trace = tracer(data) if tracer is not None else ""
+
+        logging.debug(("trace_rx:%s" % cmd.lower()) + trace)
+        
+    def dap_response(self, data):
+        return " (resp=%s)" % lookup_code(RESP_STATUS, data[1])
+
+    def _xrqst(self, tcnt, xfers):
+        # for each block transfer, a single transfer request byte is sent that
+        # applies to all data sent/received
+        lines = ''
+        for i in range(tcnt):
+            lines += '\n\t'
+            treq = xfers.pop(0)&0x3F
+
+            apndp = (treq&0b000001)
+            rnw =   (treq&0b000010) >> 1
+            a23 =   (treq&0b001100)
+            vm =    (treq&0b010000) >> 4
+            mm =    (treq&0b100000) >> 5
+
+            dapregmap = AP_REG if apndp else DP_RREG if rnw else DP_WREG
+            dapreg = ('AP' if apndp else 'DP') + '.' + dapregmap[a23]
+
+            data = None
+            lines += ('READ ' if rnw else 'WRITE') + ' ' + dapreg
+            if not rnw:
+                dword = get_word(xfers, pop=True)
+                lines += ", 0x{:08X}".format(dword)
+                data = dword
+
+                if dapregmap[a23] == 'TAR':
+                    lines += " ;  TAR = &%s" % REG_MAP.get(dword,"0x{:08X}".format(dword))
+                elif dapregmap[a23] == 'DRW':
+                    lines += " ; *TAR = 0x{:08X}".format(dword)
+            elif dapregmap[a23] == 'DRW':
+                lines += " "*12 + " ;  DRW = *TAR"
+
+            if dapregmap[a23] == 'TAR':
+                regs = sorted([(a,b) for (a,b) in MEM_MAP if a <= dword < b])
+                if regs:
+                    base = regs[-1]
+                    lines += " {%s+%d}" % (".".join(map(MEM_MAP.get,regs)),dword-base[0])
+
+            if vm != 0 or mm != 0:
+                lines += " (vmatch={:d}, mmask={:d})".format(vm,mm)
+
+            self.last_rqst.append([treq, dapreg, data])
+
+        return lines
+
+    def _xrqst_block(self, tcnt, treq, xfers):
+        # for each transfer, a single transfer request byte is sent and
+        # optionally a set of data words to write
+        apndp = (treq&0b000001)
+        rnw =   (treq&0b000010) >> 1
+        a23 =   (treq&0b001100)
+
+        dapregmap = AP_REG if apndp else DP_RREG if rnw else DP_WREG
+        dapreg = ('AP' if apndp else 'DP') + '.' + dapregmap[a23]
+        lines = ''
+
+        data = []
+        for i in range(tcnt):
+            lines += '\n\t' + ('READ ' if rnw else 'WRITE') + ' ' + dapreg
+            if not rnw:
+                dword = get_word(xfers, pop=True)
+                data.append(dword)
+
+                lines += ", 0x{:08X}".format(dword)
+                if dapregmap[a23] == 'DRW':
+                    lines += " ; *TAR++ = 0x{:08X}".format(dword)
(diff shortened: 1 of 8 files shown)


flit:
I actually added something very much like this to my architectural refactoring pull request #216. It's a bit higher level, though.

tgarc:
@flit oh yeah I remember seeing something about that in your PR. Honestly though with 8 months of commits all stuffed into a PR I didn't look at anything with too much detail. I'll try and take a closer look when I get a minute.

0xc0170:
@flit  can you point out to the specific commit ?

@tgarc Thanks for sharing. Some ppl will definitely find this helpful ! Does this need to belong here, or can be provided by a separate module?

tgarc:
@0xc0170 This module can be moved wherever, it can exist at the top level of pyOCD if you want. The only requirement is that it be allowed to attach to the usb ``Interface`` somewhere along the board initialization chain (which is why I put it adjacent to the interface module). Where it attaches right now is probably not ideal::

```.diff
diff --git a/pyOCD/pyDAPAccess/cmsis_dap_core.py b/pyOCD/pyDAPAccess/cmsis_dap_core.py
index bba1bd7..b46aadc 100644
--- a/pyOCD/pyDAPAccess/cmsis_dap_core.py
+++ b/pyOCD/pyDAPAccess/cmsis_dap_core.py
@@ -79,6 +79,10 @@ class CMSIS_DAP_Protocol(object):
     def __init__(self, interface):
         self.interface = interface
 
+        if logging.getLogger().isEnabledFor(logging.DEBUG):
+            from analyzer import ANALYZER
+            ANALYZER['cmsis_dap'].attach(interface)
+
     def dapInfo(self, id_):
         cmd = []
         cmd.append(COMMAND_ID['DAP_INFO'])
```

tgarc:
Sorry, I had to bomb that branch and force push in order to remove a past commit.

c1728p9:
Awesome work and thanks for sharing!  I pulled this down locally and gave it a try.  I can see each packet sent and info about it.  This could be a very useful tool for debugging and verifying changes made to the cmsis-dap driver.

c1728p9:
As an alternative to attaching the logger explicitly, what do you think of sending debug messages on each cmsis-dap read/write and using your analyzer code as a logging formatter?

pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
```diff
@@ -0,0 +1,464 @@
...
+
+                if dapregmap[a23] == 'TAR':
+                    lines += " ;  TAR = &%s" % REG_MAP.get(dword,"0x{:08X}".format(dword))
+                elif dapregmap[a23] == 'DRW':
+                    lines += " ; *TAR = 0x{:08X}".format(dword)
+            elif dapregmap[a23] == 'DRW':
+                lines += " "*12 + " ;  DRW = *TAR"
+
+            if dapregmap[a23] == 'TAR':
+                regs = sorted([(a,b) for (a,b) in MEM_MAP if a <= dword < b])
```
c1728p9: I get an exception here because 'dword' is undefined:
```
UnboundLocalError: local variable 'dword' referenced before assignment
```

The command I used to cause this is:
```
pyocd-gdbserver -d trace
```

pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
```diff
@@ -0,0 +1,464 @@
...
+            dapidx, (tcntL, tcntH), treq, xfers = data[1], data[2:4], data[4], data[5:]
+            tcnt = tcntH << 8 | tcntL
+
+            header = " (dap_idx=0x{:02x}, xfer_cnt=0x{:02x})".format(dapidx,tcnt)
+            trace = header + self._xrqst_block(tcnt, treq, xfers)
+        else:
+            tracer = self.get_rqst_translater(cmd)
+            trace = tracer(data) if tracer is not None else ""
+
+        # logging.debug(("trace_tx:%s" % cmd.lower()) + trace)
```
c1728p9: Any reason TX isn't printed?

tgarc:
@c1728p9 Russ, thanks for taking the time to give this a try! Could you explain more about what your proposing about logging each dap read/write? Not sure if I understand you correctly, but I do the explicit attachment so as to avoid running through all the debug logic and extra function calls when the trace functionality is not being used. I would log it all the time but the debugging messages are a lot more than just dumping the data.

pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
tgarc: @c1728p9 Mostly to reduce clutter. The transmit messages are very much similar to the receive messages, except they doesn't include any actual read data or whether the transmission succeeded, so they are pretty much redundant. What would be useful however is to store these logging messages and use them as a sort of partial traceback for a failed transmission.

flit:
Fwiw, here's the commit where I added DAP logging: 6f8199ea494277e652247e8ac357189819be979c

My version is slightly higher level. It is done in the DebugPort and MEM_AP classes. It shows the start and end of readMem and writeMem operations, so you can see which individual DP/AP accesses are part of the memory access. It also reports cached accesses that didn't actually go across SWD.

Here's an example of the output:
`````
0000000260ms:INFO:dap.ap0:writeMem:000021 (addr=0xe000edf0, size=32) = 0xa05f0001 {
0000000260ms:INFO:dap:writeDP:000023 (addr=0x00000002) = 0x00000000
0000000264ms:INFO:dap:writeAP:000022 (addr=0x00000000) = 0x23000052
0000000267ms:INFO:dap:writeDP:000025 cached (addr=0x00000002) = 0x00000000
0000000267ms:INFO:dap:writeAP:000024 (addr=0x00000004) = 0xe000edf0
0000000270ms:INFO:dap:writeDP:000027 cached (addr=0x00000002) = 0x00000000
0000000270ms:INFO:dap:writeAP:000026 (addr=0x0000000c) = 0xa05f0001
0000000273ms:INFO:dap.ap0:writeMem:000021 }
0000000274ms:INFO:dap:writeDP:000029 (addr=0x00000002) = 0x01000000
0000000277ms:INFO:dap:writeAP:000028 (addr=0x01000004) = 0x00000004
0000000290ms:INFO:dap:writeDP:000031 cached (addr=0x00000002) = 0x01000000
0000000292ms:INFO:dap:readAP:000030 (addr=0x01000000) -> 0x0089003a
0000000293ms:INFO:dap:writeDP:000033 cached (addr=0x00000002) = 0x01000000
0000000293ms:INFO:dap:writeAP:000032 (addr=0x01000004) = 0x00000000
0000000298ms:INFO:dap.ap0:readMem:000034 (addr=0xe000edf0, size=32) {
0000000298ms:INFO:dap:writeAP:000035 cached (addr=0x00000000) = 0x23000052
0000000299ms:INFO:dap:writeDP:000037 (addr=0x00000002) = 0x00000000
0000000302ms:INFO:dap:writeAP:000036 (addr=0x00000004) = 0xe000edf0
0000000306ms:INFO:dap:writeDP:000039 cached (addr=0x00000002) = 0x00000000
0000000310ms:INFO:dap:readAP:000038 (addr=0x0000000c) -> ...
0000000310ms:INFO:dap:readAP:000038 ...(addr=0x0000000c) -> 0x01030003
0000000310ms:INFO:dap.ap0:readMem:000034 (addr=0xe000edf0, size=32) -> 0x01030003 }
`````

It uses a logging MemoryHandler with a FileHandler as the target, which writes to `$(PWD)/pyocd_dap.log`. It is enabled by setting LOG_DAP in pyOCD.coresight.dap.py to True (like the debug logging options in gdbserver.py).

**Comment by c1728p9** [comment]
Hi @tgarc , look like what I was proposing won't work.  I was under the impression that you could call logging functions on raw data (like a 64 byte cmsis-dap command) and have a logging handler process it into a usable format.  Turns out this isn't really the case.

**Comment by c1728p9** [comment]
This patch adds the additional logging level 'trace' to enable logging info about each packet.  Seems like the possible places to log has grown large enough that some kind of filtering command line argument might be more appropriate than a new level.  That way the logging for every module could be very verbose, and the logging for a given module would only need to be turned to 'info' or 'debug' when it is being developed, or a bug is being investigated.

I was thinking maybe a command line option for the tools to either select between a few pre-defined logging configurations or to pass in a logging.conf file or both.  Similar to what is described here:
https://docs.python.org/2/howto/logging.html#configuring-logging

Let me know what you think, or if you have any other ideas.

**Comment by 0xc0170** [comment]
> This patch adds the additional logging level 'trace' to enable logging info about each packet. Seems like the possible places to log has grown large enough that some kind of filtering command line argument might be more appropriate than a new level. 

Is having trace level useful ? I consider debug level to cover this. However, it would be nice to have logger hierarchy in our codebase. That could be done outside of this PR. Therefore we should create a new issue to discuss this. I checked the commit referenced by Chris, there's a dap logger added.

Back to this, is trace level != debug level? @tgarc ?

**Comment by tgarc** [comment]
Sorry I went AFK for awhile there, just enjoying the holidays :) Happy new year!

@0xc0170 trace level is one level above/below debug level so setting trace logging level enables output of both debug messages and trace output. On another note, if not through setting the log level, what would you propose for enabling this functionality?

@c1728p9
What do you mean by your comment: 

>  I was under the impression that you could call logging functions on raw data (like a 64 byte cmsis-dap command) and have a logging handler process it into a usable format.

This is essentially what the logger does; it takes the raw cmsis-dap data and interprets it. The only thing it knows is whether the data is outgoing or incoming. It's configured right now so that it hooks into the read and write of the interface object, but this can be changed so it is user callable. I think that's up for debate right now. I will say though that without attaching to the read/write interface, it can't track the history of read/writes and cache values which reduces the functionality somewhat.

> I was thinking maybe a command line option for the tools to either select between a few pre-defined logging configurations or to pass in a logging.conf file or both.

So would this mean having multiple logging handlers that could be enabled through the command line? So, for example, user could pass in ``--logger-conf=trace,gdb`` which would (hypothetically) enable some pre-defined gdb and trace logger handler?

**Comment by tgarc** [comment]
@flit 

> Fwiw, here's the commit where I added DAP logging: 6f8199e

Yeah, that functionality is very much the same. I would say I prefer my approach of having all the logging code in a separate module though since it would be more maintainable and avoids the clutter of inlining logging messages. However, I think having a separate logging handler like the one you added is a good idea. Other than that and the output format the only difference seems to be a few minor features that are included in your code:

i. Shows the access size
ii. Shows 'cached' reads
iii. Shows whether an access is deferred or not

I can't do any of these in my approach, but like you said, my implementation is a bit lower-level and is only intended to show literally what was transmitted to/from the usb driver.

**Comment by c1728p9** [comment]
Hi @tgarc , Happy New Year to you as well!  I hope you enjoyed your time off.  As for your questions:

What I meant by logging raw data is logging raw CMSIS-DAP packets to a logger object, and then adding your code as a handler to that object.  The reason this isn't possible is because logger objects require everything to be string formatted.
```
logger_cmsis_tx.info(cmsis_packet)
...
logger_cmsis_tx.addHandler(cmsis_dap_analyzer)
```

As for enabling the different logging domains, I was thinking use module naming to enable them.  That way none of the loggers would need to be pre-defined.  You could globally enable logging by specifying the top level package name, or you could get as specific as you want.  Something like
``<package>.<folder>.<module>.<module specific>``.

The command line options would look something like this (either or both):
``--logger-conf=pyDAPAccess.cmsis_dap,pyOCD.gdbserver``
``--logger-conf-file=logging.conf``

**Comment by tgarc** [comment]
@c1728p9 

> What I meant by logging raw data is logging raw CMSIS-DAP packets to a logger object, and then adding your code as a handler to that object. The reason this isn't possible is because logger objects require everything to be string formatted.

I could actually modify the code so you could call it that way. The trick is to pass the logger an object containing the data that implements a ``__str__`` method as suggested in the docs here:

https://docs.python.org/2/howto/logging-cookbook.html#implementing-structured-logging

So the eventual call would look something like:

```
from cmsis_dap_analyzer import DAPMessage

# Initialize the custom object once
dapMessage = DAPMessage()

# pass the object the data on each call
logging.debug("dap_transmit: %s", dapMessage(data))
```

This might actually not be a bad way to go since (I believe) that the logger won't actually call ``__str__`` unless it is enabled for that level. Eventually my code could also be modified so that it could accept a custom formatting string. Let me know what you think and I could start working towards that direction.

> As for enabling the different logging domains, I was thinking use module naming to enable them...

I definitely agree that implementing something like this sounds like it could be very useful. I agree with @0xc0170 though that this should be discussed in a separate PR.

c1728p9:
Hi @tgarc

That is an awesome idea!  I'm all for it.  I have no problem with adding logging statements next to the read() and write() calls.

As for the read vs write distinction, what about just having different logging domains for read and write?  Something like logging.getLogger(__name__.cmsis_dap.read) and logging.getLogger(__name__.cmsis_dap.write)?  That way on the command line to whatever tools you were using you could specify if you wanted to log either or both.

pyOCD/pyDAPAccess/analyzer/cmsis_dap_analyzer.py
tgarc: This has been fixed

tgarc:
@c1728p9 Thanks again for all your input. I'm still not sure I'll do it this way but if I was to do it, what read/write methods would you recommend I add logging? If I'm to adding logging statements (instead of a decorator), I need to have one place to add them which should be a level above the interface (so that it works for all interfaces). I'm thinking to add them in ``pyDAPAccess.DAPAccessUSB._read/write()`` for now and to revisit the issue if any non-cmsis-dap interfaces are added in the future.

c1728p9:
I would recommend adding it right before the read or write calls to pyDAPAccess.DAPAccessUSB _interface object.  I think there are 2 places write is used and 3 places read is used.  If you want them in a unified place to add these a _cmsis_write() and _cmsis_read() could be added to pyDAPAccess.DAPAccessUSB so calls to the interface aren't made directly.

c1728p9:
Hi @tgarc, is this patch ready to be merged from your end?  If its ready I'll start testing on it and after that it can be merged.  Also there are no conflicts, but you might consider rebasing on top of master so there is a more linear commit history.

tgarc:
Hey Russ, Sorry I've gone awol on this the last couple months, I've been working on some different projects. I will have to rebase my changes on top of master and test them to see if they still work as intended. Unfortunately I never got the time to implement the logging like I suggested [previously](https://github.com/mbedmicro/pyOCD/pull/219#issuecomment-169087428), and I'm not sure I will get the time to do that. I can certainly rebase the changes to master and do some re-testing in a short time, but if you want to avoid adding a logging level and read/write decorators as the current implementation does I will have to take a little more time to look at the code again and see if I can't find a solution that I can implement quickly.

c1728p9:
This needs to be rebased before I can run testing.

c1728p9:
Closing this PR to reduce clutter. Please re-open once it is rebased

PR #221, opened 2016-02-20
Added pyocd-gdbserver CLI option to connect using board index

jjones646:


0xc0170:
Hello, can you please sign https://developer.mbed.org/contributor_agreement/?

jjones646:
Yes, done.

flit:
Is the length of the board ID problematic for using the --board option?

I'd rather not have two ways of selecting a board. And a board index is not guaranteed to be consistent; the board indices may change depending on which USB devices are present.

Would supporting unique prefixes on board IDs work for you?

0xc0170:
@flit good points, I was planning to ask why --board is not sufficient

@jjones646 Can you provide a use case why you think --board-index is needed, and why --board is not sufficient or where it falls short?

jjones646:
@flit @0xc0170 We have some scripts that will start pyocd and then run some gdb commands after connecting. Currently, if there's more than 1 device connected, pyocd will prompt the user to select a device. This results in there being no server running when gdb tries to connect. Yes, using a board index does not guarantee uniqueness. However, you can still filter by `--target` and in this case, the binary that gets loaded is independent from the device since they're all identical targets.

_Having a way to bypass the user prompt when there are multiple devices present is the end goal_ since the overall objective is to setup a gdb session - without user input - for any device that matches `--target`. Querying for devices and passing in an arbitruary one isn't ideal since pyocd-gdbserver could reliably do it internally.

I understand this is a very specific use case, so I'm open to any suggestions.

jjones646:
@flit I'm not sure I completely follow what you mean by 'unique prefixes on board IDs'. Please elaborate if you think it fits with any of the above.

c1728p9:
Hi @jjones646 , thanks for the PR!  One thing to be aware of is the ``--target`` option does not filter.  It is meant to override the default target of a board with the target specified.  That way if you want to use a board like the FRDM-K64F to debug an offboard target of a different type you are able to.

Also, I'm not sure what your setup is, but would it make sense to use a custom python script to start the GDB server?  It might offer some more flexibility.  If you did that you could use the return_first option of MbedBoard.chooseBoard().  Something like this:
```
board_selected = MbedBoard.chooseBoard(return_first=True)
with board_selected as board:
    gdb = GDBServer(board, self.args.port_number, self.gdb_server_settings)
```

pyOCD/tools/gdb_server.py
```diff
@@ -55,6 +55,7 @@ def build_parser(self):
         parser.add_argument("-T", "--telnet-port", dest="telnet_port", type=int, default=4444, help="Specify the telnet port for semihosting.")
         parser.add_argument("--allow-remote", dest="serve_local_only", default=True, action="store_false", help="Allow remote TCP/IP connections (default is no).")
         parser.add_argument("-b", "--board", dest="board_id", default=None, help="Connect to board by board id.  Use -l to list all connected boards.")
+        parser.add_argument("-B", "--board-index", dest="board_index", default=None, help="Connect to board using an index value. First board is index 0.")
```
c1728p9: If you use ``type=int`` in add_argument then you don't need to convert it below

pyOCD/tools/gdb_server.py
```diff
@@ -222,10 +223,15 @@ def run(self, args=None):
         self.args = self.build_parser().parse_args(args)
         self.gdb_server_settings = self.get_gdb_server_settings(self.args)
         self.setup_logging(self.args)
-
         self.process_commands(self.args.commands)
 
         gdb = None
+        if self.args.board_index is not None:
+            ind = int(self.args.board_index)
+            all_mbeds = MbedBoard.getAllConnectedBoards(close=True, blocking=False)
```
c1728p9: This list needs to be sorted by unique_id so that the numbering is deterministic (at least if no new boards are attached).  I'm not sure if that makes more sense to do here or in getAllConnectedBoards.

jjones646:
@c1728p9 I was not aware of the `return_first`, so thank you! I believe it is something that can be worked with.

PR #223, opened 2016-02-27
Fix some misc bugs

flit:
- `CMSIS_DAP_Protocol.dapInfo()` didn't work in pyDAPAccess. It also returned string values with a null terminating char still present.
- Some typo fixes.
- Improved hidapi interface backend. Mostly fixes an issue where no devices could be listed if one device failed to be opened because it was already in use by some other process.
- Fixed JSON `--list` and `--list-targets` output from pyocd-gdbserver. Required for the pyOCD launch config plugin to properly list available boards and targets. Was broken after the switch to pyDAPAccess.
- Created a test or the pyocd-gdbserver JSON output.

### DIFF
--- a/pyOCD/tools/gdb_server.py
+++ b/pyOCD/tools/gdb_server.py
@@ -28,6 +28,7 @@ from pyOCD import __version__
 from pyOCD.gdbserver import GDBServer
 from pyOCD.board import MbedBoard
 from pyOCD.utility.cmdline import split_command_line
+from pyOCD.pyDAPAccess.dap_access_usb import DAPAccessUSB
 import pyOCD.board.mbed_board
 
 LEVELS = {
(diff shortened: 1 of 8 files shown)


flit:
All tests pass on a KL46Z.

pyOCD/tools/gdb_server.py
```diff
@@ -178,9 +179,20 @@ def list_boards(self):
...
+                    'product_name' : '',
                     }
+
+                # Reopen the link so we can access the USB vendor and product names from the inteface.
+                # If it's not a USB based link, then we don't attempt this.
+                if isinstance(mbed.link, DAPAccessUSB):
+                    try:
+                        mbed.link.open()
+                        d['vendor_name'] = mbed.link._interface.vendor_name
+                        d['product_name'] = mbed.link._interface.product_name
```
c1728p9: Can you get these values from link.identify(VENDOR) and link.identify(PRODUCT)?

c1728p9:
No regressions on CI testing.  The patch looks good to me

pyOCD/tools/gdb_server.py
flit: No, they're not the same. link.identify() issues a CMSIS-DAP command to get the requested value.
But most CMSIS-DAP builds don't return anything for most of the info keys, except the firmware version. The mbed CMSIS-DAP/DAPLink builds are good examples of this (and they always return "1.0" for the firmware version).

Whether or not that's changed in CMSIS-DAP builds, we'd still need the USB vendor/product string as a backup.

0xc0170:
+1

PR #234, opened 2016-04-15
Fix hanging GDB when continuing after a hardfault

c1728p9:
If a hard fault occurs while debugging the signal SIGSEGV is returned
on halt.  When trying to continue after this hard fault the GDB client
sends the vCommand "Cont;C0b:1;c".  This is ignored in the vCont
processing since nothing matches the action "C0b".  This causes the
target to remain halted and the gdb client and server to go out of
sync.

This patch fixes this problem by comparing against only the first
character in the command.  That way if a continue or step packet
comes in with a signal specified then the correct action will still be
taken.  This patch also adds an error log statement to print if there
is an unsupported command.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -345,7 +345,7 @@ class GDBServer(threading.Thread):
                         self.is_target_running = False
                         self.sendStopNotification()
                     else:
-                        logging.debug("Got unexpected ctrl-c, ignoring")
+                        logging.error("Got unexpected ctrl-c, ignoring")
                     self.packet_io.interrupt_event.clear()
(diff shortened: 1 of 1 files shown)


c1728p9:
@mbed-bot: TEST

0xc0170:
I pulled this change, I got an executable hitting hard fault. The callstack seems to be displayed better (previously was showing nonsense functions), I can close and reenter debug session (eclipse was showing some errors, had to restart entire eclipse app). However, I still can't restart current debug session. Stays at the fault after pressing restart current process.

c1728p9:
@mbed-bot: TEST

0xc0170:
+1, lets merge it

PR #242, opened 2016-06-23
RTOS task awareness

flit:
This PR adds support for RTOS modules that provide thread information to GDB. Two RTOSes are currently supported: FreeRTOS, and my own Argon. As part of this change set, some refactoring was done.

There is a new `DebugContext` class that has methods for reading/writing memory and registers. `DebugContext` objects can be stacked, allowing filtering of these accesses. The RTOS module creates `DebugContext` instances for each thread. This is then used to redirect register accesses to the thread's saved state (or live registers if the thread is the current one).

`GdbServer` now uses a façade class named  `GDBDebugContextFacade` that wraps `DebugContext` to provide its gdb-specific data. With this change, the gdb-specific methods were removed from `CortexM` and `Target`.

RTOS modules implement a subclass of `ThreadProvider`. There is a table of RTOS modules similar to the targets list in `./targets/__init__.py`. RTOSes are automatically discovered.

### DIFF
--- a/pyOCD/rtos/freertos.py
+++ b/pyOCD/rtos/freertos.py
@@ -1,0 +1,503 @@
+"""
+ mbed CMSIS-DAP debugger
+ Copyright (c) 2016 ARM Limited
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from .provider import (TargetThread, ThreadProvider)
+from .common import (read_c_string, HandlerModeThread)
+from ..debug.context import DebugContext
+from ..coresight.cortex_m import CORE_REGISTER
+from pyOCD.pyDAPAccess import DAPAccess
+import logging
+
+FREERTOS_MAX_PRIORITIES	= 63
+
+LIST_SIZE = 20
+LIST_INDEX_OFFSET = 16
+LIST_NODE_NEXT_OFFSET = 8 # 4?
+LIST_NODE_OBJECT_OFFSET = 12
+
+THREAD_STACK_POINTER_OFFSET = 0
+THREAD_PRIORITY_OFFSET = 44
+THREAD_NAME_OFFSET = 52
+
+# Create a logger for this module.
+log = logging.getLogger("freertos")
+
+class TargetList(object):
+    def __init__(self, context, ptr):
+        self._context = context
+        self._list = ptr
+
+    def __iter__(self):
+        prev = -1
+        found = 0
+        count = self._context.read32(self._list)
+        if count == 0:
+            return
+
+        node = self._context.read32(self._list + LIST_INDEX_OFFSET)
+
+        while (node != 0) and (node != prev) and (found < count):
+            try:
+                # Read the object from the node.
+                obj = self._context.read32(node + LIST_NODE_OBJECT_OFFSET)
+                yield obj
+                found += 1
+
+                # Read next list node pointer.
+                prev = node
+                node = self._context.read32(node + LIST_NODE_NEXT_OFFSET)
+            except DAPAccess.TransferError:
+                log.warning("TransferError while reading list elements (list=0x%08x, node=0x%08x), terminating list", self._list, node)
+                node = 0
+
+## @brief
+class FreeRTOSThreadContext(DebugContext):
+    # SP/PSP are handled specially, so it is not in these dicts.
+
+    COMMON_REGISTER_OFFSETS = {
+                 4: 0, # r4
+                 5: 4, # r5
+                 6: 8, # r6
+                 7: 12, # r7
+                 8: 16, # r8
+                 9: 20, # r9
+                 10: 24, # r10
+                 11: 28, # r11
+            }
+
+    NOFPU_REGISTER_OFFSETS = {
+                 0: 32, # r0
+                 1: 36, # r1
+                 2: 40, # r2
+                 3: 44, # r3
+                 12: 48, # r12
+                 14: 52, # lr
+                 15: 56, # pc
+                 16: 60, # xpsr
+            }
+    NOFPU_REGISTER_OFFSETS.update(COMMON_REGISTER_OFFSETS)
+
+    FPU_BASIC_REGISTER_OFFSETS = {
+                -1: 32, # exception LR
+                 0: 36, # r0
+                 1: 40, # r1
+                 2: 44, # r2
+                 3: 48, # r3
+                 12: 42, # r12
+                 14: 56, # lr
+                 15: 60, # pc
+                 16: 64, # xpsr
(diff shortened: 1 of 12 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
So does this allow RTOS task awareness without any extra plugins to eclipse?  If so that is awesome!

What is the easiest way to try out this change?  Pull down the KSDK and build a FreeRTOS project?

flit:
@c1728p9 Right, you don't need any additional Eclipse plugins. It's all done through gdb and standard Eclipse functionality.

KSDK is probably the easiest way to test. Go to [kex.nxp.com](http://kex.nxp.com) and generate an SDK for your board (make sure to include FreeRTOS support). The FreeRTOS examples will be under `boards/*/rtos_examples`.

flit:
Looking into failure…

flit:
Silly bug. I forgot to update cortex_test.py to use the `GDBDebugContextFacade`.

c1728p9:
@mbed-bot: TEST

flit:
That darn LPC1768 again… ;) Strange that it passed in the first test run but failed just now.

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

dumarjo:
Is this port support FPU correctly ?

regards

Jonathan

flit:
@dumarjo Yes, the FPU is fully supported. Tested on Kinetis K22F and K64F.

flit:
Rebased.

@c1728p9 Have you had a chance to test this?

c1728p9:
Hi @flit, unfortunately, I haven't yet. It is on my list though, as this would be a very useful feature to have.

flit:
No worries, just curious.

matthewelse:
@flit: do you have any plans to add support for RTX/mbed OS?

flit:
@matthewelse Not myself, though I'd be quite happy to help someone else add RTX. The RTOS support is designed to be easy to add new RTOSes.

c1728p9:
I'm not sure this will scale very well if support for each RTOS needs to be added do pyOCD. What would be nice is if the rtos specific code could be moved out of pyOCD and instead live as a plugin.

flit:
I did think about that :) But in practice, there are far fewer RTOSes in common use than even the number of targets or flash algos we have. Not enough to warrant the plugin machinery, imo. Adding a new RTOS via a PR would be similar to adding target devices, except less frequent.

The other major RTOSes used on Cortex-M that are not supported here are: µC/OS-II, µC/OS-III, and RTX. FreeRTOS is by far the most common.

If we do decide to switch to plugins, that could be done in a separate PR.

flit:
Rebased and squashed.

flit:
Rebased and refactored commits. Now based on #285, so the debug context and RTOS support have separated. Once #285 is merged, I will rebase this PR to make the changes easier to see. Or you can just look at de9e8a7 (the last commit) to see the RTOS-specific changes.

flit:
Rebased to remove debug context related commits that were merged in #285.

c1728p9:
Testing passes on this, but it needs to be rebased

flit:
Rebased.

pyOCD/rtos/provider.py
```diff
@@ -0,0 +1,98 @@
...
+    @property
+    def is_current(self):
+        raise NotImplementedError()
+
+    @property
+    def context(self):
+        raise NotImplementedError()
+
+## @brief Base class for RTOS support plugins.
+class ThreadProvider(object):
```
c1728p9: It looks like the functions get_ipsr() (used in common.py) and get_current_thread_id() (used in gdbserver.py) are required to be implemented by subclasses of this. It might be good to add them to this class with NotImplementedError's.

flit: Good catch. Fixed in commit I just pushed. `get_ipsr()` was moved up to `ThreadProvider` since it had a common implementation.

pyOCD/rtos/freertos.py
```diff
@@ -0,0 +1,499 @@
...
+        if not self.is_enabled:
+            return False
+        self.update_threads()
+        return threadId in self._threads
+
+    def get_current_thread_id(self):
+        if not self.is_enabled:
+            return None
+        if self.get_ipsr() > 0:
+            return 2
```
c1728p9: Why is the handler thread fixed to 2?

c1728p9:
Also, I noticed that stepping gets considerably slow when the RTOS is on. Is that primarily what https://github.com/mbedmicro/pyOCD/pull/296 is trying to fix? Could there be an option on the command line to conditionally enable this?

pyOCD/rtos/freertos.py
flit: It just needs an arbitrary, fixed thread ID so gdb will recognize it as the same thread. The default Thread mode (before threading is enabled) thread ID is 1.

flit:
To be honest, I hadn't even tried stepping a non-current thread! I'm used to that not working as expected in other debuggers and trained myself to not even try. I'll have to check what gdb sends for the vCont command to see if it's something we could even support. But I suspect it would be very complicated at the least.

Stepping is definitely slower. gdb is causing a thread scan and reading the register context of every thread for each step command, so potentially even multiple times for a user-visible step. You can turn on debug logging to see how frequently this happens.

#296 was indeed intended to help with stepping performance. And it does, a little. The biggest help, of course, is high-speed USB. It's the latency of HID that really gets in the way here.

We could certainly add a command line option to disable RTOS support. Although, imo, you can't really debug a threaded application like that.

pyOCD/rtos/freertos.py
c1728p9: Ah ok, that makes sense

flit:
Fyi, there is an issue I've run into (at least with Argon) where the thread names will end up with garbage values. This is caused by modified code being loaded on a target that was running old code. The SRAM contents will be stale until the app starts running and inits its data and bss. The problem is, there's no good way to identify when the app has reached main().

I have implemented a fix, but it has limits. Basically, it holds off the ThreadProvider from reading thread info until the first resume command following a reset or flash programming. For the common case, this works fine. But there are issues if you don't run to main().

Right now, the fix is in a stash in my local repo. It depends on an event notification system added on another branch, so it's not something I can directly add to the RTOS PR without also bringing in the notification system.

c1728p9:
In addition to caching, deferred reads could be used. It would make the driver more complicated, but it would enable many reads to be packed into the same packet, rather than requiring 1 packet for each read.

I use pyOCD to debug my mbed-os code and it's fairly straight forward even with an RTOS, so I think this is a valid use case.

c1728p9:
I have noticed something similar. If I reset the device, I'll still see the tasks that were running, since the ram is still preserved. I'm not sure if it is that big of an issue though.

One possible (slight intrusive) way to improve this could be to write to memory to set the RTOS state to inactive on reset. At least for RTX2 there is a variable which indicates the state of the RTOS. I'm not sure if Argon or FreeRTOS have something similar.

flit:
That's a good idea, I hadn't thought of clearing just the RTOS state flag. That should work for Argon and FreeRTOS, as well.

Anyway, I pushed up a branch containing my current workaround for you to review: [dev_fix_stale_rtos](https://github.com/flit/pyOCD/tree/dev_fix_stale_rtos)

studavekar:

PR #244, opened 2016-07-01
Add TOSHIBA TZ10xx target / Cerevo BlueNinja board supports.(REDO)

f-okuhara:
We added the target "[TOSHIBA TZ1000 Series](http://toshiba.semicon-storage.com/ap-en/product/assp/applite/tz1000.html)" and added the board "Cerevo BlueNinja".
Please merge.

Run the automated_test.py, we have all passed.

```
------ Speed Test Performance ------
Target    Read Speed      Write Speed     

tz10xx    28.306230 KB/s  23.245899 KB/s  



------ Analyzer Performance ------
Target    Analyzer    Rate              Time              

tz10xx    PAGE_READ   127.688259 KB/s   8.21200013161 s   



------ Test Rate ------
Target    Chip Erase          Page Erase          Page Erase (Same data)

tz10xx    9.176944 KB/s       1.114198 KB/s       19.353919 KB/s      

Target         Test                 Result         Time           

tz10xx         Basic Test           Pass           13.8659999371  
tz10xx         Gdb Server Json Test Pass           0.601999998093 
tz10xx         Speed Test           Pass           58.4079999924  
tz10xx         Cortex Test          Pass           8.71800017357  
tz10xx         Flash Test           Pass           1963.78699994  
tz10xx         Gdb Test             Pass           157.769999981  

Test Time: 2203.15400004
All tests passed
```

Best regards.


c1728p9:
Thanks for the PR!

c1728p9:
@mbed-bot: TEST

pyOCD/target/target_tz10xx.py
```diff
@@ -0,0 +1,352 @@
...
+            self.__initializePrescaler()
+            self.__powerOnDomain()
+            self.__initializeIOCell()
+            self.__resetOnDomain()
+            self.__powerOffDomain()
+            self.__stopPrescaler()
+            self.__initializeADPLL()
+            self.__initializeRTC()
+            self.__initializeMacro()
+            self.__initializeMiscRegisters()
```
c1728p9: Do all these components get powered off from a software reset?  What about a hardware reset?

pyOCD/flash/flash_tz10xx.py
```diff
@@ -0,0 +1,99 @@
...
+    'page_size' : 0x00000100,
+    'analyzer_supported' : False,
+    'analyzer_address' : 0x00000000  # ITCM, Analyzer 0x00000000..0x000000600
+};
+
+class Flash_tz10xx(Flash):
+    def __init__(self, target):
+        super(Flash_tz10xx, self).__init__(target, flash_algo)
+
+    def erasePage(self, flashPtr):
```
c1728p9: If you treat sector size as 0x00000100 then you shouldn't need the extra logic here.  The flash driver should treat a page erase as a sector erase.

pyOCD/target/target_tz10xx.py
f-okuhara: I tried the hardware reset, but was not stable after reset operation from the GDB.
So, I was ported initialization script provided by the chip makers.

pyOCD/flash/flash_tz10xx.py
f-okuhara: Unfortunately TZ1000's  PAGE_SIZE and SECTOR_SIZE is different...
  PAGE_SIZE: 0x00000100 (256Bytes)
  SECTOR_SIZE: 0x00001000 (4kB)

PROGRAM is each page, ERASE is each sector.
So, I wrote the additional logic.

pyOCD/target/target_tz10xx.py
c1728p9: Does this initialization sequence follow the "PMU initialization setting when connecting a debugger" in the TZ1000 Series Reference Manual, or is this board specific?

f-okuhara: > Does this initialization sequence follow the "PMU initialization setting when connecting a debugger" in the TZ1000 Series Reference Manual

Yes.
But, specific steps have not been written in the reference manual.
So, I was in reference to initialization script for the MDK-ARM.

> is this board specific?

This board specific initialize process not included.

c1728p9:
Hi @f-okuhara, I gave this a try, but none of the testing worked unless SW1 was pressed. Also, some of the flash operations took a very long time. Is that expected? What are the sector erase and programming times? Could a clock be set to the wrong speed causing the slow flashing?

pyOCD/target/target_tz10xx.py
c1728p9: If the power button is held then this code is not needed. Maybe instead of customizing pyOCD so much here, you could have DAPLink keep power enabled when a debugger is connected. What do you think @f-okuhara? In future versions of pyOCD the goal is to pull target support from cmsis-packs. Because of this there will no longer be target customization like here since this detailed sequence might not be in cmsis-packs.

f-okuhara:
>  I gave this a try, but none of the testing worked unless SW1 was pressed.
BlueNinja's specification is powerd on only while SW1 is pressed.
If the users want to keep the power on even if SW1 is released, there is a need to perform holding operation with software.
So it is a matter of course if the software that performs the power hold operation is not running.

We designed BlueNinja so that the user can choose the power source and usage according to the user's application.
So, we think development tools should not intervene in power operation, and should give the user choices.

I understand about unnatural looking from the common sense of mbed.
But since BlueNinja is already a selling product, changes that affect product concept are impossible.

f-okuhara:
> I gave this a try, but none of the testing worked unless SW1 was pressed.

BlueNinja's specification is powerd on only while SW1 is pressed.
If the users want to keep the power on even if SW1 is released, there is a need to perform holding operation with software.
So it is a matter of course if the software that performs the power hold operation is not running.

We designed BlueNinja so that the user can choose the power source and usage according to the user's application.
So, we think development tools should not intervene in power operation, and should give the user choices.

I understand about unnatural looking from the common sense of mbed.
But since BlueNinja is already a selling product, changes that affect product concept are impossible.

f-okuhara:
> Also, some of the flash operations took a very long time. Is that expected? What are the sector erase and programming times? Could a clock be set to the wrong speed causing the slow flashing?

I am aware.

TZ100x seriese's Flash ROM has no page erase function, page erase is re-implemented by sector erase, so it is very very late.

PyOCD calls as is FlashAlgo's "EraseSector" function in page erase processing.
This seemed to assume only FlashROM with the same sector size and page size.
Even in the test, it seemed to assume that the sector size and the page size are the same.

So I implemented it according to the prerequisite of the test.

I know that the performance is not good.

pyOCD/flash/flash_tz10xx.py
c1728p9: @f-okuhara as an alternative, what about setting 'page_size' to 0x1000? That way you don't need to have erasePage or eraseSector here.

f-okuhara:
@c1728p9 
> Why is it a problem to keep the board powered when a debugging session is active? If the board loses power then you can't debug anyway.

I could understand what you are saying.
During the debugging session, BlueNinja should keep the power supply even if SW1 is released.

It is reasonable.
We will assert TZ 1000's PowerHold DigitalOut at the start of the debug session.

f-okuhara:
@c1728p9 
> @f-okuhara as an alternative, what about setting 'page_size' to 0x1000? That way you don't need to have erasePage or eraseSector here.

Thank you for presenting a solution.
I'll give it a try.

c1728p9:
Hi @f-okuhara any updates on this PR?

f-okuhara:
@c1728p9 
Sorry, not update...
I could not work due to other jobs.

Please give me time as there is a corresponding intention.

flit:
@f-okuhara Any chance you can update this PR so we can merge it in? It needs to be rebased and requested changes applied. Thanks!

f-okuhara:
Sorry.
Unfortunately, this project was discontinued.
So, never updates...

I close this PR.

flit:
That's too bad, thanks for the update.

PR #247, opened 2016-08-08
Remove rom table initialization

arekzaluski:
Currently ROM table is not being used. I commented it because it increase the time of the connection by 4 times.

@c1728p9 @flit @matthewelse

### DIFF
--- a/pyOCD/coresight/ap.py
+++ b/pyOCD/coresight/ap.py
@@ -84,7 +84,8 @@ class AccessPort(object):
 
             self.inited_primary = True
         if not self.inited_secondary and self.has_rom_table and bus_accessible:
-            self.init_rom_table()
+            #Initializing and reading rom table decrease the performance. Connecting to board takes 4 times longer.
+            #self.init_rom_table()
             self.inited_secondary = True
 
     def init_rom_table(self):


c1728p9:
Thanks for the PR

c1728p9:
@mbed-bot: TEST

flit:
@arekzaluski Note that the ROM table will be used in the future to dynamically build the graph of CoreSight objects (like the J-Link and other debug probes do). It's also useful to see what features are present on the device.

I'd prefer that the code be optimized instead of turning off the reading of the ROM table entirely. There are several possible optimizations, including reading the CoreSight registers as a block instead of individually.

c1728p9:
Hi @flit, could this optimization be done at them time the ROm table is used? Reading the rom table is causing significant usability problems in the configuration @arekzaluski is using.

flit:
@c1728p9, I'd rather at least attempt to fix the real problem now, and find a solution for @arekzaluski with the ROM table reading enabled. Otherwise we might end up in this situation again when full ROM table support is implemented, even with optimizations. By turning off the ROM table reading, we're just ignoring a problem that could arise again.

The key code that can be optimized easily is [CoreSightComponent.read_id_registers()](https://github.com/flit/pyOCD/blob/master/pyOCD/coresight/rom_table.py#L100) and  [CoreSightComponent.read_id_register_set()](https://github.com/flit/pyOCD/blob/master/pyOCD/coresight/rom_table.py#L119). Changing `read_id_register_set()` to a single contiguous read may significantly speed up table reading. Right now it reads 4x 32-bit registers as separate, immediate reads—pretty naïve implementation, sorry!  `read_id_registers()` calls `read_id_register_set()` 3x times. For CoreSight-class components, it also does two 32-bit reads that could be combined into a single contiguous read.

The `ROMTable.read_table_32()` and `ROMTable.read_table_8()` methods could probably also be sped up by reading blocks of 4 or 8 words at a time. But I'm not sure exactly what happens if you read off the end of the ROM table, need to check into that.

I can take a look at this within the week, probably today or tomorrow. Shouldn't take long to make the changes.

c1728p9:
Thanks for looking into this @flit! I'll hold off on merging this so you can implement an alternative fix.

flit:
Created PR #249 with the optimizations. Went from ~540ms to ~29ms.

arekzaluski:
@flit Thank you. I've checked your PR #249 and initialisation is much faster. I'm closing this PR and I'm looking forward to see your PR merged. It should solve my performance issues.

PR #254, opened 2016-08-11
Fix breakpoints

c1728p9:
Allow debugging generic cortex_m devices without a memory map present. Also improve handling when out of breakpoints.

### DIFF
--- a/pyOCD/coresight/breakpoints.py
+++ b/pyOCD/coresight/breakpoints.py
@@ -216,10 +216,12 @@ class BreakpointManager(object):
         except KeyError:
             raise RuntimeError("Unknown breakpoint type %d" % type)
 
-        # Save the bp.
-        if bp is not None:
-            self._breakpoints[addr] = bp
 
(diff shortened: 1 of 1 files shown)


c1728p9:
@mbed-bot: TEST

c1728p9:
@flit, this fixes some problems with breakpoints. If you don't mind could you give it a quick review?

MarceloSalazar:
Tested on an unsupported platform (beetle) and works great!
It's important to build with symbols and update the target with the corresponding binary (otherwise symbols won't match).

Minor suggestion: this message shown in the gdb_server console could be changed into a WARNING, so people don't get confused:
`ERROR:root:Unsupported board found 5003`

Thanks @c1728p9!

flit:
@MarceloSalazar I agree with your suggestion about changing the unsupported board message to a warning.

flit:
@c1728p9 LGTM. Sorry I didn't get to it until after you merged!

PR #257, opened 2016-08-12
Display a warning for unsupported boards

c1728p9:
Display a warning rather than an error when the target board is
unsupported. This is not a fatal error, since pyOCD can still connect
and be used to debug. It just cannot flash the target.


c1728p9:
@mbed-bot: TEST

Issue #259, opened 2016-08-17
pyocd need sudo to work


aurabindo:
Hi,

Simply running pyocd-gdbserver on terminal gives me 

``` python
Traceback (most recent call last):    
  File "/usr/bin/pyocd-gdbserver", line 5, in <module>    
    from pkg_resources import load_entry_point    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2927, in <module>    
    @_call_aside    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2913, in _call_aside    
    f(*args, **kwargs)    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2952, in _initialize_master_working_set    
    add_activation_listener(lambda dist: dist.activate())    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 956, in subscribe    
    callback(dist)    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2952, in <lambda>    
    add_activation_listener(lambda dist: dist.activate())    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2513, in activate    
    for pkg in self._get_metadata('namespace_packages.txt'):    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2503, in _get_metadata    
    for line in self.get_metadata_lines(name):    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1477, in get_metadata_lines    
    return yield_lines(self.get_metadata(name))    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1469, in get_metadata    
    return self._get(self._fn(self.egg_info, name))    
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1580, in _get    
    with open(path, 'rb') as stream:    
IOError: [Errno 13] Permission denied: '/usr/lib/python2.7/site-packages/protobuf-3.0.0b3-py2.7.egg/EGG-INFO/namespace_packages.tx
```

Same case with pyocd-tool, `sudo pyocd-tool --version` is `0.1.dev24`

flit:
You shouldn't need root permissions for any of the pyOCD tools on any system (OS X, Linux, Windows).

[...]

PR #261, opened 2016-08-23
pyusb backend: add exception for Hexiwear

aurabindo:
Enable pyOCD to run as normal user, a hack for Hexiwear.
Without this hack, only root user can debug/flash.
Kinetis Design Studio launches pyOCD as normal user.

Signed-off-by: Aurabindo J <mail@aurabindo.in>

### DIFF
--- a/pyOCD/pyDAPAccess/interface/pyusb_backend.py
+++ b/pyOCD/pyDAPAccess/interface/pyusb_backend.py
@@ -86,6 +86,9 @@ class PyUSB(Interface):
                 # This can cause an exception to be thrown if the device
                 # is malfunctioning.
                 product = board.product
+            except ValueError:
+                if board.idVendor is 0xd28 and board.idProduct is 0x204:
+                    continue
             except usb.core.USBError as error:
                 logging.warning("Exception getting product string: %s", error)
                 continue


aurabindo:
This fixes #259

pyOCD/pyDAPAccess/interface/pyusb_backend.py
```diff
@@ -86,6 +86,9 @@ def getAllConnectedInterface():
                 # This can cause an exception to be thrown if the device
                 # is malfunctioning.
                 product = board.product
+            except ValueError:
+                if board.idVendor is 0xd28 and board.idProduct is 0x204:
```
c1728p9: So this code is ignoring exceptions that occur on boards with a vid,pid other than 0xd28, 0x204? Can you provide any more details on this? Which VID/PID does Hexiwear have?

aurabindo:
The intention is to ignore the exception for the specified VID/PID, which belongs to the baseboard, which is part of Hexiwear IoT Kit. This board provides extension ports and most importantly an OpenSDA debug adapter. I've been discussing with pyusb devs regarding [this](this), since its  not easy for me to make eclipse (Kinetis Design Studio) run pyOCD as root, and they dont seem to be interested in [fixing the issue ](nxp)

[issue]:https://sourceforge.net/p/pyusb/mailman/pyusb-users/thread/GwV3hDpxIGyh5ucEH76ywnJIc84EHGdM2HHykpY42146%40mailpile/#msg35297171
[nxp]:https://community.nxp.com/thread/432509

c1728p9:
@aurabindo, I can't get the links in your last post to work. Could you provide updated links?

aurabindo:
@c1728p9 github seems to be doing something nasty. Try these:

1: https://sourceforge.net/p/pyusb/mailman/pyusb-users/thread/CAOsEZohWUs%3D6i7%2BQEz36pFvX3JYiDiu6s_3AAv5qTPzsY2zkWg%40mail.gmail.com/#msg35291047
2: https://community.nxp.com/thread/432509

rsalveti:
Just proposed #262 and saw this pull request. With #262 I'm generically catching valueerror, which can happen with any device (since it probes all the devices you have in your system).

That + a udev file for your own device should be enough.

Happy to close that and update yours with a more generic solution.

aurabindo:
I was not sure if this is specific to hexiwear, which is why I didn't go for a generic solution as upstream might not like it. I am still expecting a comment from @c1728p9

Feel free to update mine if you feel that is appropriate. But you have merge rights, and you think a generic solution makes more sense, please go ahead and merge yours.

c1728p9:
Hi @aurabindo, sorry for not responding sooner. It looks like this is problem for boards in general and not specific to the Hexiwear, so the generic solution is probably more appropriate.

aurabindo:
Thanks for the update @c1728p9. #262 will be more appropriate then

PR #273, opened 2017-01-11
Output of Python code should be listed as comments, so the 'Hello world' is valid Python

janjongboom:
See the confusion in http://stackoverflow.com/questions/41587382/odd-python-syntax

@c1728p9


c1728p9:
@mbed-bot: TEST

PR #276, opened 2017-02-15
Separate SWJ sequence from connect

c1728p9:
Move the sending of the SWJ sequence out of connect. This allows the SWJ sequence to be sent multiple times without re-initializing the SWD/JTAG pins. It also allows the SWJ sequence to be sent while reset is held low.

With this patch the function connect just configures debugger IO pins for the desired protocol. The function swj_sequence() must then be used to prepare the target for communication.


c1728p9:
CC: @flit

c1728p9:
@mbed-bot: TEST

PR #278, opened 2017-02-25
Clear sticky errors if a fault happens due to a flush

flit:
This fixes a case where a write to an invalid address is put into a packet and not immediately flushed. When it was flushed, the sticky flags weren't cleared, so the next valid read or write would fail.

- `CoreSightTarget` and `CortexM` classes override `flush()` to call the `DebugPort.flush()` method. 
- Handling errors in `DebugPort.flush()` by clearing sticky flags.

Also includes 2 other fixes:
- If the target subclass does not provide a memory map, an empty one will be created. This currently only applies to the generic 'cortex_m' target.
- The write memory commands in pyocd-tool will now always perform a flush after writing.

### DIFF
--- a/pyOCD/coresight/cortex_m.py
+++ b/pyOCD/coresight/cortex_m.py
@@ -386,6 +386,9 @@ class CortexM(Target):
         """
         return self.dp.read_id_code()
 
+    def flush(self):
+        self.dp.flush()
+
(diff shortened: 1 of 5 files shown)


c1728p9:
@mbed-bot: TEST

PR #281, opened 2017-03-14
Support all vector catch options

flit:
Reworked vector catch APIs in `Target`. There is now a single `setVectorCatch()` that takes a mask and `getVectorCatch()` that returns a mask.

There are new mask constants defined in `Target`.
```py
    CATCH_NONE = 0
    CATCH_HARD_FAULT = (1 << 0)
    CATCH_BUS_FAULT = (1 << 1)
    CATCH_MEM_FAULT = (1 << 2)
    CATCH_INTERRUPT_ERR = (1 << 3)
    CATCH_STATE_ERR = (1 << 4)
    CATCH_CHECK_ERR = (1 << 5)
    CATCH_COPROCESSOR_ERR = (1 << 6)
    CATCH_CORE_RESET = (1 << 7)
    CATCH_ALL = (CATCH_HARD_FAULT | CATCH_BUS_FAULT | CATCH_MEM_FAULT | CATCH_INTERRUPT_ERR \
                | CATCH_STATE_ERR | CATCH_CHECK_ERR | CATCH_COPROCESSOR_ERR | CATCH_CORE_RESET)
```

Added `--vector-catch` option to pyocd-gdbserver. It accepts a list of vector catch options to enable as a string composed of these characters: h=hard fault, b=bus fault, m=mem fault, i=irq err, s=state err, c=check err, p=nocp, r=reset, a=all.

### DIFF
--- a/pyOCD/tools/gdb_server.py
+++ b/pyOCD/tools/gdb_server.py
@@ -65,7 +65,7 @@ class GDBServerTool(object):
         parser.add_argument("-t", "--target", dest="target_override", choices=supported_targets, default=None, help="Override target to debug.  Supported targets are: " + ", ".join(supported_targets), metavar="TARGET")
         parser.add_argument("-n", "--nobreak", dest="break_at_hardfault", default=True, action="store_false", help="Disable halt at hardfault handler.")
         parser.add_argument("-r", "--reset-break", dest="break_on_reset", default=False, action="store_true", help="Halt the target when reset.")
-        parser.add_argument("-C", "--vector-catch", default=None, help="Select enabled vector catch options, one letter per enabled source in any order. (h=hard fault, b=bus fault, m=mem fault, i=irq err, s=state err, c=check err, p=nocp, r=reset, a=all)")
+        parser.add_argument("-C", "--vector-catch", default='', help="Select enabled vector catch options, one letter per enabled source in any order. (h=hard fault, b=bus fault, m=mem fault, i=irq err, s=state err, c=check err, p=nocp, r=reset, a=all, n=none)")
         parser.add_argument("-s", "--step-int", dest="step_into_interrupt", default=False, action="store_true", help="Allow single stepping to step into interrupts.")
         parser.add_argument("-f", "--frequency", dest="frequency", default=1000000, type=int, help="Set the SWD clock frequency in Hz.")
         parser.add_argument("-o", "--persist", dest="persist", default=False, action="store_true", help="Keep GDB server running even after remote has detached.")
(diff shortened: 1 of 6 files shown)


c1728p9:
@flit please rebase this PR

flit:
Will do.

flit:
@c1728p9 Rebased.

c1728p9:
@mbed-bot: TEST

c1728p9:
@mbed-bot: TEST

flit:
@c1728p9 Is this a test harness failure, or something in the changeset?

c1728p9:
I think this is a real failure, but it is manifesting as a timeout.

flit:
Ok, thanks. I'll look into it more. Do you have any idea if it's failing for all targets or just some?

c1728p9:
@flit it appears to be all of them

c1728p9:
@mbed-bot: TEST

pyOCD/tools/gdb_server.py
```diff
@@ -65,6 +65,7 @@ def build_parser(self):
         parser.add_argument("-t", "--target", dest="target_override", choices=supported_targets, default=None, help="Override target to debug.  Supported targets are: " + ", ".join(supported_targets), metavar="TARGET")
         parser.add_argument("-n", "--nobreak", dest="break_at_hardfault", default=True, action="store_false", help="Disable halt at hardfault handler.")
         parser.add_argument("-r", "--reset-break", dest="break_on_reset", default=False, action="store_true", help="Halt the target when reset.")
+        parser.add_argument("-C", "--vector-catch", default='', help="Select enabled vector catch options, one letter per enabled source in any order. (h=hard fault, b=bus fault, m=mem fault, i=irq err, s=state err, c=check err, p=nocp, r=reset, a=all, n=none)")
```
c1728p9: Should this default to the string "none"?

flit: (It would be 'n' rather than "none".) It could, but it doesn't really matter. It only checks for the characters to enable catch options; it doesn't actually check for the none option. But I thought it would be good to document a char you could use for none, since that's easier to type on a command line than an empty string (i.e., `-Cn` versus `-C ""`).

c1728p9: 'n' sounds good to me. The code treats '' the same as 'n'?

flit: Right. It really just ignores any character it doesn't recognize, so for example 'z' is also equivalent. The relevant code that converts from the string to mask is in `GDBServer.__init__()`:

```c
        mask = ((Target.CATCH_HARD_FAULT if ('h' in self.vector_catch or self.break_at_hardfault) else 0) \
                | (Target.CATCH_BUS_FAULT if 'b' in self.vector_catch else 0) \
                | (Target.CATCH_MEM_FAULT if 'm' in self.vector_catch else 0) \
                | (Target.CATCH_INTERRUPT_ERR if 'i' in self.vector_catch else 0) \
                | (Target.CATCH_STATE_ERR if 's' in self.vector_catch else 0) \
                | (Target.CATCH_CHECK_ERR if 'c' in self.vector_catch else 0) \
                | (Target.CATCH_COPROCESSOR_ERR if 'p' in self.vector_catch else 0) \
                | (Target.CATCH_CORE_RESET if ('r' in self.vector_catch or self.break_on_reset) else 0) \
                | (Target.CATCH_ALL if 'a' in self.vector_catch else 0))
```

PR #282, opened 2017-03-20
Remove flash parameter for Board class

c1728p9:
Flash is always the same as the target, so specifying it is redundant. Remove this parameter and use the target name for this purpose instead.


c1728p9:
@mbed-bot: TEST

c1728p9:
cc: @flit

c1728p9:
@mbed-bot: TEST

flit:
LGTM 👍

PR #286, opened 2017-03-23
Fix two error log related exceptions

flit:
Fixed exceptions raised when logging errors from two places:
1. DebugPort when scanning APs.
2. Kinetis target class when MDM-AP IDR match fails.


c1728p9:
@mbed-bot: TEST

PR #288, opened 2017-03-25
Memory map improvements

flit:
Improvements to the memory map classes.

- Added device (peripheral) and alias memory region types.
- New `MemoryRange` class that represents a range within a memory region.
- Methods to test containment and intersection.
- Cache-related properties for regions: `isCacheable`,` invalidateCacheOnRun`
- Added `getRegionByName()` to `MemoryMap`.

Also added a memory map unit test.


c1728p9:
@mbed-bot: TEST

PR #290, opened 2017-04-06
Fix issues programming DAP under python3

utzig:
Trying to program the FRDM-K64F DAP under python3 currently fails because
of the changed semantics for division which returns `float` on py3 and
unicode string literals which cause an abort on XML processing.

NOTE: This was tested under Python 3.6.1 and 2.7.12.

### DIFF
--- a/pyOCD/coresight/ap.py
+++ b/pyOCD/coresight/ap.py
@@ -356,9 +356,9 @@ class MEM_AP(AccessPort):
             n = self.auto_increment_page_size - (addr & (self.auto_increment_page_size - 1))
             if size*4 < n:
                 n = (size*4) & 0xfffffffc
-            self._writeBlock32(addr, data[:n/4])
-            data = data[n/4:]
-            size -= n/4
+            self._writeBlock32(addr, data[:n//4])
+            data = data[n//4:]
+            size -= n//4
             addr += n
         return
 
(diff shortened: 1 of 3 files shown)


c1728p9:
Thanks for the PR @utzig. Python 3 support will be important to have.

c1728p9:
@mbed-bot: TEST

carlosperate:
I think like 289 might also need the floor division operator:

```python
         # try to read aligned block of 32bits
         if (size >= 4):
             mem = self.readBlockMemoryAligned32(addr, size/4)
             res += conversion.u32leListToByteList(mem)
             size -= 4*len(mem)
             addr += 4*len(mem)
```

utzig:
@carlosperate I don't think it would be a problem to do this change, but in practice it is not required because the function only does basic math and comparisons, and it calls `_readBlock32` which lower down in the call chain will do the required conversion.

NOTE: I only tested with one target, which probably means that many possible paths still might need to be updated and also I made the changes as simple as possible to solve my problem.

carlosperate:
> @carlosperate I don't think it would be a problem to do this change, but in practice it is not required because the function only does basic math and comparisons, and it calls _readBlock32 which lower down in the call chain will do the required conversion.

Logically speaking yes, but unless we add the `from __futures__ import division`, then the code would technically be doing a different operation in Py2 vs Py3.

utzig:
OK, signed!

c1728p9:
Please resolve the conflicts and then I can merge.

utzig:
Is there anything remaining to be done on my part?

c1728p9:
Nope. I just need to run testing and it can be merged.

studavekar:

utzig:
@c1728p9 Is there anyone working on Python3 porting? I also had issues running debugging (pyocd-gdbserver). If nobody is working on it I can give a try fixing those issues too.

c1728p9:
No one is actively working on python 3 support at the moment, but this is definitely something that would be good to have. Any fixes would be greatly appreciated.

20161220-rootfest:
Will it be possible to release this fix to PyPI? The current version (pip3 install pyOCD) is not usable

PR #303, opened 2017-05-05
Timeouts for Kinetis mass erase

flit:
Created a `Timeout` context manager utility class, with a small unit test. This is used in `Kinetis.massErase()` to prevent a mass erase from hanging (which can occur in some rare situations I've seen, mostly with new devices). And, of course, the `Timeout` class can be used in many other parts of pyOCD.

### DIFF
--- a/pyOCD/test/test_utility/test_timeout.py
+++ b/pyOCD/test/test_utility/test_timeout.py
@@ -1,0 +1,42 @@
+"""
+ mbed CMSIS-DAP debugger
+ Copyright (c) 2017 ARM Limited
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


pyOCD/test/test_utility/test_timeout.py
```diff
@@ -0,0 +1,42 @@
...
+
+from pyOCD.utility.timeout import (Timeout, TimeoutException)
+from time import (time, sleep)
+import pytest
+
+class TestTimeout:
+    def test_no_timeout(self):
+        with Timeout(0.5) as to:
+            while to.check():
+                sleep(0.4)
```
c1728p9: Rather than actually sleeping here you might want to mock out the ```time``` function instead. That way this test has no external dependencies and runs faster.

flit: Ah, good idea!

c1728p9:
@flit ping me when this is updated and I'll trigger testing

PR #304, opened 2017-05-06
Fix for halt on connect and disconnect resume

flit:
The halt on connect and disconnect resume features were broken due to the options not being properly passed down to the `CortexM` instance, and the change to clear the DHCSR upon disconnect. This changeset does just that.

When a core is added in `CoreSightTarget.add_core()`, the halt on connect setting to set on the new core instance. (`add_core()` is expected to be called before the core is inited.)

For disconnect resume, the code in `Board.uninit()` to resume the target was removed. Instead, the resume parameter is passed to the target's `disconnect()` method, which passes it to `CortexM.disconnect()` for each core. The core will not clear DHCSR if resume is not requested.

This change includes a comprehensive test for these features.

### DIFF
--- a/pyOCD/board/board.py
+++ b/pyOCD/board/board.py
@@ -53,22 +53,15 @@ class Board(object):
     def uninit(self, resume=True):
         """
         Uninitialize the board: link and target.
-        This function resumes the target
         """
         if self.closed:
             return
         self.closed = True
 
-        logging.debug("uninit board %s", self)
-        if resume and self.initiated:
-            try:
-                self.target.resume()
-            except:
-                logging.error("target exception during uninit:")
-                traceback.print_exc()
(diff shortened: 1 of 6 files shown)


studavekar:

c1728p9:
This PR is failing on all non-Kinetis devices - lpc1114, lpc4088dm, lpc824, nrf51.

flit:
Thanks. I'll look into it. Do you have any details about the failures?

c1728p9:
It was failing only on the ConnectTest for those targets. I don't have access to the logs now so I can't give much more detail.

flit:
Alright, thanks. I'll see if it fails on the LPC1768.

flit:
CI results: Mac passed, Linux passed except for a device that's currently stuck in bootloader, Windows build machine is nonfunctional and failing.

PR #311, opened 2017-05-23
Fix bug: frdmk82f gdb reported an warning when connecting

Hsiangkai:
gdb could not recognize 'L' in hexadecimal number. In the other hand,
python may generate 'L' in hexadecimal number when using hex() built-in
function. To ensure the hexadecimal number format is conformed to gdb
restriction, I remove 'L' from the end of result generated by hex().

### DIFF
--- a/pyOCD/core/memory_map.py
+++ b/pyOCD/core/memory_map.py
@@ -265,10 +265,10 @@ class MemoryMap(object):
     def getXML(self):
         root = ElementTree.Element('memory-map')
         for r in self._regions:
-            mem = ElementTree.SubElement(root, 'memory', type=r.type, start=hex(r.start), length=hex(r.length))
+            mem = ElementTree.SubElement(root, 'memory', type=r.type, start=hex(r.start).rstrip("L"), length=hex(r.length).rstrip("L"))
             if r.isFlash:
                 prop = ElementTree.SubElement(mem, 'property', name='blocksize')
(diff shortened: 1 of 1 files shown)


c1728p9:
Good catch. Thanks for submitting this PR as a fix.

Before this PR can be merged the contributors agreement must be signed:
https://developer.mbed.org/contributor_agreement/

pyOCD/core/memory_map.py
```diff
@@ -265,10 +265,10 @@ def getIntersectingRegions(self, start, end=None, length=None, range=None):
     def getXML(self):
         root = ElementTree.Element('memory-map')
         for r in self._regions:
-            mem = ElementTree.SubElement(root, 'memory', type=r.type, start=hex(r.start), length=hex(r.length))
+            mem = ElementTree.SubElement(root, 'memory', type=r.type, start=hex(r.start).rstrip("L"), length=hex(r.length).rstrip("L"))
```
Hsiangkai: According to Python documentation, hex() is only used for int or long. It should be enough for memory address and length.

Hsiangkai:
I have signed the contributors agreement. Thanks for your reminder.

pyOCD/core/memory_map.py
c1728p9: Ok cool, then what you have here is good.

studavekar:

PR #319, opened 2017-07-08
GDB server log, command processing, and XML encoding improvements

flit:
- `GDBServerPacketIOThread` and `GDBServer` classes create their own logger instances.
- Using a custom logger format for pyocd-gdbserver that shows the module name and timestamp. The module name is used instead of the logger name since most pyOCD code still uses the root logger.
- Moved to a table-driven gdb command processor in `GDBServer.handleMsg()`. See the `COMMANDS` dictionary that is built in `GDBServer.__init__()`.
- Properly encoding XML data sent back to gdb. Added `GDBServer.escape()`, which is used in `GDBServer.handleQueryXML()`.

### DIFF
--- a/pyOCD/gdbserver/gdbserver.py
+++ b/pyOCD/gdbserver/gdbserver.py
@@ -828,6 +828,14 @@ class GDBServer(threading.Thread):
 
         return data
 
+    def escape(self, data):
+        result = ''
+        for c in data:
+            if c in '#$}*':
+                result += '}' + chr(ord(c) ^ 0x20)
+            else:
+                result += c
+        return result
 
     def getMemory(self, data):
         split = data.split(',')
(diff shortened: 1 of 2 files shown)


pyOCD/gdbserver/gdbserver.py
```diff
@@ -269,6 +271,34 @@ def __init__(self, board, port_urlWSS, options={}):
         self.telnet_console = semihost.TelnetSemihostIOHandler(self.telnet_port, self.serve_local_only)
         self.semihost = semihost.SemihostAgent(self.target_context, io_handler=semihost_io_handler, console=self.telnet_console)
 
+        # Command handler table.
+        self.COMMANDS = {
```
c1728p9: Could you add a comment indicating what each column is?

flit: Yes, of course.

flit:
This last commit prevents a double init when connecting to the gdb server from Eclipse. It makes the 'init' monitor command a no-op, which should be fine since it really wasn't safe to use, anyway.

c1728p9:
Thanks for the update. Don't worry, extra testing isn't a problem.

PR #325, opened 2017-08-05
Basic multicore debug

flit:
This PR adds basic support for debugging multiple cores simultaneously. pyocd-gdbserver automatically creates one GDBServer instance per core. The first core is given the user-specified port number. Additional cores have port numbers incremented from there. To prevent reset requests from multiple connected gdb instances causing havoc, the reset monitor commands are only honoured for core 0.

To debug a multicore device, run pyocd-gdbserver as usual. This will connect to the device, detect the cores, and create the gdb server instances on separate ports. Next, start up two gdb instances and connect to the two gdb server ports. For instance, on a dual core device if you pass 3333 for the port, connect to port 3333 for the first core and port 3334 for the second core.

On many devices, secondary cores are by default held in reset until released by the primary core. Because gdb does not have a concept of a core held in reset, pyOCD will report a core held in reset by telling gdb that there is a single thread with the name "Reset". This is visible if you run the `show threads` gdb command, and will appear in the Eclipe Debug view's list of threads. All register values will be reported as 0 until the core is released from reset.

Usually you want to have the primary core load code for the secondary core, so configure the second core's gdb to not load any code to the target. This is highly device-specific, though, and may depend on whether the secondary core's code is running out of flash or RAM.

### DIFF
--- a/pyOCD/gdbserver/context_facade.py
+++ b/pyOCD/gdbserver/context_facade.py
@@ -143,3 +143,10 @@ class GDBDebugContextFacade(object):
         else:
             return None
 
+    def getTargetXML(self):
+        return self._context.core.getTargetXML()
+
+    def flush(self):
+        self._context.core.flush()
+
+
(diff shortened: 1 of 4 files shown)


flit:
Any feedback on this PR? Is it ready to be merged? Thanks.

c1728p9:
@flit what is the behavior when task aware debugging is turned on? Do they work together?

flit:
@c1728p9 They work "together". Each core is basically handled separately, with its own GDBServer instance. The only thing that's shared is the link.

PR #327, opened 2017-08-10
Add command-line option to limit USB packet count

MaureenHelm:
Addresses USB issues found when running pyOCD in a VirtualBox VM.

Similar to #154, but implemented as a command-line option rather than always limiting the packet count.


cederom:
Hello @MaureenHelm :-) If the packet limit touches `_packet_count` why not call the variable `packet_count_limit` so it reflects the affected area? What is the actual limit size? Always the same? :-)

c1728p9:
Hi @cederom, _packet_count is set based on how many buffers the device reports it has. The variable `_packet_count ` is used elsewhere, so if it is renamed, it should probably be done in a seperate PR.

MaureenHelm:
Hi @c1728p9 , @cederom , is this ready to merge?

PR #329, opened 2017-08-11
Support GDB range step

Hsiangkai:
If target supports range step, GDB will ask for range step by default.
Under range step, GDB will issue 'r start,end' instead of multiple 's'
packets to target stub. It could reduce greatly the traffic between GDB
and target stub, especially step out one line loop.

### DIFF
--- a/pyOCD/core/coresight_target.py
+++ b/pyOCD/core/coresight_target.py
@@ -113,8 +113,8 @@ class CoreSightTarget(Target):
     def halt(self):
         return self.selected_core.halt()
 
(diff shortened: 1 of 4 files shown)


Hsiangkai:
I measured the performance difference of range stepping using the test cases in test/gdb_test.py on my local machine.

Without range stepping:
Average single step time: 0.185346150398

With range stepping:
Average single step time: 0.150705850124

flit:
Closing. These changes were updated and merged in via #626.

PR #333, opened 2017-09-09
rtos: Introduce Zephyr thread awareness

MaureenHelm:
Adds a new ZephyrThreadProvider class modeled after the existing
FreeRTOSThreadProvider and ArgonThreadProvider classes.

Zephyr is highly configurable, and member offsets into the kernel and
thread structures can change from application to application. To support
thread-aware debugging in OpenOCD, Zephyr added a configuration option
to store these offsets within the application. This same option can be
used for pyOCD as well. Enable it by setting CONFIG_OPENOCD_SUPPORT=y in
your Zephyr application.

Floating point stack frames are not yet supported.

Tested with Zephyr v1.9.0 on FRDM-K64F.

### DIFF
--- a/pyOCD/rtos/__init__.py
+++ b/pyOCD/rtos/__init__.py
@@ -17,8 +17,10 @@
 
 from .argon import ArgonThreadProvider
 from .freertos import FreeRTOSThreadProvider
+from .zephyr import ZephyrThreadProvider
 
 RTOS = {
           'Argon' : ArgonThreadProvider,
           'FreeRTOS' : FreeRTOSThreadProvider,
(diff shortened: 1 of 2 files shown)


MaureenHelm:
The CI test passes when I run it locally. Is it possible to rerun the test in Travis or can you please help investigate the failure?

MaureenHelm:
@c1728p9 @cederom @flit Is there any interest from the maintainers for this feature? How do we get it into a mergeable state?

flit:
I think it is a valuable addition. Float frames would need to be supported before it could be merged, imo.

flit:
Closed due to creation of #364.

PR #336, opened 2017-10-21
Board listing improvements

flit:
- Moved board ID table into a separate file for easier maintenance.
- Improved the command line board selection interface by colorizing. Listing is also sorted by board name, so the boards always have the same numbers.
- Board ID matching for the "--board" argument for tools uses a contains match now instead of requiring the entire board ID.


flit:
Strange… the Travis build segfaulted during the semihosting test.

c1728p9:
Hi @flit I retriggered the travis job twice and it segfaulted both times, so this isn't an intermittent failure.

flit:
Very odd. It's not just this PR that is failing with the segfault, it seems to be every recent build.

PR #340, opened 2017-12-01
target: fix flash, ram size and identifier for KW41Z512xxx4

jfischer-phytec-iot:
The size of the Flash and RAM is wrong for target MKW41Z512xxx4,
fix that and use kw41z512xxx4 identifier for this target.


jfischer-phytec-iot:
@MaureenHelm FYI

c1728p9:
CC @mmahadevan108

flit:
It's true the flash and RAM sizes are wrong. Looks like they were copied from KW40 and not updated.

Why do you want to change the target name to 'kw41z512xxx4'? This would be unlike all the other targets in pyOCD.

jfischer-phytec-iot:
>Why do you want to change the target name to 'kw41z512xxx4'? This would be unlike all the other targets in pyOCD.


Regarding the Flash and RAM configuration, there are two variants of the SoC, one with 256KB Flash / 64KB RAM and 512KB Flash / 128KB RAM. Probably `kw41z512` is also clear enough.

pyOCD/target/__init__.py
```diff
@@ -78,7 +78,7 @@
           'kv11z7': target_MKV11Z128xxx7.KV11Z7,
           'kw01z4': target_MKW01Z128xxx4.KW01Z4,
           'kw40z4': target_MKW40Z160xxx4.KW40Z4,
-          'kw41z4': target_MKW41Z512xxx4.KW41Z4,
+          'kw41z512xxx4': target_MKW41Z512xxx4.KW41Z4,
```
c1728p9: The name should be kept the same for now. If there is a switch to a new naming scheme, all the devices will be updated at the same time.

flit:
We tend to directly support only the maximally-sized device variant, rather than attempting to include configurations for each. The target will work for all variants with smaller memory sizes, which is why the target name doesn't specify the memory size. This lets us support more devices with fewer targets.

When we eventually switch to CMSIS-Pack based configurations, we will be able to support all device variants directly.

jfischer-phytec-iot:
ok, will rework and change the name back to 'kw41z4'

flit:
Closing this PR because the fixes have been incorporated into #336.

PR #349, opened 2018-03-15
Correct Flash Python init API to match C API

orenc17:
* Correct init API
* Add uninit API
* Calls to program and erase are wrapped with init and uninit

@c1728p9


pyOCD/flash/flash_builder.py
```diff
@@ -573,8 +588,15 @@ def _page_erase_program_double_buffer(self, progress_cb=_stub_progress):
...
                 current_addr = page.addr
                 current_weight = page.getEraseProgramWeight()
+
+                self.flash.init(fnc=self.flash.INIT_FUNC_ERASE)
                 self.flash.erasePage(current_addr)
+                self.flash.unInit(self.flash.INIT_FUNC_ERASE)
+
+                self.flash.init(fnc=self.flash.INIT_FUNC_PROGRAM)
                 self.flash.startProgramPageWithBuffer(current_buf, current_addr) #, erase_page=True)
+                self.flash.unInit(self.flash.INIT_FUNC_PROGRAM)
```
c1728p9: You'll need to wait for the program operation to complete before uninit is called. If you move uninit to after self.flash.waitForCompletion() on line 609 it should be good.

pyOCD/flash/flash.py
```diff
@@ -105,20 +109,42 @@ def __init__(self, target, flash_algo):
...
         # update core register to execute the init subroutine
-        result = self.callFunctionAndWait(self.flash_algo['pc_init'], init=True)
+        result = self.callFunctionAndWait(self.flash_algo['pc_init'], 
+                                          r0=adr, r1=clk, r2=fnc, init=True)
 
         # check the return code
         if result != 0:
             logging.error('init error: %i', result)
 
+    def unInit(self, fnc):
```
flit: Can you change "unInit" to "uninit" (single word, no camel-case)?

pyOCD/flash/flash.py
```diff
@@ -75,6 +75,10 @@ class Flash(object):
     """
     This class is responsible to flash a new binary in a target
     """
+    INIT_FUNC_ERASE = 1
+    INIT_FUNC_PROGRAM = 2
+    INIT_FUNC_VERIFY = 3
```
flit: Spell out "INIT_FUNCTION" for all three constants instead of shorting to "INIT_FUNC".

flit:
I'm concerned about the impact this will have on performance for the cases where the FlashBuilder is switching back and forth between erase and program. Each execution of a flash algo function adds noticeable latency.

Do you have a specific need for the init function to be invoked with the operation to be performed?

c1728p9:
> Do you have a specific need for the init function to be invoked with the operation to be performed?

This is a requirement of the flash algo specification. If it is not done then some cmsis-pack flash algos won't work work correctly. Documentation on the init function can be found in the [cmsis-pack documentation](http://www.keil.com/pack/doc/CMSIS/Pack/html/algorithmFunc.html#Init).

pyOCD/flash/flash_builder.py
orenc17A: fixed

pyOCD/flash/flash.py
orenc17A: fixed

pyOCD/flash/flash.py
orenc17A: fixed

orenc17A:
[automated_test_result.txt](https://github.com/mbedmicro/pyOCD/files/1822708/automated_test_result.txt)
[automated_test_summary.txt](https://github.com/mbedmicro/pyOCD/files/1822709/automated_test_summary.txt)

Automated tests results on K64F

orenc17A:
[automated_test_result.txt](https://github.com/mbedmicro/pyOCD/files/1825733/automated_test_result.txt)
[automated_test_summary.txt](https://github.com/mbedmicro/pyOCD/files/1825735/automated_test_summary.txt)

Tests results for nrf52

alzix:
>I'm concerned about the impact this will have on performance for the cases where the FlashBuilder is switching back and forth between erase and program. Each execution of a flash algo function adds noticeable latency.

@orenc17A , could you please provide time measurements before and after the change?
in addition i see test results contain performance tests - could you also upload the results before the change to emphasize the impact

pyOCD/flash/flash.py
```diff
@@ -105,20 +110,42 @@ def __init__(self, target, flash_algo):
     def minimumProgramLength(self):
         return self.min_program_length
 
-    def init(self):
+    def init(self, adr=None, clk=0, fnc=None):
         """
         Download the flash algorithm in RAM
         """
+        if adr is None:
```
alzix: what cases do you anticipate for the `adr` and `clk` to be passed explicitly?
if none - please consider removing them

pyOCD/flash/flash.py
```diff
@@ -105,20 +110,42 @@ def __init__(self, target, flash_algo):
...
 
         # check the return code
         if result != 0:
             logging.error('init error: %i', result)
 
+    def uninit(self, fnc):
+        """
+        Download the flash algorithm in RAM
+        """
+        assert fnc in self.INIT_FUNCTIONS
```
alzix: the API is symmetric, why do we pass an operation to uninit. We do not support multi-threaded execution as i can see.

pyOCD/flash/flash.py
```diff
@@ -105,20 +110,42 @@ def __init__(self, target, flash_algo):
     def minimumProgramLength(self):
         return self.min_program_length
 
-    def init(self):
+    def init(self, adr=None, clk=0, fnc=None):
```
alzix: these init uninit functions look as a best match for context management using the `with` clause

pyOCD/flash/flash_builder.py
```diff
@@ -252,6 +250,7 @@ def program(self, chip_erase=None, progress_cb=None, smart_flash=True, fast_veri
                 flash_operation = self._page_erase_program(progress_cb)
 
         self.flash.target.resetStopOnReset()
+        self.flash.init_called_num = 0
```
alzix: why do we manage this counter from outside of the class?
what is the rule of thumb here? `self.flash.target.setTargetState("PROGRAM")` should be called once after `self.flash.target.resetStopOnReset()`? so why not calling it directly? It seems that `FlashBuilder` knows about `flash.target` already

pyOCD/tools/flash_tool.py
```diff
@@ -200,6 +201,7 @@ def main():
                         print("Erasing sector 0x%08x" % page_addr)
                         flash.erasePage(page_addr)
                         page_addr += page_info.size
+                    flash.unInit(flash.INIT_FUNCTION_ERASE)
```
flit: "unInit" misspelled.

flit:
Closing this PR as it is way out of date. New issue #486 created, and I'm cleaning up the changes from this PR in advance of #357.

PR #352, opened 2018-04-23
Fixed MemoryMap for K82F, inaccessible RamRegion was defined that cau…

linlingao:
…sed DAPLink MSD test to fail

### DIFF
--- a/pyOCD/target/target_MK82FN256xxx15.py
+++ b/pyOCD/target/target_MK82FN256xxx15.py
@@ -91,19 +91,9 @@ class Flash_k82f25615(Flash_Kinetis):
 class K82F25615(Kinetis):
 
     memoryMap = MemoryMap(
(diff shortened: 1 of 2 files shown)


flit:
@linlingao How does DAPLink MSD test access the RamRegions? I would much rather fix this by changing the test so it doesn't attempt to access regions without backing physical memory. Maybe also change pyOCD to explicitly call out internal RAMs.

The reason for wanting to address this issue differently is that the memory regions in CMSIS-Packs may include all architectural memory regions, whether or not there is a physical memory backing that region.

linlingao:
@flit DAPLink finds the first region with the type == 'ram' in memory_map. 
  ram_regions = [region for region in memory_map if region.type == 'ram']
  ram_region = ram_regions[0]
 we can come up with a way to distinguish the internal RAM from the rest in pyOCD, it would work. Keep in mind that if you add an identifier to this target, you would have to add identifier to all targets.

PR #359, opened 2018-06-11
Fix stale rtos

flit:
This change set provides a workaround for stale RTOS information caused by previous firmware builds leaving valid but stale data in RAM across a reset. It also brings in a simple notification system for target events.


MaureenHelm:
I presume you'd like to see this implemented for zephyr too?

flit:
@MaureenHelm Yup! 😄

PR #364, opened 2018-06-22
rtos: Introduce Zephyr thread awareness (rebased)

MaureenHelm:
Rebased #333 to the latest master and added an event handler per #359. Updated the commit message to reflect testing with a newer version of Zephyr.


flit:
Thanks, @MaureenHelm !

PR #370, opened 2018-07-05
Corrected board name.

linlingao:


linlingao:
@marcuschangarm @flit Marcus pointed out the board name wasn't right, hence this PR.

PR #374, opened 2018-07-15
Python 3 support

flit:
Numerous changes in order to support running pyOCD in Python 3.

The biggest areas of change are:

1. The most involved changes are in the gdb server, where all `str` objects were changed to `bytes`. 
2. The other most noticeable change is that all module imports within pyOCD must now be relative.
3. Replaced all uses of `has_key()` with the `in` or `not in` operators.

All tests pass both on Python 2.7.15 and 3.6.5. Travis CI config is updated to test on 2.7, 3.6.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,6 @@ setup(
             'pyocd-tool = pyOCD.tools.pyocd:main',
         ],
     },
-    use_2to3=True,
     packages=find_packages(),
     include_package_data=True,  # include files from MANIFEST.in
 )
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,6 @@ if sys.platform.startswith('linux'):
 elif sys.platform.startswith('win'):
     install_requires.extend([
         'pywinusb>=0.4.0',
-        'hidapi',
     ])
 elif sys.platform.startswith('darwin'):
     install_requires.extend([
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,15 @@
 from setuptools import setup, find_packages
 import sys
 
-install_requires = ['intelhex', 'six', 'enum34', 'future', 'websocket-client', 'intervaltree', 'colorama']
+install_requires = [
+    'intelhex',
+    'six',
+    'enum34',
+    'future',
+    'websocket-client',
+    'intervaltree',
+    'colorama',
+    ]
 if sys.platform.startswith('linux'):
     install_requires.extend([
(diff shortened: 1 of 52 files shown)


setup.py
```diff
@@ -18,14 +18,23 @@
...
+    'colorama',
+    ]
 if sys.platform.startswith('linux'):
     install_requires.extend([
         'pyusb>=1.0.0b2',
     ])
 elif sys.platform.startswith('win'):
     install_requires.extend([
         'pywinusb>=0.4.0',
+        'hidapi',
```
c1728p9: Why is `hidapi` needed on windows? Shouldn't `pywinusb` cover windows?

test/automated_test.py
```diff
@@ -99,7 +99,7 @@ def main():
     test_time = (stop - start)
 
     print_summary(test_list, result_list, test_time)
-    with open(summary_file, "wb") as output_file:
+    with open(summary_file, "w") as output_file:
```
c1728p9: was this causing problems or just changed since it isn't a binary file?

test/gdb_script.py
```diff
@@ -69,16 +77,16 @@
     "help",
     "help reset",
     "help halt",
-    "help resume",
     "init",
     "reset",
     "reset halt",
-    "reset halt resume",
```
c1728p9: why were these removed?

test/gdb_script.py
```diff
@@ -96,11 +104,16 @@
 }
 
 
+def gdb_execute(cmd):
+    print("Executing command:", cmd)
```
c1728p9: Was this intentionally left in or is it just for debugging?

setup.py
flit: It was added because there are issues with pywinusb on some boards.
Will remove and deal with separately.

test/automated_test.py
flit: In Python 3, files open in binary mode only accept bytes-like objects in their `write()` method, and return `bytes` from `read()`.

test/gdb_script.py
flit: The "resume" monitor commands were disabled back in 2015.

test/gdb_script.py
flit: Intentional. It's helpful to have the gdb commands listed in the `output.txt` file.

PR #387, opened 2018-08-06
Changes to CC3220SF to make the reset work properly

dragoniteArm:


flit:
Please rebase and squash to clean up your commits. This needs to be done before this PR can be merged. Thanks!

dragoniteArm:
rebased and squashed

flit:
@dragoniteArm Thanks for rebasing and squashing!

I'm afraid I just merged in some init sequence changes (#385) that affect this PR. Instead of overriding `init()` in your `CC3220SF` class, you'll have to override `create_init_sequence()`. See `pyOCD.target.family.target_kinetis.Kinetis.create_init_sequence()` for an example. In your case, replace the "create_cores" task (use `CallSequence.replace_task()`) with your own that manually creates the `CC3220SF_CortexM` instance. You don't need the `setHaltOnConnect()` or `if busAccessible` bits anymore.

Note that the ctor params for `CortexM` have changed by removing the dp param.

PR #390, opened 2018-08-08
Clean up rtos logging

MaureenHelm:
- Cleans up rtos modules to consistently use the debug logging level (there were a few stray usages of the info logging level)
- Eliminates unrelated rtos logging messages (e.g., freertos logging messages when debugging a zephyr application)

### DIFF
--- a/pyOCD/rtos/argon.py
+++ b/pyOCD/rtos/argon.py
@@ -328,9 +328,6 @@ class ArgonThreadProvider(ThreadProvider):
         self._all_threads = None
         self._threads = {}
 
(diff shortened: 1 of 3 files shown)


flit:
Hi @MaureenHelm , thanks for the PR! You can just completely remove the debug logs about invalidating thread lists.

MaureenHelm:
> You can just completely remove the debug logs about invalidating thread lists.

They did help me debug a problem when I was rebasing #364 so I think they're worth keeping.

flit:
Ok, that's fine. I'll have to be careful when I bring in one of my branches, because I had already removed the FreeRTOS 'invalidating' log (it was done before #364 was merged in).

PR #394, opened 2018-08-14
RTX5: Return None as a current thread, when not yet running

SeppoTakalo:


SeppoTakalo:
@c1728p9 @flit Can you guys try if this helps on the problem that you had?

flit:
Just need @c1728p9 to verify the fix now.

PR #395, opened 2018-08-16
List supported boards

arekzaluski:
List supported boards and print them in JSON format.


thegecko:
@flit Can we have some feedback for this PR?

thegecko:
Can this be merged, now?

PR #397, opened 2018-08-17
Add specific Python versions supported to setup.py

johnthagen:
This explicitly lists support for which version of Python the package supports. This helps new users when browsing PyPI know if their Python version supports the package.

I only added 2.7 and 3.6 due to those being the only ones tested in `.travis.yml`. If support for Python 3.4, 3.5, or 3.7 is also considered supported, `.travis.yml` could be updated to add those and they could be called out in `setup.py` as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,6 +58,10 @@ setup(
         "Development Status :: 4 - Beta",
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


flit:
Hi @johnthagen, have you signed the [Contributor Agreement](https://os.mbed.com/contributor_agreement/)? If not, could you please create an account on https://os.mbed.com/ and signed it?

johnthagen:
@flit I may have but can't remember. I have an account on os.mbed.com but the password reset form (https://os.mbed.com/account/reset_password/) isn't working for me, but having it send me my username does. Strange. 😕

flit:
It looks like you haven't signed the Contributor Agreement. I'll forward your troubles with the password reset form to the web team.

PR #409, opened 2018-09-21
Fix LPC546XX Erase chip function

mmahadevan108:
The erase chip function was not erasing the whole flash.


mmahadevan108:
cc @flit

PR #413, opened 2018-09-27
Fix exception when core type cannot be read

c1728p9:
Log "unknown" for core type when it cannot be read rather than raising a KeyError. This makes it possible to run scripts on locked ST targets.


c1728p9:
CC @brianesquilona

PR #422, opened 2018-10-05
STLinkV2 debug probe support

flit:
This patch adds support for connecting to STLinkV2 and V2-1 debug probes. The minimum STLink firmware version is V2J24. V3 is not fully supported yet.

STLinkV2 probes will show up and can be used just like CMSIS-DAP probes. The main difference is that you currently need to manually specify the target type using the `--target` option on command line tools (see the first known issue below).

Here is example output from `pyocd-tool list` showing both a DAPLink and STLinkV2 probe:
```
## => Board Name | Unique ID
-- -- ----------------------
 0 => FRDM-K64F [k64f] | 0240000029164e4500440012706e0007f301000097969900
 1 => STM32 STLinkV2-1 | 066EFF555051897267233656
```

Known issues:
- Currently, the board type cannot be automatically detected like it can with the DAPLink firmware. This will change when mbed-ls functionality is integrated to read board ID from the HTML file on the STLinkV2-1 USB MSC volume.
- Timeouts have been reported while debugging.
- `cortex_test.py` memory accesses fail for an unknown reason.
- `flash_test.py` fails for some ST devices. It's unclear whether this is related to STLink or the target support.
- (There are probably more issues.)

The purpose of this PR is to make the STLinkV2 support available as soon as feasible so the community can test and provide feedback. Existing functionality with CMSIS-DAP probes should not be affected in any way.

pyusb is now a dependency on all operating systems. However, libusb does not get installed automatically via pip dependency management. How to install libusb depends on your OS:
- Linux: should already be installed.
- macOS: use Homebrew: `brew install libusb`
- Windows: download libusb from [libusb.info](http://libusb.info) and place the DLL in your Python installation folder next to python.exe.

### DIFF
--- a/pyOCD/board/board.py
+++ b/pyOCD/board/board.py
@@ -17,6 +17,7 @@
 
 from ..target import (TARGET, FLASH)
 import logging
+import json
 import six
 
 log = logging.getLogger('board')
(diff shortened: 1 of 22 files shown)


flit:
Thanks a lot @c1728p9!

PR #423, opened 2018-10-08
enum34 is only needed for Python2

codehelp:
When building for Python3, omit the dependency
on support backported to Python2 from Python3.4

Signed-off-by: Neil Williams <codehelp@debian.org>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ import sys
 install_requires = [
     'intelhex',
     'six',
-    'enum34',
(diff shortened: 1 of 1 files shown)


flit:
Hi @codehelp, thanks for the PR. I wasn't even aware the marker syntax had been added! So, I'm going to update setup.py to exclusively use markers for OS-specific dependencies, and will include your enum34 change as well. As such, I'm going to close this PR, but your change will be merged in shortly.

PR #435, opened 2018-10-22
Thread info: reduce clutter

kjbracey-arm:
Remove unnecessary or inaccurate information from thread XML. This
improves the clarity of GDB command line or Eclipse UI view.

* Thread providers don't give core information, so remove `core="0"`.
* Remove "handle" string - this is just duplicating the ID, and if
  it is provided, it is supposed to be a byte string, not an integer,
  so we would have often been malformed.
* Do not include thread name in description - this is just repeating
  the name supplied via attribute. It's not unreasonable to expect
  users to have a post-2015 GDB if they want to see thread names.


pyOCD/gdbserver/gdbserver.py
```diff
@@ -1280,21 +1280,15 @@ def getThreadsXML(self):
...
                 hexId = "%x" % thread.unique_id
-                t = SubElement(root, 'thread', id=hexId, core="0", name=thread.name, handle=hexId)
-
-                desc = thread.description
-                if desc:
-                    desc = thread.name + "; " + desc
-                else:
-                    desc = thread.name
-                t.text = desc
+                t = SubElement(root, 'thread', id=hexId, name=thread.name)
```
flit: Add back in the core attribute and I'll be happy:

```suggestion
                t = SubElement(root, 'thread', id=hexId, name=thread.name, core=str(self.core))
```

While thread providers don't provide core information (SMP is not supported), the gdbserver knows the core to which it is attached. The actual core number should have been included, but that got missed somehow.

pyOCD/gdbserver/gdbserver.py
```diff
@@ -1280,21 +1280,15 @@ def getThreadsXML(self):
         if not self.is_threading_enabled():
             t = SubElement(root, 'thread', id="1", core=str(self.core))
             if self.is_target_in_reset():
-                t.text = "Reset"
+                t.set('name', "Reset")
             else:
-                t.text = "Thread mode"
+                t.set('name', "Thread mode")
```
flit: Eventually we need to properly report all core states when there isn't a thread provider, such as handler mode, sleeping, lockup. (Just thinking out loud. 😄)

pyOCD/gdbserver/gdbserver.py
kjbracey-arm: Still seems like clutter to me. If not SMP, it's not per-thread info. Having it present makes Eclipse put "[core 0]" on every line, which is a bit like putting "[pid 720]" on every line. You should already know which CPU/process you're looking at from context or UI indication (image name/connection name/port number etc).

No GDB UI should be mixing threads from different processes/remotes so that you would have to distinguish them like this.

pyOCD/gdbserver/gdbserver.py
kjbracey-arm: Yes, and I realise this change is wrong - that isn't the thread name, it's the extra info, so should stay in the text. Doesn't make sense to have a thread name for the single thread.

I'll take out the "Thread mode" for now, as it's currently wrong, and leave the "Reset" as description.

A follow-up to add state from IPSR as annotation in place of "Thread mode" (Thread/HardFault/SVCall/Interrupt 23...) would make sense.

If the logic of #430 was followed through to give 2 threads for non-OS, then they would have the names "Main stack" and "Process stack". They could both retain the same IPSR-based annotation (although for Process it would always be Thread).

kjbracey-arm:
Adjusted - don't touch the Reset/Thread mode bit, which is description. Made a separate PR #437 for the suggestion to put useful info there.

Persisting on the "don't keep telling GDB our only core" thing, pushing it even harder this time :)

flit:
pyOCD supports multicore targets; thus we need to report the core to gdb. It's probably ok to not report the core if the target is single-core. But it definitely needs to be there for multicore targets.

kjbracey-arm:
MULTICORE_DEBUG.md:
> It does this by serving one gdb server per core, to which you connect independant gdb instances.

There's no ambiguity here - the user must have created this debugging session, and must have connected one GDB to one remote target per core.

In Eclipse, at least, putting the "core" info in prompts it to put a "[cores: 0]" after the name of the process, which may be a teeny bit useful, but that's not actually adding new information, and it also adds the per-thread clutter.

Maybe the process annotation is a helpful label for someone who's failed to give their debug session a meaningful name. But they could give it a much more meaningful name than "[cores: 0]", given that PyOCD's core numbering is just an instance index, and possibly doesn't even match HW documentation.

Current state permits:

* useless session name [GDB PyOCD Debugging]
* * useless image name.elf [cores: 0]
* * * Thread #2 [main] (Running) [core: 0] (Suspended:Breakpoint)
* * * Thread #3 [idle] (Ready) [core: 0] (Suspended: Container)
* useless session name [GDB PyOCD Debugging]
* * useless image name.elf [cores: 1]
* * * Thread #1 [core: 1] (Running: User Request)

with the user apparently relying on the "cores:" to figure out which one is which (?). 

If you were debugging separate processes on a normal OS you wouldn't be able to rely on "cores:" to figure out which was which. You'd do make sure they had sensible names:

* Main Cortex-M7 [GDB PyOCD Debugging]
* * main_image.elf
* * * Thread #2 [main] (Running) Suspended:Breakpoint)
* * * Thread #3 [idle] (Ready) (Suspended: Container)
* Support Cortex-M0 [GDB PyOCD Debugging]
* * support.elf
* * * Thread #1 (Running: User Request)

Isn't that sufficient?

On the other hand, the "core" information isn't actually wrong, and arguably we should be expecting the Eclipse UI to declutter itself. It could heuristically not show "core: X" for a process with all threads indicating the same core, and remove the "core: X" line from each thread. That would satisfy me.

From the other extreme - can you even persuade the GDB command line to show the core info? Can't find it.

flit:
You've _mostly_ won the argument. 😜

But I think I'd like to control this via a session option. By default the core attributes would not be set, but if needed for some reason a user could turn on the session option.

As part of the final solution, `get_t_response()` needs to report the correct core instead of always core 0.

kjbracey-arm:
Updated with session option as per review.

kjbracey-arm:
Minor fix - was missing an `.encode()`, so the core number output asserted on Python 3.

PR #436, opened 2018-10-23
Correct IPSR masking

kjbracey-arm:
IPSR is bottom 9 bits of xPSR, not 8.

### DIFF
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -56,6 +56,12 @@ CORE_TYPE_NAME = {
 # is necessary. The values are the byte number containing the register value, plus 1 and then
 # negated. So -1 means a mask of 0xff, -2 is 0xff00, and so on. The actual DCRSR register index
 # for these combined registers has the key of 'cfbp'.
+#
+# XPSR is always read in its entirety via the debug interface, but we also provide
+# aliases for the submasks APSR, IPSR and EPSR. These are encoded as 0x10000 plus 3 lower bits
+# indicating which parts of the PSR we're interested in - encoded the same way as MRS's SYSm.
+# (Note that 'XPSR' continues to denote the raw 32 bits of the register, as per previous versions,
+# not the union of the three APSR+IPSR+EPSR masks which don't cover the entire register).
 CORE_REGISTER = {
                  'r0': 0,
                  'r1': 1,
@@ -77,6 +83,12 @@ CORE_REGISTER = {
                  'pc': 15,
                  'r15': 15,
                  'xpsr': 16,
+                 'apsr': 0x10000,
+                 'iapsr': 0x10001,
+                 'eapsr': 0x10002,
+                 'ipsr': 0x10005,
+                 'epsr': 0x10006,
+                 'iepsr': 0x10007,
                  'msp': 17,
                  'psp': 18,
                  'cfbp': 20,
@@ -127,6 +139,21 @@ def register_name_to_index(reg):
             raise KeyError('cannot find %s core register' % reg)
     return reg
 
+# Returns true for registers holding single-precision float values
+def is_float_register(index):
+    return index >= 0x40 and index <= 0x5f
+
+# Generate a PSR mask based on bottom 3 bits of a MRS SYSm value
+def sysm_to_psr_mask(sysm):
+    mask = 0
+    if (sysm & 1) != 0:
+        mask |= CortexM.IPSR_MASK
+    if (sysm & 2) != 0:
+        mask |= CortexM.EPSR_MASK
+    if (sysm & 4) == 0:
+        mask |= CortexM.APSR_MASK
+    return mask
+
 class CortexM(Target, CoreSightComponent):
 
     """
@@ -136,6 +163,11 @@ class CortexM(Target, CoreSightComponent):
        - read/write core registers
        - set/remove hardware breakpoints
     """
+
+    # Program Status Register
+    APSR_MASK = 0xF80F0000
+    EPSR_MASK = 0x0700FC00
+    IPSR_MASK = 0x000001FF
 
     # Debug Fault Status Register
     DFSR = 0xE000ED30
@@ -657,7 +689,7 @@ class CortexM(Target, CoreSightComponent):
         regIndex = register_name_to_index(reg)
         regValue = self.read_core_register(regIndex)
         # Convert int to float.
-        if regIndex >= 0x40:
+        if is_float_register(regIndex):
             regValue = conversion.u32_to_float32(regValue)
         return regValue
 
@@ -685,7 +717,7 @@ class CortexM(Target, CoreSightComponent):
         for reg in reg_list:
             if reg not in CORE_REGISTER.values():
                 raise ValueError("unknown reg: %d" % reg)
-            elif ((reg >= 0x40) or (reg == 33)) and (not self.has_fpu):
+            elif (is_float_register(reg) or (reg == 33)) and (not self.has_fpu):
                 raise ValueError("attempt to read FPU register without FPU")
 
         # Begin all reads and writes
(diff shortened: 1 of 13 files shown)


kjbracey-arm:
First commit is minimal patch correction.

Second commit is an attempt to not repeat the masking everywhere. Don't know if style suits.

pyocd/coresight/cortex_m.py
```diff
@@ -127,6 +127,9 @@ def register_name_to_index(reg):
             raise KeyError('cannot find %s core register' % reg)
     return reg
 
+def get_ipsr(context):
+    return context.readCoreRegister('xpsr') & 0x1ff
```
flit: `readCoreRegister` needs to be updated to `read_core_register`.

I'd also like to see a named constant for 0x1ff. For now it can go into `CortexM` (I'm probably going to extract the constants in `CortexM` into a constants module at some point).

pyocd/flash/flash.py
```diff
@@ -381,7 +381,7 @@ def wait_for_completion(self):
             expected_flash_algo = self.flash_algo['instructions']
             if self.use_analyzer:
                 expected_analyzer = analyzer
-            final_ipsr = self.target.read_core_register('xpsr') & 0xff
+            final_ipsr = self.target.read_core_register('xpsr') & 0x1ff
```
flit: You can use `get_ipsr()` here, too, passing in `self.target`.

pyocd/coresight/cortex_m.py
kjbracey-arm: I'm tend to be of the school that if a constant is used in exactly one place, then you don't need a define for it, but will look neat if I add it alongside the other PSR masks.

pyocd/coresight/cortex_m.py
```diff
@@ -127,6 +127,9 @@ def register_name_to_index(reg):
             raise KeyError('cannot find %s core register' % reg)
     return reg
 
+def get_ipsr(context):
```
flit: I'm wondering if this should be a method on `DebugContext` rather than a standalone function. This would fit the pyOCD style a little better, I think. The main issue is that you'd need duplicate methods on `Target` and/or `CoreSightTarget` and/or `CortexM`. But then, having these duplicates is also kind of pyOCD-style, currently! (I'm thinking of making `Target` derive from `DebugContext`, btw.)

kjbracey-arm: Not so keen on that - you have a perfectly viable "read register" API that doesn't feel like it wants to be worked around.  To make it feel more aligned, I would be more inclined to treat PSR submasks like `cfbp`. Each sub-PSR is readable by name using `MRS` in the instruction set, so would make sense to have them readable via the named register API.

pyocd/debug/cache.py
```diff
@@ -120,10 +131,15 @@ def read_core_registers_raw(self, reg_list):
         # Read uncached registers from the target.
         read_list = list(reg_set.difference(cached_set))
         reading_cfbp = any(r for r in read_list if r in self.CFBP_REGS)
+        reading_xpsr = not reading_cfbp and any(r for r in read_list if r in self.XPSR_REGS)
```
flit: What is the `not reading_cfbp` term for?

kjbracey-arm: Me pointlessly micro-optimising to make it a sort of "else", saving another list search. Will drop.

pyocd/coresight/cortex_m.py
```diff
@@ -766,6 +800,10 @@ def write_core_registers_raw(self, reg_list, data_list):
                 specialRegValue = self.read_core_register(CORE_REGISTER['cfbp'])
                 break
 
+            if (reg >= 0x10000) and (reg <= 0x10007):
+                specialRegValue = self.read_core_register(CORE_REGISTER['xpsr'])
```
kjbracey-arm: Just realised these two aren't exclusive for a list - need to be remembering xpsr and cfbp separately.

flit:
Nice, I really like the new approach! Thanks.

I'm not sure what the exact fix is for the failed Travis CI tests, but it's related to the `CORE_REGS_NO_CFBP` constant in `pyocd/test/test_regcache.py`.

Whilst you're working on `pyocd/test/test_regcache.py`, could you add some more tests for the xPSR code?

kjbracey-arm:
Caching tests fixed and extended

flit:
The flash test is failing in CI on targets without an FPU. Specifically nRF51 and KL82Z.

This is the exception:

```
Exception attempt to read FPU register without FPU when testing board 9900000041454e450041600a0000006a0000000097969901
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/test/flash_test.py", line 104, in run
    result = self.test_function(board.unique_id)
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/test/flash_test.py", line 182, in flash_test
    info = flash.flash_block(addr, data, False, False, progress_cb=print_progress())
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/flash/flash.py", line 310, in flash_block
    info = fb.program(chip_erase, progress_cb, smart_flash, fast_verify)
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/flash/flash_builder.py", line 224, in program
    self.flash.init()
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/flash/flash.py", line 140, in init
    result = self._call_function_and_wait(self.flash_algo['pc_init'], init=True)
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/flash/flash.py", line 423, in _call_function_and_wait
    return self.wait_for_completion()
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/flash/flash.py", line 384, in wait_for_completion
    final_ipsr = self.target.read_core_register('ipsr')
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/core/coresight_target.py", line 210, in read_core_register
    return self.selected_core.read_core_register(reg)
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/coresight/cortex_m.py", line 698, in read_core_register
    vals = self.read_core_registers_raw([reg])
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/coresight/cortex_m.py", line 717, in read_core_registers_raw
    raise ValueError("attempt to read FPU register without FPU")
ValueError: attempt to read FPU register without FPU
```

You can run the flash test locally by:

```
$ cd test
$ python ./flash_test.py
```

Ideally, we'd update the regcache test to verify functionality both with and without an FPU, as simulated by `MockCore`. But that can be done later since it's more complex than just adding test cases (added to todo list).

kjbracey-arm:
No non-FPU test platforms handy, but think I've fixed the issue - tightened up the "is this a FP register" checks throughout. Squashed the last two commits.

pyocd/coresight/cortex_m.py
```diff
@@ -127,6 +139,21 @@ def register_name_to_index(reg):
             raise KeyError('cannot find %s core register' % reg)
     return reg
 
+# Returns true for registers holding single-precision float values
+def is_float_register(index):
+    return index >= 0x40 and index <= 0x5f
```
flit: Haha, hate to say it, but you opened the door… 😁 Could we have functions similar to `is_float_register()` for `is_cfbp_subregister()` and `is_xpsr_subregister()`? I'll do the same on the [dev_cm7](https://github.com/flit/pyOCD/tree/dev_cm7) branch with the double-precision float registers.

Eventually it might be good to change this so each register is represented by an object that knows how to perform the special logic associated with it, if any. I've never been too fond of these magic numbers.

kjbracey-arm: Yes, sure.

flit:
Perfect! 👍

PR #437, opened 2018-10-23
Show active exception or interrupt in thread description

kjbracey-arm:
Implementation is straightforward, just not totally certain where each bit should live in the structure.

### DIFF
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -131,6 +131,20 @@ CORE_REGISTER = {
                  's31': 0x5f,
                  }
 
+CORE_EXCEPTIONS = {
+    0 : "Thread mode",
+    1 : "Reset",
+    2 : "NMI",
+    3 : "HardFault",
+    4 : "MemManageFault",
+    5 : "BusFault",
+    6 : "UsageFault",
+    11 : "SVCall",
+    12 : "DebugMon",
+    14 : "PendSV",
+    15 : "SysTick",
+}
+
 def register_name_to_index(reg):
     if isinstance(reg, str):
         try:
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -131,20 +131,6 @@ CORE_REGISTER = {
                  's31': 0x5f,
                  }
 
-CORE_EXCEPTIONS = {
-    0 : "Thread mode",
-    1 : "Reset",
-    2 : "NMI",
-    3 : "HardFault",
-    4 : "MemManageFault",
-    5 : "BusFault",
-    6 : "UsageFault",
-    11 : "SVCall",
-    12 : "DebugMon",
-    14 : "PendSV",
-    15 : "SysTick",
-}
-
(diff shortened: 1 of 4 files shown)


kjbracey-arm:
IRQ names would make it even more awesome. So, do you want me to incorporate that commit into this PR now?

kjbracey-arm:
Added your commit as-is, then a fixup commit combining the two.

Now shows CMSIS SVD interrupt name.

pyocd/coresight/cortex_m.py
```diff
@@ -937,3 +894,39 @@ def get_target_context(self, core=None):
...
+    # Names for built-in Exception numbers found in IPSR
+    CORE_EXCEPTION = [
+           "Thread",
+           "Reset",
+           "NMI",
+           "HardFault",
+           "MemManage",
+           "BusFault",
+           "UsageFault",
+           "Exception 7",
```
flit: Should go ahead and add SecureFault.

pyocd/coresight/cortex_m.py
```diff
@@ -937,3 +894,39 @@ def get_target_context(self, core=None):
...
+
+    def exception_number_to_name(self, exc_num, name_thread=False):
+        if exc_num < len(self.CORE_EXCEPTION):
+            if exc_num == 0 and not name_thread:
+                return None
+            else:
+                return self.CORE_EXCEPTION[num]
+        else:
+            irq_num = exc_num - len(self.CORE_EXCEPTION)
+            name = None;
```
flit: Semicolon 😉

pyocd/gdbserver/gdbserver.py
```diff
@@ -1309,6 +1309,13 @@ def is_threading_enabled(self):
     def is_target_in_reset(self):
         return self.target.get_state() == Target.TARGET_RESET
 
+    def exception_name(self):
+        try:
+            ipsr = self.target_context.read_core_register('xpsr') & 0x1ff
```
flit: I'd like to get #436 in first, then make use of `get_ipsr()` here.

flit:
Ok, now that #436 is merged in, if you update `GDBServer.exception_name()`, I'll get this one in, too. Thanks!

kjbracey-arm:
Rebased and adjusted. Are you happy with the `Interrupt[%s]` and `Interrupt %d` formatting?

Not 100% sure whether I like those, but not sure what would be better.

kjbracey-arm:
> The only thing I can think of that might improve the descriptions is to always include the irq number, something like `Interrupt 17` and `Interrupt 29 (UART1)`. The irq number is important if you are correlating with NVIC registers.

I think this (just about) comes under my "clutter" category, but it is marginal. The actual interrupt number is still available if you happen to need it, by examining IPSR. Don't need to show all metadata in the thread list.

Guess it could be `Interrupt 29 [UART1]` though. 

I nicked the current style from the RTX `Waiting[xxx]`, and Eclipse also uses `[]` for thread names - although GDB uses quotes. Want to avoid round brackets as description is shown in another set of round brackets.

PR #450, opened 2018-11-13
RTOS fixes

kjbracey-arm:
All the fairly-unarguable fixes from #430

* RTOS: Correct stack offset calcs 
* RTX5: Improve "current thread" logic
* RTOS: adjust `inException` flag
* RTOS: Simplify `XXXThread::get_stack_pointer()`
* RTOS: improve "exception unavailable" registers
* RTOS: Tidy SP read out and don't mess with PSP
* RTX5: Tighten Thread LR read - match SP code
* Add `CortexM::is_vector_catch`
* RTX5/FreeRTOS: Go for live LR on vector catch
* GDB: Only do SIGSEGV etc on vector catch


kjbracey-arm:
One change since last version on #430 - extending "live LR" logic to FreeRTOS, to which it seems to apply by inspection.

PR #451, opened 2018-11-13
Correct naming error in exception-showing

kjbracey-arm:
Late edit to #437 missed a rename of `num` to `exc_num` - code was
showing external interrupts successfully, but trapped if faced with an
internal exception.


kjbracey-arm:
@SeppoTakalo next to me just hit this bug using 0.13.1 - could you get this into 0.13.2?

PR #453, opened 2018-11-16
RTOS: Add "Bare ARM-M" provider

kjbracey-arm:
If no real RTOS is present, and threading is enabled, use a
bare-metal provider that gives us two Main and Process threads.


kjbracey-arm:
First draft - no structural changes. Modifies visible name of `HandlerModeThread` to be `Main` just for this provider, as the thread isn't necessarily in Handler mode, unlike for real RTOSes.

kjbracey-arm:
Actually, there's no option to enable thread support is there, so this will always activate...

flit:
No, there isn't a thread support option yet, but it's on the list to add at some point.

pyocd/rtos/common.py
```diff
@@ -102,5 +103,154 @@ def __str__(self):
...
+    def priority(self):
+        return 0
+
+    @property
+    def unique_id(self):
+        return self.UNIQUE_ID
+
+    @property
+    def name(self):
+        log.debug("name name")
```
flit: Extraneous log, which barfs anyways since `log` isn't defined.

pyocd/rtos/__init__.py
```diff
@@ -15,14 +15,17 @@
...
 from .zephyr import ZephyrThreadProvider
 from .rtx5 import RTX5ThreadProvider
 
-RTOS = {
-          'Argon' : ArgonThreadProvider,
-          'FreeRTOS' : FreeRTOSThreadProvider,
-          'Zephyr' : ZephyrThreadProvider,
-          'RTX5' : RTX5ThreadProvider,
-         }
+RTOS = OrderedDict([
```
flit: I think I'd rather have special logic for the Bare ARM-M provider so that it gets used as a fallback instead of requiring the RTOS list to be sorted. In other words, Bare ARM-M would not be in the `RTOS` dict.

flit:
I tried running an Argon-based app with #430 + this PR. It unexpectedly failed to detect Argon, and used the Bare ARM-M provider. However, it only ever showed the Main thread. The Process thread never appeared, even when in an exception handler. (I haven't looked into the cause for any of these issues, sorry.)

pyocd/rtos/__init__.py
flit: It would also be nice to use Bare ARM-M even when another provider is selected but not yet active, for instance in `main()` before the kernel has started running. While this isn't really applicable to Mbed OS, it would be useful for the other RTOSes and RTX5 outside of Mbed.

kjbracey-arm: Agree - this is certainly one of the "edge cases" I was thinking about.

I think this is doable - my current thought is that the GDBServer will have two thread providers in use - 1 core one (always present) and 1 RTOS one.  The RTOS one gets the opportunity to override the core one if it decides the RTOS is active.

Will start working in this direction.

kjbracey-arm:
> The Process thread never appeared, even when in an exception handler. 

Well, if Argon is active, then Bare ARM-M isn't, so you wouldn't ever see the Process thread, no, the Argon one would kick in. This current implementation isn't a "fall-back when RTOS says it's not currently active" it's "fall-back when we didn't init a real RTOS provider at all".

I will do better.

flit:
This was with Argon accidentally disabled (I was running a release build that doesn't have debug symbols, though I need to double check that), so Bare ARM-M _was_ active.

Btw, would you mind changing the name to one of "Bare Cortex-M", "Bare metal Cortex-M", or "Bare metal" (whatever you prefer)? "ARM-M" is not a common term.

kjbracey-arm:
The code is designed so that the Process thread first appears whenever it sees execution stopped on the Process stack (so in a thread, if running Argon). The Process thread then remains visible until reboot.

I guess you only ever stopped on exception handlers, and never stopped at main or anything that would have made Process appear.

What it doesn't manage is to figure out that the exception came from process stack while unwinding the exception handler, and then add the Process stack at that point. I think it's too late.

I was assuming you may not want to have the Process thread visible all the time, as it may be a non-RTOSsy image that's using only the Main stack. I wanted to detect "Process stack/PSP is in use". But that's tricky.

Originally I thought CONTROL.SPSEL was going to do it - that it always held "stack for Thread mode", so would always tell me if Process stack was in use, but no. And PSP is apparently UNKNOWN at boot, so a zero check isn't architecturally sound (but may work in practice?)

Maybe this is trying too hard, and it could always be visible. No huge deal to show a non-current Process thread that's irrelevant.

flit:
Ok, that makes sense. You're right, I went from Main directly to an exception.

I wouldn't trust a zero check of PSP. I've seen plenty of bare metal code that doesn't init PSP.

c1728p9:
> If no real RTOS is present, and threading is enabled, use a bare-metal provider that gives us two Main and Process threads.

One concern I have with this approach is that is makes it appear that there are two unrelated threads. In actuality the interrupt contains and must preserve registers not already saved by hardware - R5 to R11 along with some FPU registers (if used). Modifications to these registers intentionally or unintentionally by the debugger could result in failures in the process execution context.

Furthermore, if pyOCD's RTOS awareness supports writing to thread registers (in addition to the read support it has now) the thread that the interrupt occurred on would need to be treated specially. For this thread writes to some registers would not be allowed/possible since these are held in the ISR registers and not necessarily in memory. The code would have to handle this special case by either silently failing or by showing some sort of error message to the user. Various cases, with this special case in bold, below:
1. Thread is ready to run or suspended - Safe to write to thread registers
2. Thread is running, no ISR active - Safe to write to thread registers
3. **Thread is running, ISR active** - Writes to some FPU registers and R5 to R11 not allowed

This problem would be further complicated with vector catch faults showing the current thread (#430) as it would appear that you are in an isolated context which you can manipulate, when really the execution is at a different location and (some) register modifications to the thread registers are not allowed.


Some assumptions I have when debugging with GDB in general (pyOCD doesn't not fully support this behavior yet) are below:
- When halting a multithreaded process with GDB each thread GDB shows is equally safe to read/write registers on, set breakpoints on (even if shared), or otherwise manipulate.
- When manipulating a callstack with GDB it is always reasonable/safe to modify the top frame of the callstack. It is typically not reasonable/possible to modify lower frames of the callstack.
- Each thread reported by GDB can execute in a stand-alone manner. If all threads on the system are halted any given thread can be selected and single stepped (assuming the given thread is not blocked in an OS call dependent on another thread).

The change proposed in this PR would violate these assumptions. @kjbracey-arm I know you have some [thoughts against GDB performing the stack unwinding itself](https://github.com/mbedmicro/pyOCD/pull/430#issue-224278378), but this may be a good alternative. Some advantages of GDB unwinding internally when an ISR occurs while a thread is executing:
1. No special case handling for the current thread with an ISR active as mentioned above - Its clear you can't modify the current thread's registers because the current thread is further down on the callstack (ISR portion of the callstack is on top of the thread portion). It is always safe to modify the register of the top frame of a callstack.
2. Existing debugging commands can such as step and return be used to return from ISRs. Even without GDB integration you can return from an ISR back to the current thread by instruction stepping to the return instruction. Having GDB support for unwinding would both make the stackframe you are returning to visible, and allow the `return` command to be used directly to return from each frame. Returning back to thread mode would be no different than returning from a nested ISR or even just returning from another stackframe. In cases where the ISR isn't fatal, such as an SVC, this would be particularly useful as both the MSP and PSP frames would be relevant to what is being debugged (RTX uses SVC ISR almost literally as a function call, taking parameters and returning a value).
3. What you see is what you get - If you have a hardfault and run/single step you don't end up off somewhere else. The debugger shows you where you really are so breakpoints get hit where expected and single stepping works as expected.

Let me know what your thoughts are @flit @kjbracey-arm

kjbracey-arm:
> Modifications to these registers intentionally or unintentionally by the debugger could result in failures in the process execution context.

That's not quite an exception-specific problem. Modifying R4-R11 could cause a failure a long way up the current stack frame.

Also, there are lots of ways to mess with the system if you start modifying registers manually. The debugger itself should never mess with any register that wasn't already saved by hardware or software, assuming it's following the debug info.

I'm not sure it necessarily follows that "protection" rules like "if you mess with R4-R11 in the IRQ handler, you only blow up the IRQ handler, not the thread it interrupted" are particularly meaningful or useful. If you blow up the IRQ handler, you're not going back to the thread anyway...

> Furthermore, if pyOCD's RTOS awareness supports writing to thread registers (in addition to the read support it has now) the thread that the interrupt occurred on would need to be treated specially. 

It currently doesn't at all, but certainly something to make sure we don't preclude.

>     Thread is running, ISR active - Writes to some FPU registers and R5 to R11 not allowed

Those writes can actually be allowed at the very instant of vector catch, but beyond that are not possible, as soon as the handler has potentially used them. GDB does understand the concept of "unavailable" registers, so we could arrange better presentation of that case.  (Previously pyOCD would read as 0 but still let you write - current version reads live as it works at vector catch, and anything's better than 0. Could read live at vector catch instant and unavailable otherwise).

> This problem would be further complicated with vector catch faults showing the current thread (#430) as it would appear that you are in an isolated context which you can manipulate, when really the execution is at a different location and (some) register modifications to the thread registers are not allowed.

For now, with no thread-specific writing, 430 doesn't really change anything.

If thread-specific writing was added, it could just work after 430. You write to the handler thread, you're modifying the live registers for the handler. You write to anything else, you're modifying that thread. Non-hardware-stacked registers could certainly be protected after vector catch for the interrupted thread, as described above.

> Some assumptions I have when debugging with GDB in general (pyOCD doesn't not fully support this behavior yet) are below:
> 
>     When halting a multithreaded process with GDB each thread GDB shows is equally safe to read/write registers on, set breakpoints on (even if shared), or otherwise manipulate.

Would certainly be nice.

Note that in some ways the interrupted thread IS conceptually distinct from the ISR more than a normal callstack. It is possible to modify its PC and stack pointer, which you couldn't do if it was just higher up the callstack. It would in principle be possible to tell the interrupted thread to "step out" while you're still in the ISR handler thread (aside from the problem of finding the R4-R11 saved over in the ISR callstack).

>     When manipulating a callstack with GDB it is always reasonable/safe to modify the top frame of the callstack. It is typically not reasonable/possible to modify lower frames of the callstack.

I do expect to be able modify lower frames of the callstack. I'd expect this to work fine now. Only needs unwind info, and I don't think pyOCD is involved. 

>     Each thread reported by GDB can execute in a stand-alone manner. If all threads on the system are halted any given thread can be selected and single stepped (assuming the given thread is not blocked in an OS call dependent on another thread).

That would need a lot of work - certainly nowhere near possible now. Handlers being a separate thread does put in a wrinkle there, but handlers being a separate thread isn't new behaviour - it's how pyOCD has worked for a while. The wrinkle may just be to make it so that the handler thread cannot be halted (and obviously it runs as higher priority than all other threads - so can't step another thread until it's completed).

> The change proposed in this PR would violate these assumptions. @kjbracey-arm I know you have some thoughts against GDB performing the stack unwinding itself, but this may be a good alternative. Some advantages of GDB unwinding internally when an ISR occurs while a thread is executing:
> 
>     No special case handling for the current thread with an ISR active as mentioned above - Its clear you can't modify the current thread's registers because the current thread is further down on the callstack (ISR portion of the callstack is on top of the thread portion). It is always safe to modify the register of the top frame of a callstack.

Actually, I don't see any reason you couldn't modify the current thread's registers in that model. GDB would find them in the unwind info (including the hardware stacked ones it was aware of).

>     Existing debugging commands can such as step and return be used to return from ISRs.

That would be sweet, certainly.

>     What you see is what you get - If you have a hardfault and run/single step you don't end up off somewhere else. 

Example? After 430, it behaves as standard GDB. Stops at the SIGSEGV point in the thread, and when you run/step you go into the signal handler.

> Let me know what your thoughts are @flit @kjbracey-arm

I agree that GDB doing the exception unwind has a number of advantages. It certainly does make it "perfect", covering a number of gaps in this approach. But they're all advantages for the 1% of cases. People who want to manipulate threads while inside IRQ handlers, etc.  My primary concern is the 99% of people/cases who just want to see where they wrote to NULL. So I'm quite prepared to do the "simple" thing in pyOCD rather than open the can of worms of "GDB across the app/OS boundary".

CPU exception unwind is something that GDB cannot do on any processor or any OS, as far as I'm aware. You're either debugging an application, in which case you only see the application threads, or you're debugging the OS, so no unwinding from one to the other.

Making IRQ handlers et al appear as on the same callframe as the thread they're interrupting is something that can be done on ARM-M, but it is very much a special case - I'm not aware of any other CPU you could do it on without major toolchain extensions. It's possible because of the "hardware does all the storing for you, so just write functions normally" feature which means we don't need any unwind extensions to find PSR etc.

Now, none of these pyOCD changes do preclude the unwind work being done to GDB as well.  If GDB can in future unwind an ARM-M exception frame, then you get a "deluxe" view on the Handler thread, going back into the thread callstack. And if pyOCD knew GDB could do that, pyOCD could choose to collapse the separate "handler" and "live process" threads down to that single callstack.

But such behaviour would be totally ARM-M specific.  pyOCD is in a slightly odd position here in that it is actually ARM-M specific so maybe could rely on as-yet-unimplemented ARM-M-specific GDB behaviour to save it having to do anything about multiple hardware stacks + modes.

But if pyOCD were to ever do ARM-A, say, then there's no way it could rely on GDB unwinding from SVC or IRQ stack to USR stack. Its only option would be this sort of thing - separate SVC, IRQ and USR "threads" representing the stacks. You would still want a "Bare ARM-A" handler even if "Bare ARM-M" was redundant.

Anyway, in summary - at present, if you have an RTOS awareness enabled, you can see where your hard faults came from, because you have separate RTOS and Handler threads. This sort of change "just" extends that out to provide the same facility with no RTOS awareness. It's a "null RTOS", basically. Not a new approach, just making sure this approach is always accessible.

Maybe on ARM-M, neither RTOS nor non-RTOS should be presenting a separate Handler thread, but should be relying on a GDB ARM-M cross-exception unwind, but when I looked at doing that in GDB, I struggled to put together how to integrate something so CPU-specific when there was no real infrastructure. I hit the same sort of block as the previous people looking at it. (Just conceptualising "PSP" and "MSP" as opposed to "SP" is a problem - GDB unwinds normally only manipulate "core" ABI registers).

This model is using the tools and language we have (threads supplied by the GDB server) to do as well as we can for hardware exceptions in a generic fashion that would work in principle on any bare-metal CPU, covering the 99% of cases that don't need cross the app/OS boundary.

I certainly wouldn't be adverse to someone tackling GDB extensions to cross the app/OS boundary - (ARM-M limited until DWARF is extended to exception handlers and PSR). It just seems very, very niche. People have coped with not being able to see where their SIGSEGV are coming from for about 5 years now, which I think is quite a mainstream demand, so I don't know how much demand there will be for perfection on unwinds from depths of SVC or IRQ handlers.

kjbracey-arm:
Couple more thoughts - separating the OS threads from the handlers does have its own advantages too.

1) Simplicity - ability to totally ignore the OS, as if just debugging an application. You can see the true callstack of every real thread even if you stop in a handler. The handler isn't glued onto a real thread.
2) You can achieve the standard "SIGSEGV at file:line" (my number 1 priority), by pointing at the relevant process thread. If the process thread always has the handler stuck on, all you can get is "SIGSEGV at HardFault_Handler", unless you actually lie about the PC for the process+handler combined thread.
3) If stepping through a context switch, it's less weird. All OS threads remain stable during the thread switch - flow goes smoothly from process thread to handler thread to other process thread, rather than the current thread shifting under your feet mid function.

I think I'd be inclined to go for the best of both worlds - GDB server continues to separate the "handler" from the OS threads, but GDB is also capable of unwinding from the handler into the interrupted thread. Thus you would see the callstack of the current process thread in two places - in its OS thread and as the base of the handler thread.

And 430 still makes sense - you set the config option for which of the 2 possible threads to stop on for vector catch depending on whether you're more interested in debugging the fault handler or the process.

kjbracey-arm:
Final summary:

1) pyOCD already has the facility to separate Handler from Process threads - it always does this if RTOS awareness is on - and this has a number of advantages, and is absolutely vital for fault debugging in the absence of GDB exception unwinding.
2) This brings non-RTOS behaviour into line with RTOS behaviour, and can be justified on that consistency ground. 
3) If the non-RTOS presentation is deemed correct and the separation is unwanted, then RTOS should be brought into line - remove the special "handler" processing and always show live registers for current thread like non-RTOS, even if in an ISR or fault, and wait for a potential GDB update. Until then you'd be back to not being able to see where your hard faults occurred unless you can manually dissect a stack, and even then you'd never be able to get the user-friendly "SIGSEGV at file:line" output. Massive downgrade.

If/when GDB ever does an ARM-M exception unwind, we can revisit the handler behaviour of pyOCD for both RTOS and non-RTOS, but as per my previous message I'd be inclined to not change it anyway, and retain the separation for the handler, not just show live registers in place of current OS thread.

c1728p9:
Thanks for the detailed response @kjbracey-arm. Since GDB support for unwinding through ISRs and pyOCD writing to the non-active thread both aren't currently supported, there is no reason to limit ourselves for features that may not even happen. When/if these features are added we could always re-evaluate and switch the way threads are displayed if it turns out there are clear advantages (which there may not be).

You have my support for moving this forward.

kjbracey-arm:
Just dropping this here in case you want to have a look - will be back to work on it after the weekend. This is my work in progress on a composing `ThreadProvider` design, as per review comment.

No limit to composition in principle, but at the minute we just have the CortexMThreadProvider giving us the hardware's base stack pair, then the RTOS providers replace its PSP thread with the RTOS set.

Minimal testing done with RTX5 - basically operational. Not rechecked other providers are consistent with it.

Quite happy because it generalises the model beyond Cortex-M, removing some of the existing Cortex-M knowledge in the general code, plus more I added in #430. Had to generalise anyway to remove include dependency loops to let the core have a thread provider.

Think I should be able to further simplify the RTOS providers by making their current thread pass the "read register" call on to the parent's PSPProcessThread. (They would note it for this purpose, but not pass it on to the GDBServer).

kjbracey-arm:
Reworked further - eliminated the "no threading" support from GDB server - that's now handled by a "RootThreadProvider". We then compose "CortexMThreadProvider" with the PSP/MSP split on top, and then compose a single RTOS provider on top of that.

The CortexMThreadProvider and RTOS provider can be omitted to get "bare metal" debug. RTOS provider without CortexM will work, but display might be a bit weird.

Still WIP - full of debug statements and fiddling, and probably only RTX5 functional.

kjbracey-arm:
Tidied up - it's now might be in a fit state to be tested and merged. This incorporates #430, but ends up reworking it a lot, including renaming the option.

kjbracey-arm:
Finally managed to eliminate the exception and current thread handling from all the RTOS providers. It's basically as simple as making `FreeRTOSThread`'s `context`  return either a `PSPThreadContext` or a `FreeRTOSThreadContext`, depending on whether the thread is current or not. Then `FreeRTOSThreadContext` only has to deal with accessing threads parked in a task control block.

kjbracey-arm:
At this point what's left of `read_core_registers_raw` for each `XXXThreadContext` is basically the same. All RTOSes could probably all share the same implementation, just providing different register layout tables.

kjbracey-arm:
Rebased and retested.

flit:
Thanks for rebasing!

And yes, I have a todo item to refactor the RTOS support to make as much as possible common.

flit:
@kjbracey-arm `gdb_test.py` failed with the following exception (repeatedly). Oddly, only the K64F failed, but it failed on all OSes.

```
0534094:ERROR:gdbserver:Unhandled exception in handle_message: 2
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/gdbserver/gdbserver.py", line 559, in handle_message
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/gdbserver/gdbserver.py", line 775, in v_command
    return self.v_cont(cmd)
  File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/gdbserver/gdbserver.py", line 811, in v_cont
    if thread_actions[currentThread] is None:
KeyError: 2
```

kjbracey-arm:
Just added a commit addressing #555 - also makes the register-in-memory lookup common. I'll now look at your issues above.

kjbracey-arm:
The exception is down to an inconsistency in the Cortex-M provider. It's giving only thread 3 (process stack) in its list of threads, but saying thread 2 (main stack) is current.

That's because its "emergency weird situation" logic has been activated, because it's seeing SP different from PSP in the register file. That case was intended to handle "what if some future core has an extra mode we don't know about", but it's producing an inconsistent result, so needs to be fixed.

The reason SP is different from PSP is that the register cache does not forget PSP/MSP if you write to SP. I'll fix that too.

kjbracey-arm:
I think I'm getting through this, and I'm running `gdb_test.py` with my K64F, but I'm seeing some unreliability - it seems to tend to get stuck while running, but not sure why. Maybe just not hitting the set breakpoint.

Seems to get worse with debug on, and I can't quite decide whether my patches make it worse or better. Seems to be there with master. Any tips?

kjbracey-arm:
I'm now fairly confident it's a race - I haven't yet figured out where the GDB test is actually defined, but this log shows:

> Executing command: set $pc = main
> Executing command: c&
> 
> Breakpoint 1, 0x1fff0028 in main ()
> Executing command: c&
> Executing command: interrupt

and in the pyocd logs I see

> DEBUG:gdbserver:Tresponse=b'T0507:f007ff1f;0d:f007ff1f;0e:f30b0000;0f:2a00ff1f;thread:3;'
> ERROR:gdbserver:Got unexpected ctrl-c, ignoring

It looks like the ctrl-C from `interrupt` is being processed out-of-band before it processes the `continue` command. More debug I add, the more likely this gets.

Is there really any need for ctrl-C to be processed out-of-band like that, rather than just a normal part of the packet queue?

kjbracey-arm:
Actually, it's more complex than that still. The test is doing 

        # Let target run to initialize variables
        gdb_execute("c&")
        event = yield(0.1)

When it's actually stopped on a breakpoint.  That seems to make GDB clear the breakpoint (z), step (vCont;s), re-add the breakpoint (Z) and continue (vCont;c).

But the Ctrl-C is coming through the protocol pipe in the middle of that series - it happens before we get the continue command.

I think we need to remember the interrupt event and handle it on next resume/step.

kjbracey-arm:
Okay, some bugs fixed, so it doesn't crash any more (in particular avoiding any possibility of inconsistency between "current thread" and the "thread list" for the Cortex-M provider), but it still reports a number of failures in gdb_test, apparently because it's just a bit too slow. I haven't yet figured out why it's slower, or how much by, but it seems that it will with a K64F now occasionally miss the 0.1s deadline, at least on my laptop, more than master seems to.

I think the "don't ignore Ctrl-C" may also have made that a little bit tighter, as if a "ctrl-c" and the "I've stopped" messages cross, the Ctrl-C will still be responded to rather than ignored.

kjbracey-arm:
Any profiling tips would be welcome. Last thing I want to do is make my debugger any slower, but I can't see anything obvious comparing traces. Cache hit readouts suggest basically the same amount of data being read over the DAP link on master versus this PR, yet it's slower :(

kjbracey-arm:
Ran `_run_connection` under the cProfile, and nothing terribly conclusive shown. It shows all the core transfer routines (read/write mem/regs) being issued less if anything, but then I guess the timeouts reduce the total amount of test work done.

The only significant cause of extra work I can see is `CortexMThreadProvider.get_current_stack_pointer_id` reading all of SP,MSP,PSP,CONTROL and IPSR to do its job (including sanity checks). Stripping that down to the bare read of CONTROL doesn't seem to help.

flit:
Sorry, I'm just catching up on everything over the past day…

pyocd/coresight/cortex_m.py
```diff
@@ -1304,3 +1328,272 @@ def exception_number_to_name(self, exc_num, name_thread=False):
...
+            return CortexM.PSP
+        else:
+            return CortexM.MSP
+
+#
+# Thread support
+#
+
+## @brief Class representing the handler mode.
+class MainStackThread(TargetThread):
```
flit: Could you move everything below here to a new module, maybe under `pyocd.debug`?

kjbracey-arm:
Rebased and split as suggested. Import of `CortexMThreadProvider` stuck at end of cortex_m.py, as I'm no good at resolving Python circular dependencies.

PR #458, opened 2018-11-24
Cortex-M7 support

flit:
This pull request adds support for the Cortex-M7 and FPBv2.

The two targets included with this PR are the NXP i.MX RT1050 and RT1020. Debugging images loaded into flash is known to not work, but debugging from RAM works fine. This issue will be addressed soon.


flit:
cc @MaureenHelm

flit:
Will do!

PR #472, opened 2018-12-13
Refine target description register types+groups

kjbracey-arm:
* Explicitly label FP registers as 'ieee_double' and 'ieee_float'.
* Mark msp and psp as 'data_ptr' type like sp.
* Assign msp, psp, primask, control, basepri and faultmask to 'system'
  group.

Not sure if FP type change has any affect, but is consistent with XML
in GDB.

Marking msp and psp as "data_ptr" means that "info register psp" will
show a symbol match like "<_main_stack+4080>", as it does with sp.

Making other registers system means they can be shown with "info
registers system". However, it doesn't currently remove them from "info
registers" or "info registers general" - these both appear to have
hard-coded behaviour of showing all non-vector non-float registers, not
specifically "general" ones. The intent is that only r0-12,sp,lr,pc,xpsr
would be shown in general, if GDB were to obey that label as some
documentation suggests.


kjbracey-arm:
I admit this is somewhat pointless fiddling. My main hope was just to reduce "clutter" from `info registers` by reducing the number of "general" registers, but doesn't work with current GDB, notwithstanding documentation.

msp/psp data_ptr change at least does something. Only GDB CLI pays attention though - Eclipse register view seems to be hard-coded for sp and pc only.

Another thought for possible future investigation - GDB documentation claims that if you declare the d0-d15 registers, it will auto-generate s0-s31 at its end. If true, might improve performance on those devices.

flit:
Thanks! Do you have a pointer to the gdb docs about the register XML? Last I looked there wasn't much there. I've also played with this a bit, but I was only looking at the Eclipse Registers view.

One of the experiments I did is in commit flit/pyOCD@945a7fcd68448bf6ba372ff72cf3e81a4e0c770d, where a flags type is created for control and xpsr so you can see the individual flags of those registers. This is visible in Eclipse, haven't tried command line. If you think this is a good change, it might make sense to include it in this PR. Actually, this one might be something that should be controllable via session option.

kjbracey-arm:
I suspect I haven't found any docs you haven't. Basically just the manual pages, and staring at the source a bit:

https://sourceware.org/gdb/onlinedocs/gdb/Target-Description-Format.html
https://sourceware.org/gdb/onlinedocs/gdb/ARM-Features.html
https://sourceware.org/gdb/wiki/Internals%20Adding-Target-Described-Register-Support

Looking at the source, I see there may be a bit of hardcoded magic for certain register numbers, which I can't decide would cause us grief - our registers in the 16-24 range overlap with historical FPA usage. I also wonder what happens with your negative double register numbers, which GDB ends up seeing, I think.

kjbracey-arm:
If there is register number magic, then we'd probably want a bit of care to only explicitly give register numbers to the core ones that GDB knows about (that predate target descriptions), and then I think other declared registers just get "virtual" auto-assigned numbers.

As for your experiment - looks good, I'll try it when I've got a chance.

kjbracey-arm:
Hmm, did you find that `flags` doesn't work for anything other than single-bit fields? I just seem to get nonsense results - if I say `GE` is bits 16-19, GDB seems to be looking at bits 13-16. But it's fine if the same bitfield is in a `struct`.

kjbracey-arm:
And Eclipse does actually just follow GDB's CLI `print` formatting for everything. When I wasn't seeing the `data_ptr` for MSP/PSP do anything it was just because I  I just had MSP and PSP set to "Hex" rather than "Default" for some reason.

kjbracey-arm:
Been staring at the double/single thing a bit. Looking at it, I'm thinking we really should be supporting the standard "org.gnu.gdb.arm.vfp" feature, which does require us to have d0-d15. And it's correct that we should do so. Even if we don't have double-precision arithmetic, the ARM can handle the double precision registers, and use them to load and store and pass double-precision numbers - they probably would often hold double precision numbers when using hardfp ABI.

kjbracey-arm:
Okay, I think I am going to follow up further - using `flags` or `struct` as per your test, and using "org.gnu.gdb.arm.vfp". But those will be separate PRs. They're more intricate than this.

PR #473, opened 2018-12-13
flash/loader: fix python3 compatibility issue with hex files

aabadie:
This PR is fixing a compatibility issue with Python3 when trying to load an hex file.

I had this issue using Python 3.6 while trying to load a micro:bit firmware in hex format.
Tested with Python 2 and 3 and now it works as before.

Since this is my first PR on the project, I apologize if I'm not correctly following any development guideline. If a non regression test is needed, I'll be happy to write one but would need a bit of help.

Related to https://github.com/RIOT-OS/RIOT/pull/9847

### DIFF
--- a/pyocd/flash/loader.py
+++ b/pyocd/flash/loader.py
@@ -122,7 +122,11 @@ class FileProgrammer(object):
         try:
             # Open the file if a path was provided.
             if isPath:
-                file_obj = open(file_or_path, "rb")
+                mode = 'rb'
+                if format == 'hex':
+                    # hex file must be read as plain text file
+                    mode = 'r'
+                file_obj = open(file_or_path, mode)
             else:
                 file_obj = file_or_path
 


flit:
Hi @aabadie, thanks a lot for the fix! (And it's great to see pyOCD support added to RIoT.)

The only thing we need is for you to sign the [Mbed Contributor Agreement](https://os.mbed.com/contributor_agreement/). Just follow that link and create an account, then click through to sign the agreement.

Don't worry about adding a test. Technically it would be best, but there is a *lot* more testing needed on the `pyocd.flash.loader` classes, so I'll cover this case when I get to it.

pyocd/flash/loader.py
```diff
@@ -122,7 +122,11 @@ def program(self, file_or_path, format=None, **kwargs):
         try:
             # Open the file if a path was provided.
             if isPath:
-                file_obj = open(file_or_path, "rb")
+                mode = 'rb'
+                if format == 'hex':
```
aabadie: Using `format` is not a good idea in Python since it's already used by a built-in function. I would suggest (but not in this PR) to use a different name, e.g `file_type` or whatever.

aabadie:
> The only thing we need is for you to sign the Mbed Contributor Agreement

Done, my account name is also aabadie on mbed.

> I'll cover this case when I get to it

Ok, thanks!

pyocd/flash/loader.py
flit: Ah, good point! I was just reusing existing code with that variable name. Todo note taken.

flit:
Thanks for signing the CLA!

flit:
Just did! 😄 0.14.1 includes this fix. Thanks again!

PR #475, opened 2018-12-14
RTX5: Detect post-reset state

kjbracey-arm:
Just after reset, osRtxInfo will still show us running, but we can deduce we're not by the fact we're in Thread mode on the main stack.

### DIFF
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -204,6 +204,11 @@ class CortexM(Target, CoreSightComponent):
     EPSR_MASK = 0x0700FC00
     IPSR_MASK = 0x000001FF
 
+    # Control Register
+    CONTROL_FPCA = (1 << 2)
+    CONTROL_SPSEL = (1 << 1)
+    CONTROL_nPRIV = (1 << 0)
+
     # Debug Fault Status Register
     DFSR = 0xE000ED30
     DFSR_EXTERNAL = (1 << 4)
@@ -1105,3 +1110,7 @@ class CortexM(Target, CoreSightComponent):
                 return "Interrupt[%s]" % name
             else:
                 return "Interrupt %d" % irq_num
+
+    def in_thread_mode_on_main_stack(self):
+        return (self._target_context.read_core_register('ipsr') == 0 and
+                (self._target_context.read_core_register('control') & CortexM.CONTROL_SPSEL) == 0)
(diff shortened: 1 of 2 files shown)


kjbracey-arm:
This seems to make hitting the "restart" button nicer, if we're trapping on reset. It doesn't seem like there are any other mechanisms to stop the RTOS provider kicking in between the reset and any memory clear that wipes out its data structures.

If this makes sense, can extend to other providers - that's why I put the reusable subroutine in `CortexM`

pyocd/coresight/cortex_m.py
```diff
@@ -1105,3 +1105,10 @@ def exception_number_to_name(self, exc_num, name_thread=False):
                 return "Interrupt[%s]" % name
             else:
                 return "Interrupt %d" % irq_num
+
+    def in_thread_mode_on_main_stack(self):
+        try:
+            return self._target_context.read_core_register('ipsr') == 0 and (self._target_context.read_core_register('control') & 2) == 0
```
flit: Let's define a constant for `CONTROL.SPSEL`.

pyocd/coresight/cortex_m.py
```diff
@@ -1105,3 +1105,10 @@ def exception_number_to_name(self, exc_num, name_thread=False):
...
+        except exceptions.TransferError as exc:
+            logging.debug("Transfer error checking mode: %s", exc)
+            return False
```
flit: I'm not sure returning False on a transfer failure is the right thing to do. It's probably better to just let the exception pass further up. Returning False says (obviously) that the core is not in Thread mode on MSP, but we really don't know that.

flit:
There is actually some code to try to handle this situation in a generic manner. `GDBServer` maintains a `first_run_after_reset_or_flash` flag and uses it to set the thread provider's `read_from_target` flag. The presumption is that after a reset or flash, the RAM contents are likely to be out of sync with the RTOS state, and the thread provider shouldn't bother reading from the target until code has had a chance to run (resume). There are problems with this scheme, but it works for the most part.

What we really need is a way to know that C startup has executed and inited .bss and .data. After that, the RTOS state should be valid. So as an alternative to the existing scheme, we could try setting an internal breakpoint at `main()` and only allowing the thread provider to read from the target after it's hit.

kjbracey-arm:
I saw the first-run stuff, but couldn't quite get my head around it. So it wasn't about trying to avoid reads because they would trap or something, just about the memory being probably stale and not worth reading?

I think if the RTOS providers all can detect this "thread mode main stack" state, and we assume that zero-init happens while in that state - I think that's a safe assumption - then does this test totally supersede the current use of the first-run logic?

PR #478, opened 2018-12-18
Fixed serving localhost only for telnet

flit:
Regression introduced in 0.14.0. The semihosting telnet server would always open the port to remote access. This was fixed for the gdbserver prior to the 0.14.0 release, but telnet was missed.


flit:
CI servers are down because they are being moved, so tested locally.

arekzaluski:
👍

PR #481, opened 2018-12-20
Remove tests for specific cores

kjbracey-arm:
Don't use core type (PARTNO) to check for features - use architecturally-defined checks.

### DIFF
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -241,8 +241,8 @@ class CortexM(Target, CoreSightComponent):
     CPUID_REVISION_POS = 0
 
     CPUID_IMPLEMENTER_ARM = 0x41
-    ARMv6M = 0xC
-    ARMv7M = 0xF
+    ARMv6M = 0xC # also ARMv8-M without Main Extension
+    ARMv7M = 0xF # also ARMv8-M with Main Extension
 
     # Debug Core Register Selector Register
     DCRSR = 0xE000EDF4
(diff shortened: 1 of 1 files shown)


pyocd/coresight/cortex_m.py
```diff
@@ -524,13 +529,15 @@ def _check_for_fpu(self):
...
-            if self.core_type == ARM_CortexM7:
-                if self.has_fpu_double:
-                    fpu_type = "FPv5-DP"
-                else:
-                    fpu_type = "FPv5-SP"
+            mvfr2 = self.read32(CortexM.MVFR2)
+            vfp_misc_val = (mvfr2 & CortexM.MVFR2_VFP_MISC_MASK) >> CortexM.MVFR2_VFP_MISC_SHIFT
+
+            if self.has_fpu_double:
+                fpu_type = "FPv5"
```
flit: Can we change this to "FPv5-DP"?

kjbracey-arm: Could do, and it does seem more explicit, so I'd prefer it, but all the architectural references I can find call it just "FPv5", with the "-SP" as a possible restriction limiter.

The only reference to "FPv5-DP" I can find is in the Cortex-M7 TRM, but it seems to me that may be effectively an error.

(PyOCD actually doesn't need to care in the slightest - double arithmetic support or not has no impact on a debugger, as per #483. That removes `self.has_fpu_double`.)

flit: I must have been looking at the CM7 TRM because I do remember seeing "FPv5-DP" in Arm docs somewhere. :)

PR #482, opened 2018-12-20
Correct FPB revision checks

kjbracey-arm:
FPB revision was being stored in `fp_rev` but often tested from `fpb_rev`.

Combine these two variables.

Spotted while browsing code - by inspection, this appears to have made `FPB::can_support_address` pessimistic.


flit:
Good catch! That must've been a typo on my part…

PR #483, opened 2018-12-20
Present standard VFP view to GDB 

kjbracey-arm:
Previously we were adding s0-s31 to the org.gnu.gdb.arm.m-profile
feature, which bypassed GDB's built-in VFP support.

GDB's built-in VFP support uses the feature org.gnu.gdb.arm.vfp, and
that requires at least fpscr and d0-d15. From that, it will locally
generate s0-s31.

There is no reason not to use the standard GDB view - we can now
process double registers internally, so it is not onerous to have the
GDB context facade process them.

The existing code did declare both double and single registers to GDB,
if the CPU has double support, but this ends up transferring more data
than necessary.

If the chip does not have double support, we should still allow FP
registers to be inspected as doubles - such chips can still load, store
and move double values, and will use the registers for doubles if using
hardware FP calling conventions. So all restrictions on `has_fpu_double`
removed.

Context facade code appeared to not fully support doubles, despite the
fact we were already declaring them to GDB if has_fpu_double was set.
Support filled in.


kjbracey-arm:
Based on top of #481 and #482.

pyocd/coresight/cortex_m.py
```diff
@@ -798,12 +769,10 @@ def read_core_registers_raw(self, reg_list):
                 raise ValueError("unknown reg: %d" % reg)
             elif is_fpu_register(reg) and (not self.has_fpu):
```
flit: Shouldn't `is_fpu_register()` be updated to include double float regs?

pyocd/utility/conversion.py
```diff
@@ -80,6 +80,19 @@ def float64_to_u64(data):
     d = struct.pack(">d", data)
     return struct.unpack(">Q", d)[0]
 
+def u64_to_hex16le(val):
```
flit: Could you add a unit test for the new functions in this file?

pyocd/coresight/cortex_m.py
kjbracey-arm: Hm, probably should have been that already. I think this would error correctly anyway as doubles recurse back in to do the single read, but will adjust.

kjbracey-arm:
Added unit tests and made `is_fpu_register` detect doubles.

Renamed `is_float_register` to `is_single_float_register` for clarity, leaving old one with deprecation warning.

PR #491, opened 2019-01-02
Pin major version numbers for packages with stable apis

### LINKED ISSUE #477: requirements: versioning being more strict
Follow up from https://github.com/mbedmicro/pyOCD/pull/476. We experienced how easy is to break the chain of modules that depend on pyOCD.

We should not fetch any major version (at least major, not certain how strict we want to be) without testing first. Good example in 476.

bridadan:
This implements #477. It does not pin version requirements for packages that have unstable apis though (v0.x).

Closes #477

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,18 +48,18 @@ setup(
     # Allow installation on 2.7.9+, and 3.4+ even though we officially only support 3.6+.
     python_requires=">=2.7.9, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires = [
-        'intelhex',
-        'six',
-        'enum34;python_version<"3.4"',
+        'intelhex>=2.0,<3.0',
+        'six>=1.0,<2.0',
+        'enum34>=1.0,<2.0;python_version<"3.4"',
         'future',
         'websocket-client',
         'intervaltree>=3.0.2,<4.0',
         'colorama',
         'pyelftools',
-        'pyusb>=1.0.0b2',
+        'pyusb>=1.0.0b2,<2.0',
         'pywinusb>=0.4.0;platform_system=="Windows"',
         'hidapi;platform_system=="Darwin"',
-        'pyyaml',
+        'pyyaml>=3.0,<4.0',
         ],
     classifiers=[
         "Development Status :: 4 - Beta",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -37,18 +37,18 @@
...
-        'intelhex',
-        'six',
-        'enum34;python_version<"3.4"',
+        'intelhex>=2.0,<3.0',
+        'six>=1.0,<2.0',
+        'enum34>=1.0,<2.0;python_version<"3.4"',
         'future',
         'websocket-client',
-        'intervaltree<3.0',
+        'intervaltree>=2.0,<3.0',
```
flit: We should be able to set intervaltree to >=3.0.1,<4.0. The problem that triggered all this was fixed, and the 3.x versions fix a number of warnings caused by use of a deprecated API.

flit:
Thanks for taking care of this! 🙇

setup.py
bridadan: I bumped it to 3.0.2 since it looks like there was an [encoding fix](https://github.com/chaimleib/intervaltree/releases/tag/3.0.2):

"On some systems, setup.py opened README.md with a non-unicode encoding. My fault for leaving the encoding flapping in the breeze. It's been fixed."

flit:
Ah, seems there was a breaking API change in intervaltree. Sorry I didn't realise that! Let me take care of updating the code to use the latest intervaltree, then we'll bring this PR in afterwards.

bridadan:
@flit I can also just pin v2.x of intervaltree, that's what I was planning on doing originally.

flit:
No worries, I'm about to merge the intervaltree update, then will merge this one.

flit:
@bridadan Would you mind rebasing? There will be a merge conflict with changes from #496, though GitHub isn't showing that for some reason. Thanks!

bridadan:
Done!

bridadan:
Heads up, when the intervaltree dependency broke, the Mbed OS folks pinned it in the OS: https://github.com/ARMmbed/mbed-os/blob/master/requirements.txt

We'll want to remove that line there.

flit:
Does the OS use intervaltree?

bridadan:
> Does the OS use intervaltree?

Not directly, they pinned it when trying to get their CI going again when intervaltree had the issue with its 3.0 release.

bridadan:
I can't see the test results from morph, do you have a valid link?

flit:
The test failure was an unrelated CI problem we're having with the mac OS node right now.

PR #503, opened 2019-01-10
Ignore HEX sections which don't belong to Flash regions

### LINKED ISSUE #499: Ignore invalid addresses when programming hex files
`FlashLoader` must skip any addresses in hex files that do not belong to a known memory region (as defined by the target's memory map).

Cypress-OpenOCD:
Display warning message instead of throwing an exception. 
Resolves #499

### DIFF
--- a/pyocd/flash/loader.py
+++ b/pyocd/flash/loader.py
@@ -401,7 +401,8 @@ class FlashLoader(object):
             # Look up flash region.
             region = self._map.get_region_for_address(address)
             if region is None:
-                raise ValueError("no memory region defined for address 0x%08x" % address)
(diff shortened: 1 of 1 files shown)


flit:
Thanks for the fix. I'd rather have this handled in `FileProgrammer`, so that `FlashLoader` still raises an exception for invalid addresses. The easiest way to handle this would be to add a **try-except** block around the call to `add_data()` on line 162 of `loader.py`. You amend you commit and force push in order to continue using this PR.

Cypress-OpenOCD:
I would like co close this PR and open another one (with suggested fixes) - this PR does not follow GitHub flow.

PR #518, opened 2019-01-23
Improved reset configuration

### LINKED ISSUE #479: Broken python API example documentation
https://github.com/mbedmicro/pyOCD/blob/057528698f69cbd4e333334b6acb31ef4e3e5c7e/docs/API_EXAMPLES.md#hello-world-example-code

For mbedBoard there is no "flash" attribute already.
```
with ConnectHelper.session_with_chosen_probe() as session:
    board = session.board
    target = board.target
    flash = board.flash
```

### LINKED ISSUE #500: Target-configurable reset type
The default reset type needs to be configurable by target support. This includes hardware versus software reset, as well as `SYSRESETREQ` versus `VECTRESET` for software reset. In addition, the reset type must be configurable per core.

flit:
This set of changes enhances the reset management for cores.

- `reset_stop_on_reset()` was renamed to `reset_and_halt()`.
- Added `Target.ResetType` enums.
- The `reset()` and `reset_and_halt()` APIs take a `ResetType` enum instead of bool|None.
- Added an emulated software reset capability, so that cores that don't support `VECTRESET` have something to fall back to that doesn't reset the entire system as does `SYSRESETREQ` (nominally, at least).
- `CortexM` has two new properties, `default_reset_type` and `default_software_reset_type`.
- Removed `set_target_state()` API that was only used to prepare the core for flash programming. The check to ensure the Thumb bit in XPSR is set was moved to `reset_and_halt()`.

Documentation has also been updated, both to match the reset changes as well as to fix several cases where it was out of date with recent changes (especially the per-region flash instances).

Closes #500 
Closes #479

### DIFF
--- a/docs/ADDING_NEW_TARGETS.md
+++ b/docs/ADDING_NEW_TARGETS.md
@@ -7,7 +7,8 @@ For background information, review the [architecture overview](ARCHITECTURE.md) 
 
 ### Steps to add a new target
 
-1. Create a new `CoreSightTarget` subclass and `Flash` subclass in a file under `pyocd/target/`. You can copy one of the existing target files like `pyocd/target/target_ncs36510.py` and rename the classes.
+1. Create a new `CoreSightTarget` subclass in a file under `pyocd/target/`. You can copy one of the
+    existing target files like `pyocd/target/target_ncs36510.py` and rename the class.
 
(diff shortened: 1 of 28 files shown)


flit:
cc @Cypress-OpenOCD

PR #519, opened 2019-01-24
Chip erase option for gdbserver monitor erase command

flit:
Passing `--chip` to the `erase` monitor command will perform a chip erase.


flit:
cc @Cypress-OpenOCD

Cypress-OpenOCD:
@flit Having "monitor erase --chip" definitely speeds up erase for the targets with large amount of sectors. Tested it locally - works well, thanks!

PR #522, opened 2019-01-29
Ignore HEX sections which don't belong to Flash regions

### LINKED ISSUE #499: Ignore invalid addresses when programming hex files

Cypress-OpenOCD:
This is a rework of PR #503 with suggestions from @flit :

> Thanks for the fix. I'd rather have this handled in FileProgrammer, so that FlashLoader still raises an exception for invalid addresses. The easiest way to handle this would be to add a try-except block around the call to add_data() on line 162 of loader.py. You can amend you commit and force push in order to continue using this PR.

Closes #499

### DIFF
--- a/pyocd/flash/loader.py
+++ b/pyocd/flash/loader.py
@@ -146,8 +146,10 @@ class FileProgrammer(object):
         
         file_obj.seek(kwargs.get('skip', 0), os.SEEK_SET)
         data = list(bytearray(file_obj.read()))
-        
-        self._loader.add_data(address, data)
+        try:
+            self._loader.add_data(address, data)
+        except ValueError as e:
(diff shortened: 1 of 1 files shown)


flit:
Sorry, one more request—invalid addresses should be ignored only for hex files.

Binary files (obviously) don't contain addresses, so you have to assume that all data is part of the application and is required. For ELF files, any metadata that's not part of the application code will be held in a section that doesn't have the `SHF_WRITE` flag set. Only in a hex file would it make sense to use invalid addresses to carry metadata.

As part of the change, could you please add a comment with the above thoughts next to the **try** statement?

Thanks!

PR #525, opened 2019-02-03
CMSIS DFP target support

flit:
This change set adds initial, experimental support for using [CMSIS Device Family Packs](http://arm-software.github.io/CMSIS_5/Pack/html/createPack_DFP.html) to enable support for new targets.

To use this feature, you need to point pyOCD at one or more DFPs. This can be done with the `--pack` command line argument or by setting the `pack` session option in a config file (i.e., `pyocd.yaml`). pyOCD will examine the pack(s) and make the targets defined in the pack(s) available for selection with the `--target` command line argument or `target_override` session option. It creates targets for each `<device>` and `<variant>` element in the pack's PDSC file.

The targets defined by a pack can be listed with `pyocd list --targets --pack=<path>`. Note that due to how config files are currently loaded, packs listed in a config file won't be processed for `pyocd list`.

Notable changes:
- Builtin targets were moved to `pyocd.target.builtin`. The table of builtin targets is now `pyocd.target.builtin.BUILTIN_TARGETS`. The `pyocd.target.TARGET` table still exists and is initialized with a copy of `BUILTIN_TARGETS`, but will be extended at runtime with targets created from CMSIS DFPs.
- Vendor and family names are available for targets now, including builtin targets. The vendor name is listed when `pyocd list --targets` is run. Both vendor and families are included in the JSON targets output.
- Some memory map improvements.
- There is now a `pyocd.target.family.FAMILIES` table that defines regular expressions to map from family names to a family target superclass. This is used by the DFP support when creating target classes.
- The DFP support is in the `pyocd.target.pack` subpackage. A DFP is represented by a `CmsisPack` object. Each device or variant defined within a DFP is represented by `CmsisPackDevice` objects, which is responsible for converting DFP PDSC elements into pyOCD-compatible form.

The next step is to enable using cmsis-pack-manager to automatically manage packs for you, so you don't have to manually download them.


bridadan:
Awesome!!

PR #528, opened 2019-02-04
Replace encoded ellipses character with three periods

bridadan:
Certain terminals have issue with installing pyocd because the
long_description field in setup.py reads and parses README.md. This file
contains specially encoded ellipses characters which can cause
UnicodeDecodeErrors.

This issue was raised here https://github.com/ARMmbed/mbed-cli/issues/827#issuecomment-460123586. The `position 6073` shown in the error got me curious if there was a specific offending character, and these appear to be it!


flit:
The issue reported in the mbed-cli PR comment should be fixed already. pyOCD now sets the encoding to UTF-8 when reading the readme. The problem appeared on systems that had a default encoding of something other than ASCII or UTF-8.

bridadan:
Ah good to know

PR #534, opened 2019-02-12
Add missing % to format strings

mbolivar:
A few strings around the code base are not being properly formatted with %. Fix them.

Please note: I've accepted the contributor agreement under username https://os.mbed.com/users/martifoundries/.

### DIFF
--- a/pyocd/tools/pyocd.py
+++ b/pyocd/tools/pyocd.py
@@ -896,10 +896,10 @@ class PyOCDCommander(object):
             # Look up flash region.
             region = self.session.target.memory_map.get_region_for_address(addr)
             if not region:
-                print("address 0x%08x is not within a memory region", pagaddre_addr)
+                print("address 0x%08x is not within a memory region" % addr)
                 return 1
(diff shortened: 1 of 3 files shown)


pyocd/tools/pyocd.py
```diff
@@ -896,10 +896,10 @@ def do_write(self, args, width):
             # Look up flash region.
             region = self.session.target.memory_map.get_region_for_address(addr)
             if not region:
-                print("address 0x%08x is not within a memory region", pagaddre_addr)
+                print("address 0x%08x is not within a memory region" % pagaddre_addr)
```
flit: Since you're already touching on this line, would you mind fixing the bad `pagaddre_addr` variable name (I just noticed it)?

flit:
Thanks for the fixes!

pyocd/tools/pyocd.py
mbolivar: Done!

flit: Thanks!

PR #538, opened 2019-02-15
Added catch for unknown directory when parsing ELF symbols

geky:
Hit this issue when parsing elf file containing Pelion example + Mbed OS. (https://os.mbed.com/teams/ST/code/pelion-example-disco-iot01/). Note that this binary includes the bootloader, which doesn't have symbol information, maybe that is the source of the problem?

Not sure if this is a good fix, but didn't want to lose this bug. This solved the issue for us.

cc @flit, @korjaa

### DIFF
--- a/pyocd/debug/elf/decoder.py
+++ b/pyocd/debug/elf/decoder.py
@@ -182,9 +182,16 @@ class DwarfAddressDecoder(object):
 
                 # Looking for a range of addresses in two consecutive states.
                 if prevstate and not skipThisSequence:
-                    fileinfo = lineprog['file_entry'][prevstate.file - 1]
-                    filename = fileinfo.name
-                    dirname = lineprog['include_directory'][fileinfo.dir_index - 1]
+                    try:
+                        fileinfo = lineprog['file_entry'][prevstate.file - 1]
+                        filename = fileinfo.name
+                        try:
+                            dirname = lineprog['include_directory'][fileinfo.dir_index - 1]
+                        except IndexError:
+                            dirname = ""
+                    except IndexError:
+                        filename = ""
+                        dirname = ""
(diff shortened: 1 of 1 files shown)


pyocd/debug/elf/decoder.py
```diff
@@ -184,7 +184,10 @@ def _build_line_search_tree(self):
                 if prevstate and not skipThisSequence:
                     fileinfo = lineprog['file_entry'][prevstate.file - 1]
```
flit: Let's wrap the code to get the filename in a similar exception handler. I guess we haven't seen it fail yet, but I can imagine it might in some cases.

pyocd/debug/elf/decoder.py
```diff
@@ -184,7 +184,10 @@ def _build_line_search_tree(self):
...
                     filename = fileinfo.name
-                    dirname = lineprog['include_directory'][fileinfo.dir_index - 1]
+                    try:
+                        dirname = lineprog['include_directory'][fileinfo.dir_index - 1]
+                    except IndexError:
+                        dirname = '<unknown dir>'
```
flit: Can you change the unknown dirname to just "<unknown>"? We already know it's a dir because of the attribute to which it is attached.

pyocd/debug/elf/decoder.py
geky: @flit, how does the update look? It looks like fileinfo is only used for the filename/dirname.

pyocd/debug/elf/decoder.py
flit: I actually meant "\<unknown\>", but Markdown apparently thought that was an HTML tag. Oops. Anyway, I'm perfectly fine with empty strings!

PR #583, opened 2019-03-20
Update pyyaml requirement to >=5.1.

### LINKED ISSUE #582: PyYAML `>=4.2b,<5.0` requirement does not seem to be workable
Hi, as part of https://github.com/zephyrproject-rtos/zephyr/issues/14460 I learned the pyOCD setup.py has a strange YAML version range: `'pyyaml>=4.2b1,<5.0'`:

https://github.com/mbedmicro/pyOCD/blob/master/setup.py#L62

This is not a good idea; could you please update to 5.0 or continue working with 3.13? Either one would resolve our issue and I think is a good idea for pyOCD as well. From one of new lead maintainers of PyYAML:

```
There seems to be some suggesting that people use one of the
failed 4.2bx releases to get #74 behavior. This is a bad idea. 3.13
is the current supported release. I could delete the 4.2b-s from PyPI
but I haven't. I almost certainly will after 5.1 goes out.
```

https://github.com/yaml/pyyaml/issues/193#issuecomment-457387449

In other words, the current requirements are explicitly requesting users to install a PyYAML version that the maintainers are calling a "failed" release, and which they are actively discouraging users to install.

FWIW, the changes in the Zephyr tree to support PyYAML v5.0+ were minor and are discussed here: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

Can anyone from pyOCD comment on what's holding the code back from working with 5.0?

flit:
Closes #582

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ setup(
         'pyusb>=1.0.0b2,<2.0',
         'pywinusb>=0.4.0;platform_system=="Windows"',
         'hidapi;platform_system=="Darwin"',
-        'pyyaml>=4.2b1,<5.0',
+        'pyyaml>=5.1,<6.0',
         ],
     classifiers=[
         "Development Status :: 4 - Beta",


setup.py
```diff
@@ -59,7 +59,7 @@
         'pyusb>=1.0.0b2,<2.0',
         'pywinusb>=0.4.0;platform_system=="Windows"',
         'hidapi;platform_system=="Darwin"',
-        'pyyaml>=4.2b1,<5.0',
+        'pyyaml>=5.1,<6.0',
```
bridadan: Since this didn't require any code changes, you could leave the min version at 4.x, right? So its more compatible with other environments?

mbolivar:
Thanks @flit !

setup.py
flit: D'oh, sorry I didn't see your review. No, the 4.x versions are "bad" according to the pyyaml authors. Seems they even plan to remove them from PyPI. See the info in #582.

flit:
Thanks for letting us know about pyyaml 5.x!

setup.py
bridadan: Oh good point, I hadn't quite got that far in the post! Thanks for the response.

PR #589, opened 2019-03-22
Update to udev rules and related docs

flit:
- Added STLink rules files from STSW-LINK007. Thanks for STMicro!
- Renamed `50-pyocd.rules` to `50-cmsis-dap.rules` and removed STLink rules from this file.
- Added a readme to the `udev/` directory with detailed instructions.
- Simplified udev instructions in the top level readme and pointed to the udev readme.


jeromecoutant:
@svastm

PR #607, opened 2019-04-08
Add decoding of serial number from string to hexadecimal for ST-Link V2 probes

anthrax-0:
I could not find other solution for the ST-Link V2 probes serial numbers. Seems it's quite an old problem, for example, [here](http://wunderkis.de/stlink-serialno/).
 If there is some other planned fix for this, please, discard the PR.

### DIFF
--- a/pyocd/probe/stlink/usb.py
+++ b/pyocd/probe/stlink/usb.py
@@ -26,7 +26,6 @@ from collections import namedtuple
 import platform
 import errno
 from binascii import hexlify
-from sys import hexversion
 
 # Set to True to enable debug logs of USB data transfers.
 LOG_USB_DATA = False
@@ -128,11 +127,11 @@ class STLinkUSBInterface(object):
         # reading the strings, everything is ok. This workaround doesn't cause any issues with
         # Linux or macOS.
         try:
-            if self._dev.idProduct == 0x3748:  # Workaround for unprintable characters in the ST-Link V2 probes
-                if hexversion >= 0x03020000:  # To align with libusb encoder
-                    self._serial_number = bytes(self._dev.serial_number, 'utf-16-le')[::2].hex().upper()
-                else:
-                    self._serial_number = hexlify(self._dev.serial_number.encode('utf-16-le')[::2]).upper()
+            if len(self._dev.serial_number) == 12:  # Workaround for unprintable characters in the ST-Link V2 probes
+                self._serial_number = hexlify(self._dev.serial_number.encode('utf-16-le')[::2])\
+                    .decode('utf-8', 'replace').upper()
+            else:
+                self._serial_number = self._dev.serial_number
             self._vendor_name = self._dev.manufacturer
             self._product_name = self._dev.product
         finally:
--- a/pyocd/probe/stlink/usb.py
+++ b/pyocd/probe/stlink/usb.py
@@ -25,6 +25,8 @@ import threading
 from collections import namedtuple
 import platform
 import errno
+from binascii import hexlify
+from sys import hexversion
 
 # Set to True to enable debug logs of USB data transfers.
 LOG_USB_DATA = False
@@ -126,8 +128,12 @@ class STLinkUSBInterface(object):
         # reading the strings, everything is ok. This workaround doesn't cause any issues with
         # Linux or macOS.
         try:
-            self._serial_number = self._dev.serial_number
-            self._vendor_name =  self._dev.manufacturer
+            if self._dev.idProduct == 0x3748:  # Workaround for unprintable characters in the ST-Link V2 probes
+                if hexversion >= 0x03020000:  # To align with libusb encoder
+                    self._serial_number = bytes(self._dev.serial_number, 'utf-16-le')[::2].hex().upper()
+                else:
+                    self._serial_number = hexlify(self._dev.serial_number.encode('utf-16-le')[::2]).upper()
+            self._vendor_name = self._dev.manufacturer
             self._product_name = self._dev.product
         finally:
             usb.util.dispose_resources(self._dev)


pyocd/probe/stlink/usb.py
```diff
@@ -126,8 +126,11 @@ def __init__(self, dev):
         # reading the strings, everything is ok. This workaround doesn't cause any issues with
         # Linux or macOS.
         try:
-            self._serial_number = self._dev.serial_number
-            self._vendor_name =  self._dev.manufacturer
+            if self._dev.idProduct == 0x3748:  # Workaround for unprintable characters in the ST-Link V2 probes
+                self._serial_number = bytes(self._dev.serial_number, 'utf-16')[2::2].hex().upper()
```
flit: Unfortunately this line won't work for Python 2.7.

anthrax-0: Sorry, completely missed it. Is it better now?

pyocd/probe/stlink/usb.py
```diff
@@ -126,8 +128,12 @@ def __init__(self, dev):
         # reading the strings, everything is ok. This workaround doesn't cause any issues with
         # Linux or macOS.
         try:
-            self._serial_number = self._dev.serial_number
-            self._vendor_name =  self._dev.manufacturer
+            if self._dev.idProduct == 0x3748:  # Workaround for unprintable characters in the ST-Link V2 probes
```
flit: Do you know if it's only the V2 probe firmware (PID==0x3748) that has this issue? Or does it affect V2-1 and V3?

pyocd/probe/stlink/usb.py
flit: Looks good!

pyocd/probe/stlink/usb.py
anthrax-0: I'm not 100% sure, but we are using V2-1 without problems. On the other hand, we are using V2 as the boxed ready-to-use probe [here](https://www.st.com/en/development-tools/st-link-v2.html), and V2-1 as cuttable PCB as part of the Nucleos boards.

flit:
Here's the traceback from the test failure.

```
10:46:55 Traceback (most recent call last):
10:46:55   File "automated_test.py", line 325, in <module>
10:46:55     main()
10:46:55   File "automated_test.py", line 281, in main
10:46:55     board_list = ConnectHelper.get_all_connected_probes(blocking=False)
10:46:55   File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/core/helpers.py", line 79, in get_all_connected_probes
10:46:55     sortedProbes = sorted(allProbes, key=lambda probe:probe.description + probe.unique_id)
10:46:55   File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/core/helpers.py", line 79, in <lambda>
10:46:55     sortedProbes = sorted(allProbes, key=lambda probe:probe.description + probe.unique_id)
10:46:55   File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/probe/stlink_probe.py", line 67, in unique_id
10:46:55     return self._link.serial_number
10:46:55   File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/probe/stlink/stlink.py", line 155, in serial_number
10:46:55     return self._device.serial_number
10:46:55   File "/Users/arm/jenkins/workspace/pyocd/hosts/mac/pyocd/probe/stlink/usb.py", line 180, in serial_number
10:46:55     return self._serial_number
10:46:55 AttributeError: 'STLinkUSBInterface' object has no attribute '_serial_number'
```

pyocd/probe/stlink/usb.py
flit: I'll try to get an answer from ST.

In the meantime, this **if** statement needs an **else** clause that sets `self._serial_number` for other PIDs. Or you could leave original line 129 in place and just override for 0x3748. Sorry I didn't notice this before (but that's what CI is for! 😄 ).

schstm:
Hello,
The issue resides in the « not-updatable » part of ST-Link/V2 code. As long as there is no mechanism
of software renumeration of the USB on those old boards, the bug remains visible even when the board is no more in “firmware upgrade mode” (and no fix is possible through firmware upgrade).
The bug has been fixed in later boards (V2-1, V3, ...)

For information in our STLinkUSBDriver we implemented a workaround against the issue (post computing of serial number by the library), based on the size of the descriptor, not on the USB PID (wrong SN descriptor has a size of 26 bytes [the read bytes has to be converted into ASCII], good one have a size of 50 bytes).

Because the ST-Link/V2-1 boards, when in firmware upgrade mode (reachable with STLinkUpgrade application between “connection” and “upgrade” action), also enumerate with PID 0x3748 but don’t have the bug.

flit:
Hi @schstm , thanks very much for the detailed information!

pyocd/probe/stlink/usb.py
flit: Based on the info from @schstm, could you please change this to check for a serial number size of 26 characters?

Also maybe remove the hex conversion, so each UTF-16 char is split into two 8-bit ASCII chars, depending on what @schstm says.

anthrax-0: > In the meantime, this if statement needs an else clause that sets self._serial_number for other PIDs. Or you could leave original line 129 in place and just override for 0x3748. Sorry I didn't notice this before (but that's what CI is for! 😄 ).

Sorry for the typo, lost the **else** in the copy-paste.

anthrax-0:
@flit As far as I see in our probes, it's not just wrong decoding of the 8-bit chars. According to the USB spec, all string descriptors are UTF-16-LE encoded. In case of ST-Link probes, we always have higher byte of UTF-16 codepoint as 0x00 and lower byte often is a control symbol that should not be printed, instead of the ASCII printable value. 
 Unfortunately, I don't have the description of the STLinkUSBDriver output. From lengths that @schstm mentioned, it seems they read values directly from the libusb, so they get a raw bytes, while pyOCD gets string decoded by pyUSB.
 I was trying to get the same values as STM32 ST-Link util for windows and st-link util for linux. 

@schstm Can you confirm that just reading hexadecimal values of every second byte is enough?

schstm:
@anthrax-0
Yes the bug is that instead of returning UTF-16 values, the ST-Link firmware returned an hexadecimal value each 2 bytes, for instance instead of returning 00 41 00 35 for 'A5' it returned 00 A5.
The workaround looks like this:
	if (sNDescriptor[0] == 26) {
		// Size value returned by the "old" ST-Link DFU;
		// In that case the DFU returns 12 8-bits values, separated by 12 zeros;
		for (i = 0; i < 12; i++) {
			// Useful values reside at sNDescriptor[2],[4] ... [24]
			// Each 8-bits value will be expressed on 2 hexadecimal digits in sNString
			sprintf((char *) &(sNString[2 * i]), "%02hX", (unsigned short)sNDescriptor[2 * (i + 1)]);
where sNDescriptor is the raw USB desciptor (would be at pyUSB level in your case I think)

PR #610, opened 2019-04-12
cmsis-pack-manager integration

### LINKED ISSUE #563: cmsis-pack-manager integration
Integrate cmsis-pack-manager so it's much easier to manage CMSIS DFPs for adding target support.

A new `packs` subcommand should be added that gives the user the needed control without having to also learn the `pack-manager` tool. This also prevents having to distribute 2 tools for binary releases.

flit:
The cmsis-pack-manager project is made use of by pyOCD to make it substantially easier for users to manage CMSIS-Packs.

A new `pack` subcommand for the `pyocd` tool has been added. This subcommand can be used to refresh the global pack index, list installed packs, find device part numbers in available packs, and add target support by downloading select packs.

All a user needs to do now to add support for a new target is run a command like this:

```
pyocd pack --install da1458
```

This will first download the global pack index if it is missing, concurrently downloading all vendor indexes. Then it will install all packs that provide device part numbers matching the glob pattern 'da1458*'.

Complete usage of the `pack` subcommand:

```
usage: pyocd pack [-h] [-v] [-q] [-c] [-u] [-s] [-f GLOB] [-i GLOB] [-n]

optional arguments:
  -h, --help            show this help message and exit
  -v, --verbose         More logging. Can be specified multiple times.
  -q, --quiet           Less logging. Can be specified multiple times.
  -c, --clean           Erase all stored pack information.
  -u, --update          Update the pack index.
  -s, --show            Show the list of installed devices and packs.
  -f GLOB, --find GLOB  Look up a device part number in the index using a glob
                        pattern. The pattern is suffixed with '*'. Can be
                        specified multiple times.
  -i GLOB, --install GLOB
                        Download and install pack(s) to support targets
                        matching the glob pattern. The pattern is suffixed
                        with '*'. Can be specified multiple times.
  -n, --no-download     Just list the pack(s) that would be downloaded, don't
                        actually download anything.
```

When the target type set by the user doesn't match an builtin target or one of the packs explicitly listed with the `--pack` argument or `pack` user option, pyOCD will check the cmsis-pack-manager installed packs.

`pyocd list --targets` and `pyocd json --targets` will include any targets supported via installed cmsis-pack-manager packs.

Closes #563

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,14 @@ Requirements
 - macOS, Linux, or Windows 7 or newer
 - Microcontroller with an Arm Cortex-M CPU
 - Supported debug probe
-  - [CMSIS-DAP](http://www.keil.com/pack/doc/CMSIS/DAP/html/index.html) v1 (HID) and v2 (WinUSB),
-    such as an on-board debug probe using [DAPLink](https://os.mbed.com/handbook/DAPLink) firmware.
+  - [CMSIS-DAP](http://www.keil.com/pack/doc/CMSIS/DAP/html/index.html) v1 (HID),
+    such as:
+    - An on-board debug probe using [DAPLink](https://os.mbed.com/handbook/DAPLink) firmware.
+    - NXP LPC-LinkII
+  - [CMSIS-DAP](http://www.keil.com/pack/doc/CMSIS/DAP/html/index.html) v2 (WinUSB),
+    such as:
+    - Cypress KitProg3
+    - Keil ULINKplus
   - STLinkV2, either on-board or the standalone version.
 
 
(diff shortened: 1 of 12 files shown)


bridadan:
Does the `--add` argument need to always be supplied with the `--update` argument? Or are they two separate operations?

flit:
They are separate operations. Though I'm thinking of changing `--add` to automatically download the index if that hasn't been done yet. Another option is to always update the index for every `--find` or `--add` operation. This is similar to how homebrew always updates its recipes (via git) before performing a requested install. Probably with a `--no-update` option.

bridadan:
Got it, just curious. Thanks!

flit:
Changed `--add` to `--install`. Also, if either `--install` or `--find` are executed but there is not a pack index present, the index will be automatically downloaded instead of reporting an error. It only automatically downloads the index if it's not present.

Issue #579, opened 2019-03-18
flash_loader_test fails on DISCO_L475VG_IOT01A device


studavekar:
pyOCD : 0.18.0

Device : `DISCO_L475VG_IOT01A`

pyocd.yaml

```yaml
probes:
  0666FF555051897267232812: # Probe's unique ID.
    target_override:  stm32l475xg
    test_binary:      stm32l475vg_iot01a.bin
```

## Test command 
`python flash_loader_test.py -d`
## Linux

```
------ Test Basic Sector Erase ------
INFO:pyocd.flash.loader:Erasing sector 0x080ff000 (2048 bytes)
TEST FAILED

```

## Windows
```
------ Test Basic Sector Erase ------
INFO:pyocd.flash.loader:Erasing sector 0x080ff000 (2048 bytes)
TEST FAILED
```

flit:
Able to reproduce.

PR #631, opened 2019-04-25
Auto-detection of STLink targets

flit:
This brings in support for auto-detection of the target associated with on-board STLinkV2-1 probes, so it works just as transparently as with DAPLink. Also brought in the list of ST board IDs from mbed-os-tools.


jeromecoutant:
Hi
It seems that this re-implements mbedls ?
Can we imagine that pyOCD reuse mbedls board id table ?

flit:
Yes, it uses a highly stripped-down version of the mbed-ls (aka mbed-os-tools) code. The main reason for not reusing the original code was because it does a lot of other checks, and supports other probes. By removing these features we can make it simpler and faster.

Unfortunately we can't reuse the mbed-ls platform table because it doesn't contain the target type or test binary name that pyOCD needs. However, we're looking at adding this information to the Mbed platforms database, so pyOCD (and mbed-ls) can download it at runtime.

flit:
Fyi, the only CI tests that failed were known issues with STM32 devices, like #579.

jeromecoutant:
Can we still use a target not defined in this board_ids list ?

flit:
Yes, of course! The `--target` argument and `target_overide` user option are still available.

PR #633, opened 2019-04-26
Add stm32l4xxxx target

rgf97:
This is the beginning of a contribution that aims not only to add new STM32 builtin targets but also to reorganize and simplify them. It will consist more precisely of adding one builtin target by big family (L4, L1, F3, F4 ...).
So this is the STM32L4xx family part.


flit:
Hi @rgf97, how will this work to support all the individual part numbers for a given family? Do you plan to use auto-detection of the part number via the IDCODE register?

Another option I've been considering is to automatically install a CMSIS-Pack for a given part number. For example, if the user set the target to "stm32f476rg" and that target was not supported builtin, and a Pack was not already installed, then pyOCD would automatically download the appropriate pack. (With a user option to disable auto pack downloads.) The benefit here is that there's no target support to maintain in pyOCD, and new targets are immediately available without having to update pyOCD.

rgf97:
Hi @flit, the idea behind this contribution was to simplify the STM32 target supports. And, this command is the most simplified thing. Can't wait to use` pyocd pack` command. When will it be available?
Then i am going to delete this pull.

PR #649, opened 2019-05-06
Log the debug probe's firmware version

flit:
For STLink probes, it logs the version in the same style that STMicro uses, such as "V2J33M25".

For CMSIS-DAP, it simply reports whatever the probe returns for the firmware version string from the [DAP_Info](http://arm-software.github.io/CMSIS_5/DAP/html/group__DAP__Info.html) command. If the version string is empty, no version is logged.

### DIFF
--- a/pyocd/probe/stlink/stlink.py
+++ b/pyocd/probe/stlink/stlink.py
@@ -17,6 +17,7 @@
 from .constants import (Commands, Status, SWD_FREQ_MAP, JTAG_FREQ_MAP)
 from ...core import exceptions
 from ...coresight import dap
+from ...utility.mask import bfx
 import logging
 import struct
 import six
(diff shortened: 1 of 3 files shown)


pyocd/probe/stlink/stlink.py
```diff
@@ -131,7 +131,10 @@ def get_version(self):
             #   8-9: ST_VID
             #   10-11: STLINK_PID
             response = self._device.transfer([Commands.GET_VERSION_EXT], readSize=12)
-            hw_vers, _, self._jtag_version = struct.unpack('<3B', response[0:3])
+            hw_vers, _, self._jtag_version, self._msc_version = struct.unpack('<4B', response[0:4])
+
+        self._version_str = "V%dJ%dM%d" % (self._hw_version, self._jtag_version, self._msc_version)
+        LOG.debug("STLink probe %s firmware version: %s", self.serial_number, self._version_str)
```
jeromecoutant: Seems that "LOG" is not compiling :-)

and maybe info level could be OK ?

flit: Oops! I cherry picked this change from another branch where I've cleaned up the log usage. I'll just wait until that branch is merged in to complete this PR.

The reason I used debug level is that I try to keep info level output to the minimum. If the STLink version is old enough that it won't work, a warning will be logged. Otherwise the user probably doesn't need to know the version.

flit:
We're having some issue with tests in CI that are intermittently failing, but enough tests passed by visual inspection.

PR #650, opened 2019-05-07
Logging configuration

flit:
This PR accomplishes two primary goals:
- Normalize logging usage in pyOCD. Every module that writes log messages now uses a module-specific logger with the name of that module, for instance `pyocd.coresight.cortex_m`.
- Enable logging configuration to be customized. A `logger` user option can be set to a [logging configuration dictionary](https://docs.python.org/2.7/library/logging.config.html#logging-config-dictschema) to fully configure all aspects of logging. Full documentation is included.

A few related changes were included:
- Fatal errors are logged by the `pyocd` tool at the CRITICAL level.
- The various LOG_X flags in some modules to control trace logs were converted to trace loggers that have a default level of CRITICAL, so they won't output anything unless overridden by a logging configuration. This lets you control trace output without having to modify source.
- GDB server and packet threads set the thread name (useful if you set a log formatter that prints the thread name).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -8,6 +8,7 @@
 - [Target support](target_support.md)
 - [Configuration](configuration.md)
 - [User scripts](user_scripts.md)
+- [Configuring logging](configuring_logging.md)
 - [Debugging multicore devices](multicore_debug.md)
 - [Introduction to the pyOCD Python API](python_api.md)
 - [Python API examples](api_examples.md)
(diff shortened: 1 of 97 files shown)


flit:
@hakehuang You asked for better log control long ago, it's _finally_ here! 😀

hakehuang:
@flit nice work! thanks a lot

PR #672, opened 2019-05-23
Musca A, B related changes, update flash algo

avinashmehtadelhi:


flit:
`page_size` doesn't appear to be right in the changes. (`blocksize` is an alias to both `page_size` and `sector_size`.) `sector_size` is the erase sector size of the flash, while `page_size` is the programming granularity. On the QSPI, I believe the page size was 256 bytes.

Does this fix the issues with the Musca-A1 flash algo that we discussed earlier?

avinashmehtadelhi:
Earlier issue comes into picture when new flash algo that works in DAPLink is used in pyOCD. Its still under investigation. This fix will enable the older flash algo in pyOCD to work with updated DAPLink firmware. The  test program and erase page will still fail because only 256KB of flash is memory mapped in muscaA. Consider this as a patch till the updated flash algo is added to the pyOCD project.

Issue #680, opened 2019-05-31
Target cy8c6xxa: pyocd-gdbserver: does not work with latest version 0.21.0


farrenv:
pyocd gdbserver does not work for the target cy8c6xxa for the cypress Cy8PROTO-062-4343W target with release 0.21.0.  It does work with version 0.16.1 when the target was introduced.
version 0.21.0
```
$ pyocd gdbserver
0000985:INFO:board:Target type is cy8c6xxa
0001021:INFO:dap:DP IDR = 0x6ba02477 (v2 rev6)
0001037:INFO:ap:AP#0 IDR = 0x84770001 (AHB-AP var0 rev8)
0001049:INFO:ap:AP#1 IDR = 0x84770001 (AHB-AP var0 rev8)
0001063:INFO:ap:AP#2 IDR = 0x24770011 (AHB-AP var1 rev2)
0001078:INFO:rom_table:AP#0 ROM table #0 @ 0xf1000000 (designer=034 part=102)
0001090:INFO:rom_table:AP#1 ROM table #0 @ 0xf0000000 (designer=034 part=102)
0001099:INFO:rom_table:[0]<e00ff000:ROM class=1 designer=43b part=4c0>
0001100:INFO:rom_table:  AP#1 ROM table #1 @ 0xe00ff000 (designer=43b part=4c0)
0001109:INFO:rom_table:  [0]<e000e000:SCS-M0+ class=14 designer=43b part=008>
0001115:INFO:rom_table:  [1]<e0001000:DWT-M0+ class=14 designer=43b part=00a>
0001121:INFO:rom_table:  [2]<e0002000:BPU class=14 designer=43b part=00b>
0001126:INFO:rom_table:[1]<f0002000:CTI class=9 designer=43b part=9a6 devtype=14 archid=1a14 devid=0:0:1040800>
0001133:INFO:rom_table:[2]<f0003000:MTB-M0+ class=9 designer=43b part=932 devtype=31 archid=0a31 devid=0:0:0>
0001144:INFO:rom_table:AP#2 ROM table #0 @ 0xe00ff000 (designer=034 part=102)
0001152:INFO:rom_table:[0]<e0080000:CTI class=9 designer=43b part=906 devtype=14 archid=0000 devid=0:0:40800>
0001158:INFO:rom_table:[3]<e008e000:TPIU-M3 class=9 designer=43b part=923 devtype=11 archid=0000 devid=0:0:ca1>
[...]

PR #682, opened 2019-06-03
Fixed Cypress PSoC6 creation of CortexM subclass instances

flit:


loverdeg-ep:
Does this address #680?

PR #697, opened 2019-07-10
Cumulative update for Cypress PSoC6 targets

Cypress-OpenOCD:
* All flash algorithms updated to latest version
* Added support for new secure PSoC64 target
* Added BoardIDs for new kits
* Added support for external QSPI memory


brianesquilona:
@flit I am not sure why the ci- morph is is not reported. Changes looks good to me too, probably fixed some. Can we merge this soon?

PR #700, opened 2019-07-19
Load Segments from ELF files, not just Sections

Cypress-OpenOCD:
Correct way of loading elf files is to load Segments defined in Program Headers, not just Sections. Some elf files may contain data which belong to some Segment but does not belong to any Section. With current implementation this data is skipped during programming.

### DIFF
--- a/pyocd/flash/loader.py
+++ b/pyocd/flash/loader.py
@@ -27,7 +27,7 @@ from .flash_builder import (FlashBuilder, get_page_count, get_sector_count)
 from ..core.memory_map import MemoryType
 from ..core import exceptions
 from ..utility.progress import print_progress
-from ..debug.elf.elf import (ELFBinaryFile, SH_FLAGS)
+from elftools.elf.elffile import ELFFile
 from ..utility.compatibility import FileNotFoundError_
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


Cypress-OpenOCD:
Steps to reproduce are simple:
- Erase the device
- Program attached elf file with vanilla pyOCD
- Open SEGGER Ozone and try to program the same elf:

J-Link: Flash download: Bank 0 @ 0x10000000: **1 range affected (1536 bytes)**
J-Link: Flash download: Total time needed: 1.531s (Prepare: 0.067s, Compare: 1.217s, Erase: 0.081s, Program: 0.083s, Verify: 0.072s, Restore: 0.010s)

- Program with pyOCD, verify with OpenOCD:

Error: checksum mismatch - attempting binary compare
diff 0 address 0x10008a98. Was 0x00 instead of 0x80
diff 1 address 0x10008a99. Was 0x00 instead of 0xb5
diff 2 address 0x10008a9a. Was 0x00 instead of 0x88
....

With this patch
Ozone:
J-Link: Flash download: Bank 0 @ 0x10000000: Skipped. **Contents already match**

OpenOCD:
verified 35332 bytes in 0.131700s (261.989 KiB/s)

[ble_app.elf.gz](https://github.com/mbedmicro/pyOCD/files/3410778/ble_app.elf.gz)

flit:
If an ELF has data that belongs to a segment but not a section, it is an invalid ELF. The Arm linker has a known bug (at least, we strongly believe it's a bug, but the linker team doesn't necessarily agree…) that causes it generate ELF files with some data not covered by a section.

From ELF specification v1.2 (with my bolding):

<blockquote>
<b>Sections contain all information in an object file</b> except the ELF header, the program header table, and the section header table. Moreover, object files' sections satisfy several conditions.

- Every section in an object file has exactly one section header describing it. Section headers may exist that do not have a section.
- Each section occupies one contiguous (possibly empty) sequence of bytes within a file.
- Sections in a file may not overlap. No byte in a file resides in more than one section.
- An object file may have inactive space. The various headers and the sections might not "cover'' every byte in an object file. The contents of the inactive data are unspecified.
</blockquote>

pyocd/flash/loader.py
```diff
@@ -197,17 +198,24 @@ def _program_hex(self, file_obj, **kwargs):
...
-                    and ((section.flags & (SH_FLAGS.SHF_ALLOC | SH_FLAGS.SHF_WRITE)) == SH_FLAGS.SHF_ALLOC)
-                    and (section.length > 0)
-                    and (section.region.is_flash)):
-                LOG.debug("Writing section %s", repr(section))
-                self._loader.add_data(section.start, section.data)
+        elf = ELFFile(file_obj)
+        for segment in elf.iter_segments():
+            if segment.header.p_type == 'PT_LOAD' and segment.header.p_filesz != 0:
+                addr = segment['p_paddr']
+                if sys.version_info[0] == 3:
```
flit: I think you can use `bytearray(segment.data())` here instead of switching on the Python version.

If not, please use the helpers in `pyocd.utility.conversion` or `pyocd.utility.compatibility` for data conversion. (Or add one if necessary.)

Cypress-OpenOCD: Done

PR #718, opened 2019-08-23
Update target_STM32L031x6.py

michieldwitte:
The double buffering does not work.
```
pyocd.core.exceptions.TransferFaultError: SWD/JTAG Transfer Fault @ 0x20002000-0x200027ff
```


flit:
Strange, I wonder what causes it to fail? Thanks for the fix.

PR #735, opened 2019-10-06
Connect mode option

flit:
This change replaces the `halt_on_connect` user option with `connect_mode`. There are four supported connection modes:
- "halt": Stop core(s) without changing PC.
- "pre-reset": Perform a hardware reset prior to halting the core(s).
- "under-reset": Attempt to connect while holding hardware reset asserted. Uses reset vector catch to halt before executing the first instruction. Reset is released after connecting. Not all devices will support this.
- "attach": Connect without halting the core(s).

Connect under reset is not fully working with the NXP Kinetis target family. This needs further debugging and will be fixed in a coming PR.


loverdeg-ep:
@trowbridgec, @millerc-ep

flit:
The first check in the connect tests is failing. I'm merging anyway and will release an alpha because it's needed by the Mbed Studio team.

40Grit:
Feels like the Studio team is making all the right moves.

Excited to see what @arekzaluski  has in store for TechCon.

PR #736, opened 2019-10-08
Some small commander changes

flit:
- Added new `load` command that can load any supported file type, such as binary, Intel hex, or ELF.
- Added a new `fill` command to write patterns to memory. (Does not support flash.)
- Hex dumps, e.g. from `read` commands, include a printable ASCII representation.
- Added `c` and `continue` aliases for `go` command to match gdb.
- `go` command prints more accurate information about the resulting target state, mostly importantly no longer reporting failure if the device just entered sleep, as Mbed OS and other IoT apps often do.
- `halt` command reports the resulting target state if it's not halted.
- Fixed regression with `show option` command.
- Changed some error reporting message prefixes.

### DIFF
--- a/pyocd/tools/pyocd.py
+++ b/pyocd/tools/pyocd.py
@@ -41,11 +41,12 @@ from ..probe.pydapaccess import DAPAccess
 from ..probe.debug_probe import DebugProbe
 from ..coresight.ap import MEM_AP
 from ..core.target import Target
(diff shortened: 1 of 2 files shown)


40Grit:
@farrenv @trowbridgce @aglass0fmilk

flit:
No new failures over #735.

PR #737, opened 2019-10-09
Add support for Nuvoton NuMaker-M252KG board

ccma1120:


pyocd/target/builtin/target_M252KG6AE.py
```diff
@@ -0,0 +1,126 @@
...
+    'sector_sizes': (
+        (0x0, 0x200),
+    )
+}
+
+class M252KG6AE(CoreSightTarget):
+
+    VENDOR = "Nuvoton"
+
+    has_fpu = False
```
flit: You don't need to specify `has_fpu` as it will be auto-detected.

flit:
I'd appreciate it if you could rebase and squash down to a single commit. (You'll need to force push after rebasing.)

Thanks for the contribution! 😄

PR #741, opened 2019-10-14
Add support for NuMaker-M252KG board

ychsu-tf:


0xc0170:
@flit Travis fixed, seems to be pending after restart here (sorry for hijacking this PR but was testing it here).

PR #749, opened 2019-10-31
Fixed regressions introduced in #706

flit:
Deep copies were getting tripped up on ELF files from pack-based targets. Added `MemoryMap.clone()` method instead of using deep copy and removed deep copy support for memory regions.

Plus some little goodies:
- Added equality operator for `MemoryRegion` that also compares type and attributes.
- Added indexing operator for `MemoryMap`.
- Finally fixed `ELFBinaryFile` __del__ issue with _owns_file attribute by adding a call to `hasattr()`.

### DIFF
--- a/pyocd/core/memory_map.py
+++ b/pyocd/core/memory_map.py
@@ -248,15 +248,15 @@ class MemoryRegion(MemoryRangeBase):
                 length=self.length,
                 **self._attributes
                 )
- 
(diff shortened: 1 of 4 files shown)


flit:
cc @jeromecoutant

jeromecoutant:
Hi
It's working with master branch
Thx

flit:
Thanks for testing!

PR #757, opened 2019-11-07
Make commander perform mass erase if no arguments (address and count)…

jsiverskog:
… given.

### DIFF
--- a/pyocd/tools/pyocd.py
+++ b/pyocd/tools/pyocd.py
@@ -115,8 +115,8 @@ COMMAND_INFO = {
             },
         'erase' : {
             'aliases' : [],
-            'args' : "ADDR [COUNT]",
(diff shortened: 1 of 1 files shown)


flit:
It's worth discussing if we also want to change the `pyocd erase` command to similarly perform a mass erase if no args (either `--chip` or sectors list). I intentionally made `pyocd erase` a no-op if you don't explicitly indicate what erase type you want to prevent accidental erasing of data, but maybe that's not a real concern?

jsiverskog:
I just wanted a way to erase the device from commander without having to specify address and count. This is already possible from `pyocd` (by providing `--chip`). I can't see any reason not to have that as default also for `pyocd`.

jsiverskog:
What is left to do in order to get this merged @flit ?

PR #778, opened 2019-12-06
Fix uninitialized variable reference in loader.py

samkearney:


flit:
The Travis-CI failure was due to a broken colorama version 0.4.2 (apparently fixed in 0.4.3).

flit:
Thanks for the fix!

Issue #775, opened 2019-12-04
nRF52 APPROTECT, Cannot read_ap *or* write_ap with recent ST-Link firmwares (J33,J34,J35)


nocko:
I am enabling APPROTECT in a production programming setup using pyOCD:

``` python
from pyocd.core.helpers import ConnectHelper
from pyocd.core.memory_map import FlashRegion
from pyocd.flash.flash import Flash

with ConnectHelper.session_with_chosen_probe(target_override="nrf52") as session:
    target = session.board.target
    target.reset_and_halt()
    fl = FlashLoader(session)
    mmap = target.memory_map

    uicr2 = FlashRegion(start=0x10001200, end=0x1000120f, blocksize=0x10, is_testable=False)
    uicr2.flash = Flash(target, mmap.get_boot_memory().algo)
    uicr2.flash.region = uicr2
    mmap.add_region(uicr2)
    fl.add_data(APPROTECT_REG, b'\x00')
    fl.commit()
```
It's a little dance to dynamically add the *second page* (?) of the UICR to the memory map. I wasn't sure how to upstream this (poorly documented) memory region... it's not clear if it can be erased or why the address gap with the rest of the UICR... &c. This hack works well enough for a write once operation.

I also added a way to check if the weable is locked out:
``` python
def wearable_locked():
    session = start_uninit_session()
    target = session.board.target
    locked = not target.dp.read_ap(1 << 24 | 0xc)
    session.close()
    return locked
```
and several functions that should work while the wearable is locked (via CTRL-AP @1):
``` python
def recover():
    session = start_uninit_session()
    target = session.board.target
    target.dp.init()
    target.dp.power_up_debug()
    target.dp.write_ap(1 << 24 | 4, 1)
    while target.dp.read_ap(1 << 24 | 8):
        sleep(0.1)
    _ctrl_ap_reset(session)
    session.close()

def _ctrl_ap_reset(session=None):
    if not session:
        session = start_uninit_session()
    target = session.board.target
    target.dp.write_ap(1 << 24 | 0, 1)
    sleep(0.001)
    target.dp.write_ap(1 << 24 | 0, 0)
```

This works as intended with ST-Link v2 firmware 2J29S7, but with later firmwares (including all of those still available for download from ST (2J33xx, 2J34xx, 2J35xx) have stopped working.

[...]

PR #780, opened 2019-12-12
Adds APPROTECT security check and unlock on nRF52

nocko:
As discussed in #775 

With these changes, auto_unlock works with nRF52 targets:

``` bash
(venv) $ pyocd commander --target nrf52840
0000228:WARNING:target_nRF52:NRF52840 APPROTECT enabled: will try to unlock via mass erase
Connected to NRF52840 [Halted]: I3FSNZOV
>>> 
```
and the commander has reasonable UX around the function:

``` bash 
(venv) $ pyocd commander --target nrf52840 -N
>>> initdp
>>> status
Security:       Locked
>>> unlock
>>> status
Security:       Unlocked
>>> reinit
>>> status
Security:       Unlocked
Core 0 status:  Halted
>>> 
```

Tested locally against an nRF52 DK and and nRF52840 DK using CMSIS DAP.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -1,0 +1,194 @@
+# Security and Protection Features
+
+Many targets support some way of disabling JTAG/SWD access or protecting the flash from read-back or write.
+
+If you attempt to run pyOCD against a locked target, you'll encounter a message like this:
+
+``` bash
+(venv) ~/devel/contrib/pyocd$ pyocd commander --target nrf52
+0000265:WARNING:target_nRF52:NRF52832 APPROTECT enabled: not automatically unlocking
+Exception while initing board: No cores were discovered!
+Traceback (most recent call last):
+  File "devel/contrib/pyocd/pyocd/tools/pyocd.py", line 769, in connect
+    self.session.open(init_board=not self.args.no_init)
+  File "devel/contrib/pyocd/pyocd/core/session.py", line 371, in open
+    self._board.init()
+  File "devel/contrib/pyocd/pyocd/board/board.py", line 83, in init
+    self.target.init()
+  File "devel/contrib/pyocd/pyocd/core/coresight_target.py", line 164, in init
+    seq.invoke()
+  File "devel/contrib/pyocd/pyocd/utility/sequencer.py", line 208, in invoke
+    resultSequence = call()
+  File "devel/contrib/pyocd/pyocd/core/coresight_target.py", line 298, in check_for_cores
+    raise exceptions.DebugError("No cores were discovered!")
+pyocd.core.exceptions.DebugError: No cores were discovered!
+```
+
+pyOCD has noticed that the chip is locked. It attempts routine initialization anyway, but fails to find any cores.
+
+## Supported Targets
+
+pyOCD currently supports the security features of the NXP Kinetis and Nordic Semiconductor nRF52 families. 
+
+Disabling the security features on supported targets is very straight-forward.
+
+*WARNING: **This will erase all data on the chip***
+
+You can add the option`auto_unlock` to your [configuration](/configuration.md):
+
+```bash
+(venv) ~/devel/contrib/pyocd$ pyocd commander --target nrf52 -O auto_unlock
+0000264:WARNING:target_nRF52:NRF52832 APPROTECT enabled: will try to unlock via mass erase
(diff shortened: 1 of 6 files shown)


nocko:
Brief reminder that this **doesn't** work with ST-Link v2 firmware > 2J29S7 . IIRC, that's not a deal breaker because the combination isn't common... All the testing was with LPC-Link2.

pyocd/target/family/target_nRF52.py
```diff
@@ -0,0 +1,134 @@
...
+                self.dp.create_1_ap(AXI_AP_NUM)
+            else:
+                LOG.warning("%s APPROTECT enabled: not automatically unlocking", self.part_number)
+        else:
+            LOG.info("%s not in secure state", self.part_number)
+
+    def is_locked(self):
+        if not self.ctrl_ap:
+            if CTRL_AP_NUM not in self.aps:
+                self.dp.create_1_ap(CTRL_AP_NUM)
```
flit: Is this really necessary? It doesn't hurt, so I'm mostly curious. And if you need it here, wouldn't it also be needed on `mass_erase()`?

nocko: It's not necessary, but without it the *status* command in commander throws an exception in no_init mode.

I didn't have it initially, but it seems like a lot to ask of a (potentially new) user to know the chip specifics (that another AP exists) and run `makeap 1` before *status* works.

Now that you mention it, if the user runs *unlock* before *status*, it'll fail in the same way.

They should both go the same way, but I'll defer you you on which way that is.

nocko: I promoted this logic to a separate method *_ensure_ctrl_ap*, and referenced it in both *mass_erase* and *is_locked*. The more I think about it, the more I think that commander should just "do the right thing".

I cannot think of a reason why the user would give us the specific target, but then *not* want to init the AP when needed for a requested operation.

flit: `--no-init` is primarily meant for bring-up of new chips and debugging low-level issues. So having commander just do the right thing, at least in this case, may be problematic as it may cause unexpected side effects. The way I've been using `--no-init` (and sort of how I meant it to be used) is to manually create an AP before any operation that requires it by running the `makeap` command. (This is similar to how you need to run the `initdp` command first in no-init mode. Of course, I realise none of this is documented… ☹️)

Still, I'm happy to merge the PR this way if that's what you prefer.

nocko: Let's do it your way, but can you wait on the merge? 

I'll look into adding some documentation around using commander for manual unlock.

Kinetis is similar enough that such documentation would serve both.

flit: Yes, of course, I'm happy to wait til you're ready!

nocko:
Documentation content is preliminary. Wouldn't mind comments on the structure and scope. Seems like it'll be a primer on commander, but I'm open to refactoring.

Will flesh out content and squash the commits before merge.

flit:
Thanks, I really appreciate the docs. Even preliminary/draft docs are far better than none.

nocko:
I'm pretty happy with the scope of the documentation now. 

I pared it down to a brief discussion on security features *instead* of a primer on pyOCD commander. I'll reread it again tomorrow for spelling grammar, but it's probably quite near final form.

flit:
Great, thanks. Just let me know when you're happy and I'll merge.

flit:
Great, thanks. Just let me know when you're happy and I'll merge.

nocko:
I'm happy with it now. Thanks.

docs/security.md
```diff
@@ -0,0 +1,194 @@
+# Security and Protection Features
+
+Many targets support some way of disabling JTAG/SWD access or protecting the flash from read-back or write.
+
+If you attempt to run pyOCD against a locked target, you'll encounter a message like this:
```
flit: `auto_unlock` defaults to True for every command except commander. So for most users that run the gdbserver, pyOCD will be able to connect by automatically mass erasing if the target is locked.

flit:
Really great document, thanks 😄 . There's only the one comment about auto unlocking on commander versus the other commands that should be clarified.

PR #786, opened 2019-12-21
Some nRF52 family fixes, plus docs improvements

flit:
- Fixed a reference to an init task in the just-merged nRF52 family class that was changed recently, causing an exception.
- General improvements to family matching for CMSIS-Pack based targets.
- Added an nRF52 family matcher that enables the security features in the nRF52 family class for CMSIS-Pack based nRF52 targets.
- Some docs updates, including updates to the new security features doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Configuration and customization is supported through [config files](docs/configu
 Requirements
 ------------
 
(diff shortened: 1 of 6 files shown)


flit:
cc @nocko

nocko:
Thanks for the fixes. Looks great.

PR #793, opened 2019-12-27
Added OTP program algorithm for HC32F46x. Fixed a bug of HC32x120 fla…

lennvn:
…sh algorithm. Added new targets HC32L07x, HC32L110, HC32L13x and HC32L19x of HDSC MCU.


flit:
Thanks for the PR! Would you mind rebasing to address the conflict on `svd_data.zip`?

PR #828, opened 2020-03-02
Fix issue #826. More robust detection of FreeRTOS

### LINKED ISSUE #826: FreeRTOS detection not working in gdbserver. Could also be erroneous when it does.
Sorry, long:

I want to describe the problem first, but I also have a solution. Please read and see if you can **entertain a PR.** I hope it is not a stupid user error.

**Problem**
I think I have everything right, but my FreeRTOS executable fails to be recognized. I looked at the implementation it may not work depending on the version of GCC/another-compiler you are using or optimization levels used.

I see the pyOCD source code for FreeRTOS detection. It is expecting that the symbols defined in the executable will come in the exact order as in the source. For me, the source matches pyOCD's expectation but GCC is rearranging them and I am not aware of any C/ABI spec that says how these variables have to be arranged. Neither do I know of GCC command-line options to control the behavior. So, I am stuck.

GCC version: I do not have the liberty to switch versions, but should it matter?
```text
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2017-q4-major) 7.2.1 20170904 (release) [ARM/embedded-7-branch revision 255204]
```
If this is a known issue and there is a workaround as to how to manage the compiler/linker options. *Problem solved.*

But spec. wise it is a fragile implementation to expect the FreeRTOS variables to be ordered exactly right. **Once all the addresses/sizes match the experience is wonderful though. Super nice**

**Solutions**
I have a couple. I implemented both and I wanted to present my preferred one, which I tested it many simulated failure situations.

The root cause is that the gdb server protocol does not provide for a symbol size. Just the address via 'qSymbol' and we try to guess by using symbol address offsets. But if the user supplies the elf file (optional) then we can find out the exact sizes. It is already a command-line option.

Very few lines of code changes

**Notes**
I am very familiar with OpenOCD source code as well and I did not want to implement what they did, but it is a fallback option that I also implemented. It works, but not pretty and difficult for the user to get it right.

I am one of two maintainers of the Cortex-Debug extension for VSCode. Also contributed to VSCode generic debugger

haneefdm:
Fixes #826

We were relying on the compiler/linker placing the symbols in the order they are declared and packed. I have simple executables that are jumbling up the FreeRTOS symbols we are interested in almost random ways. This makes it impossible to detect proper FreeRTOS and also a chance for improperly identifying the FreeRTOS data.

Since the user has the option of supplying an elf file, we can use it to check symbol sizes. We also make sure the address matches what gdb tells us via the `qSymbol` query as a backup because we want to have one version of the truth and cross-check the gdb vs elf versions

A couple of other changes I made
* When user supplies an elf file, we were setting the elf object only on core[0]. They should be set for all cores. In my case, what I was interested was in core#1. I verified that we can now do that for all cores. This change is very much needed for the fix
* A change I made that was not needed in pyocd/coresight/cortex_m.py . When I ran pylint, it found an undefined variable. It had to be a typo/oversight, so I fixed it.

Note: OpenOCD uses a symbol named `uxTopUsedPriority` as a workaround. The problem with using that method is that it has been long been deprecated by FreeRTOS and defining and making it survive the compiler/linker is super difficult. I think what I implemented is a much better method.

### DIFF
--- a/pyocd/core/coresight_target.py
+++ b/pyocd/core/coresight_target.py
@@ -86,8 +86,9 @@ class CoreSightTarget(Target, GraphNode):
             self._elf = None
         else:
             self._elf = ELFBinaryFile(filename, self.memory_map)
-            self.cores[0].elf = self._elf
-            self.cores[0].set_target_context(ElfReaderContext(self.cores[0].get_target_context(), self._elf))
+        for core_number, core in self.cores.items():
+            self.cores[core_number].elf = self._elf
+            self.cores[core_number].set_target_context(ElfReaderContext(self.cores[core_number].get_target_context(), self._elf))
 
     def select_core(self, num):
         """! @note Deprecated."""
--- a/pyocd/core/coresight_target.py
+++ b/pyocd/core/coresight_target.py
@@ -86,7 +86,7 @@ class CoreSightTarget(Target, GraphNode):
             self._elf = None
         else:
             self._elf = ELFBinaryFile(filename, self.memory_map)
-        for core_number, core in self.cores.items():
+        for core_number in range(len(self.cores)):
             self.cores[core_number].elf = self._elf
             self.cores[core_number].set_target_context(ElfReaderContext(self.cores[core_number].get_target_context(), self._elf))
 
--- a/pyocd/core/coresight_target.py
+++ b/pyocd/core/coresight_target.py
@@ -86,9 +86,9 @@ class CoreSightTarget(Target, GraphNode):
             self._elf = None
         else:
             self._elf = ELFBinaryFile(filename, self.memory_map)
-        for core_number in range(len(self.cores)):
-            self.cores[core_number].elf = self._elf
-            self.cores[core_number].set_target_context(ElfReaderContext(self.cores[core_number].get_target_context(), self._elf))
+            for core_number in range(len(self.cores)):
+                self.cores[core_number].elf = self._elf
+                self.cores[core_number].set_target_context(ElfReaderContext(self.cores[core_number].get_target_context(), self._elf))
 
     def select_core(self, num):
         """! @note Deprecated."""
--- a/pyocd/coresight/cortex_m.py
+++ b/pyocd/coresight/cortex_m.py
@@ -1291,7 +1291,7 @@ class CortexM(Target, CoreSightCoreComponent):
         if mask & CortexM.DEMCR_VC_BUSERR:
             result |= Target.VectorCatch.BUS_FAULT
         if mask & CortexM.DEMCR_VC_MMERR:
-            result |= TargetVectorCatch.MEM_FAULT
+            result |= Target.TargetVectorCatch.MEM_FAULT
         if mask & CortexM.DEMCR_VC_INTERR:
             result |= Target.VectorCatch.INTERRUPT_ERR
         if mask & CortexM.DEMCR_VC_STATERR:
--- a/pyocd/rtos/freertos.py
+++ b/pyocd/rtos/freertos.py
@@ -349,23 +349,27 @@ class FreeRTOSThreadProvider(ThreadProvider):
         vPortEnableVFP = self._lookup_symbols(["vPortEnableVFP"], symbolProvider)
         self._fpu_port = vPortEnableVFP is not None
 
+        elfOptHelp = " Try using the --elf option." if self._target.elf is None else ""
+
         # Check for the expected list size. These two symbols are each a single list and xDelayedTaskList2
         # immediately follows xDelayedTaskList1, so we can just subtract their addresses to get the
         # size of a single list.
         delta = self._symbols['xDelayedTaskList2'] - self._symbols['xDelayedTaskList1']
+        delta = self._get_elf_symbol_size('xDelayedTaskList1', self._symbols['xDelayedTaskList1'], delta)
         if delta != LIST_SIZE:
-            LOG.warning("FreeRTOS: list size is unexpected, maybe an unsupported configuration of FreeRTOS")
+            LOG.warning("FreeRTOS: list size is unexpected, maybe an unsupported configuration of FreeRTOS." + elfOptHelp)
             return False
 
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -88,4 +88,5 @@ pyocd/_version.py
 pyocd.yaml
 pyocd_user.py
 
-
+# Mac stuff
+.DS_Store
```
flit: Normally `.DS_Store` is ignored in your `~/.gitignore_global` file. Is it not there on your system? (I'm not sure how that gets installed.)

pyocd/core/coresight_target.py
```diff
@@ -86,8 +86,9 @@ def elf(self, filename):
             self._elf = None
         else:
             self._elf = ELFBinaryFile(filename, self.memory_map)
-            self.cores[0].elf = self._elf
-            self.cores[0].set_target_context(ElfReaderContext(self.cores[0].get_target_context(), self._elf))
+        for core_number in range(len(self.cores)):
```
flit: The reason it was only applied to the first core is that each core normally has different code. What's really needed is support for an ELF file for each core. But I'm happy to include this change as a temporary workaround.

pyocd/core/coresight_target.py
```diff
@@ -86,8 +86,9 @@ def elf(self, filename):
...
+            self.cores[core_number].elf = self._elf
+            self.cores[core_number].set_target_context(ElfReaderContext(self.cores[core_number].get_target_context(), self._elf))
```
flit: This line should probably be changed to handle `self._elf` being None. Otherwise a None can be passed into `ElfReaderContext`.

pyocd/coresight/cortex_m.py
```diff
@@ -1291,7 +1291,7 @@ def _map_from_vector_catch_mask(mask):
         if mask & CortexM.DEMCR_VC_BUSERR:
             result |= Target.VectorCatch.BUS_FAULT
         if mask & CortexM.DEMCR_VC_MMERR:
-            result |= TargetVectorCatch.MEM_FAULT
+            result |= Target.TargetVectorCatch.MEM_FAULT
```
flit: Good catch! Thanks.

pyocd/rtos/freertos.py
```diff
@@ -505,4 +509,25 @@ def get_is_running(self):
             return False
         return self._target_context.read32(self._symbols['xSchedulerRunning']) != 0
 
-
+    def _get_elf_symbol_size(self, name, addr, defValue):
+        if self._target.elf is not None:
+            symInfo = None
+            try:
+                symInfo = self._target.elf.symbol_decoder.get_symbol_for_name(name)
+            except:
```
flit: Can this except clause be made more specific? I try not to have catchall exception handlers in pyocd (which is admittedly sometimes difficult).

pyocd/rtos/freertos.py
```diff
@@ -505,4 +509,25 @@ def get_is_running(self):
             return False
         return self._target_context.read32(self._symbols['xSchedulerRunning']) != 0
 
-
+    def _get_elf_symbol_size(self, name, addr, defValue):
```
flit: Please rename `defValue` to `calculated_size` to follow PEP8 naming of parameters and make the parameter's purpose more apparent. (At some point I'd like to turn on PEP8 checks in CI, so this would fail.)

.gitignore
haneefdm: Sorry, I did not know that. I will revert that change.

flit: It might be set up that way only for certain installation methods of git, i.e. brew versus macports. Or it might have been set up on my system by the Fork UI.

haneefdm: I had to do the following first and then edit the `~/.gitignore_global`  file and then reload VSCode. Thanks for letting me know. It did not work without doing the following
```sh
git config --global core.excludesfile ~/.gitignore_global 
```

pyocd/core/coresight_target.py
haneefdm: Yes, I thought about that as well. One elf file per core. In my case, the elf file is shared.

pyocd/rtos/freertos.py
haneefdm: It should ideally never fail, I did not know what to make it specific to. I looked at other places where this was done. I will change to `RuntimeError` The only reason I added that try/except was because I was dereferencing a bit much and wanted a catch-all.

pyocd/rtos/freertos.py
haneefdm: Okay

haneefdm:
@flit I believe I made all the changes and pushed them. Unit tests are passing, my own tests passed.

pyocd/core/coresight_target.py
haneefdm: I think there are use cases for both. Shared and separated elf file. It depends on how the memory for the device was implemented. I will keep an eye out for the non-shared elf file situation. I wish the gdb server interface provided a symbol address as well as the size. I think I will ask for an enhancement with gdb but I doubt they will do it as it might break things

pyocd/core/coresight_target.py
haneefdm: I indented it back to the way it was except for the loop

flit:
Great, thanks.

flit:
Would you mind rebasing and squashing out the `.DS_Store` related commits?

haneefdm:
> Would you mind rebasing and squashing out the `.DS_Store` related commits? (Force push after the rebase so you can reuse this PR.)

@flit Done. It took me a while, sorry. I am always scared of rebasing, but I think I got it right. Rebasing wholesale is not so scary (if you have a backup).

flit:
Thanks! I too used to be very scared of screwing things up by rebasing, but it's not as tricky as it seems at first. And you can do things like tag or branch to make a backup. I've also been using the interactive rebase UI in [Fork](https://git-fork.com) a lot. And it can automatically create backups for you.

haneefdm:
Again, thanks for the pointers, a careful/thoughtful review and the speed. Much obliged

I will do better next time

PR #829, opened 2020-03-03
Set correct size for nRF52x UICR.

Dirbaio:
The UICR flash region on nRF52 is bigger than 0x100 bytes. This PR raises it to 0x400, which is the smallest multiple of 0x100 that covers all registers in the docs. This allows flashing these special registers with pyOCD.

Tested working on a real hardware nRF52840.

Relevant docs:

nrf52840: https://infocenter.nordicsemi.com/topic/ps_nrf52840/uicr.html?cp=4_0_0_3_4_0#topic
nrf52832: https://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52832.ps.v1.1/uicr.html?cp=4_2_0_13_0#topic


flit:
Thanks for this fix too!

PR #830, opened 2020-03-03
Fix bits vs bytes confusion in SWV.

Dirbaio:
SWV seems currently broken because it expects `width` to be in bits (8, 16, or 32) but the SWO code is providing it in bytes (1, 2, or 4).

This PR fixes that.

Tested working with real hardware.


flit:
Thanks very much for the fix.

PR #833, opened 2020-03-04
fix crash with zephyr threading

kiram9:
If running in eclipse or vscode, pyocd would throw an exception on connect, if early in the launch cycle of the app on the target because ZephyrThread.description would reference an invalid state index, causing a failure, make default return UNKNOWN string if there is a garbage state value.

### DIFF
--- a/pyocd/rtos/zephyr.py
+++ b/pyocd/rtos/zephyr.py
@@ -222,8 +222,8 @@ class ZephyrThread(TargetThread):
         return self._name
 
     @property
-    def description(self):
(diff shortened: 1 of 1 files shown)


pyocd/rtos/zephyr.py
```diff
@@ -222,8 +222,8 @@ def name(self):
         return self._name
 
     @property
-    def description(self):
-        return "%s; Priority %d" % (self.STATE_NAMES[self.state], self.priority)
+    def description(self):
```
0xc0170: can you please remove this space

kiram9: done

PR #845, opened 2020-04-10
NXP: i.MXRT1052 flash algo update for QSPI flash

jhqian:
- Updated flash algo for the onboard QSPI flash on mimxrt1050-evk
- Updated flash size for QSPI flash
- Add reset catch

Signed-off-by: Ryan QIAN <jianghao.qian@nxp.com>


flit:
Thanks @jhqian!

PR #851, opened 2020-04-20
Fix NXP kl28z flash programming issue

Hoohaha:


pyocd/target/builtin/target_MKL28Z512xxx7.py
```diff
@@ -190,24 +188,17 @@ def __init__(self, link):
...
         seq = super(KL28x, self).create_init_sequence()
-
-        seq.wrap_task('discovery',
-            lambda seq: seq
-                # The KL28 will lock up if an invalid AP is accessed, so replace the AP scan with a
-                # fixed list of known APs.
-                .replace_task('find_aps', self.create_kl28_aps)
-                # Before creating cores, determine which memory map should be used.
-                .insert_before('create_cores',
+        seq.replace_task('find_aps', self.create_kl28_aps)
```
flit, line 191: Have you tested against the latest code on master? The init sequence changed, so there isn't a "find_aps" task in the top level sequence. That's why I had changed it to use `wrap_task()` on "discovery".

Hoohaha: @flit 
It will raise keyerror "discovery" if use `wrap_task()`.  I guess it's relate to the parent `Flash_Kinetis` also did a `wrap_task` on "discovery".

But I tested it works in my local withtout wrap_task on the latest master.

flit: Fyi, I get a `KeyError: 'find_aps'` exception when using this branch. (This is based on fa5992a, same master commit that was current when you created your branch.)

Unfortunately, my FRDM-KL28Z seems to be in a weird state where it's locked but I can't mass erase it (it never completes). The MDM-AP Status LP Enabled and Very Low Power Mode bits are set, so I'm wondering if the device can't be mass erased because it's sleeping.

flit:
@Hoohaha Fyi, incorporated your fix into #873. It's been tested with the current HEAD of master and works fine.

PR #901, opened 2020-06-23
Add support for new revisions of secure PSoC64 devices

bohdan-tymkiv:
This change also adds support GenericMemAP class which allows to access memory of the target via Access Ports which do not have CPU core behind them.


flit:
Thanks @bohdan-tymkiv !

pyocd/__main__.py
```diff
@@ -49,6 +49,8 @@
 from .flash.eraser import FlashEraser
 from .flash.file_programmer import FileProgrammer
 from .core import options
+from .utility.cmdline import split_command_line
```
flit, line 52: Is `split_command_line()` used anywhere?

pyocd/coresight/generic_mem_ap.py
```diff
@@ -0,0 +1,153 @@
...
+
+from .component import CoreSightCoreComponent
+from ..core.target import Target
+
+LOG = logging.getLogger(__name__)
+
+DEAD_VALUE = 0
+
+
+class GenericMemAP(Target, CoreSightCoreComponent):
```
flit, line 27: Can we rename this class to `GenericMemAPTarget`?

Also, please add at least a one line doc comment for the class, doxygen style, e.g.`"""! @brief foo"""`.

Fyi: I'm thinking about adding a "target capabilities" set so the target can specify, for instance, whether breakpoints are supported.

pyocd/coresight/generic_mem_ap.py
```diff
@@ -0,0 +1,153 @@
...
+    def find_breakpoint(self, addr):
+        return None
+
+    def read_core_register(self, reg):
+        return DEAD_VALUE
+
+    def read_core_register_raw(self, reg):
+        return DEAD_VALUE
+
+    def read_core_registers_raw(self, reg_list):
```
flit, line 104: Just fyi on the r/w register methods… I'm working in the background on improving these methods so they can raise exceptions when reading or writing is invalid, and the rest of the code can handle that.

pyocd/target/builtin/cypress/flash_algos/license/EULA.txt
```diff
@@ -0,0 +1,55 @@
+CYPRESS END USER LICENSE AGREEMENT
```
flit, line 1: I will have to ask Arm lawyers to review this.

pyocd/__main__.py
```diff
@@ -691,6 +693,8 @@ def do_gdbserver(self):
                 if self._args.elf:
                     session.board.target.elf = os.path.expanduser(self._args.elf)
                 for core_number, core in session.board.target.cores.items():
+                    if isinstance(session.board.target.cores[core_number], GenericMemAP):
```
flit, line 696: Could you please add a comment here just mentioning why the GenericMemAP is skipped? It's obvious to us, but not necessarily other people. (Same for pyocd/tools/gdb_server.py.)

pyocd/__main__.py
bohdan-tymkiv: split_command_line is already imported in line 39, thanks!

PR #934, opened 2020-08-16
Debug probe client and server

flit:
This PR adds a TCP/IP client and server to enable remote access to debug probes. This version can be considered an experimental implementation intended to gather feedback and contributions. However, the currently functionality is useful enough that I wanted to make it available quickly.

Basic documentation is included in `docs/remote_probe_access.md`.

### Design 
The design is fairly simple in that there is a 1:1 correspondence between server instance (and thus TCP port) and debug probe. The protocol uses JSON-encoded messages, but is not JSON-RPC. Locking is fully supported, so many clients can connect to a single server without causing issues (at least at the lower levels; don't expect two gdb instances to behave when controlling a single core).

There are a lot of opportunities for improvements. The main addition I'd like to see is the ability to run a single server to share all connected probes.

### Server
There are three ways to run the probe server.

1. A new `server` subcommand. Runs only the probe server. This example runs the probe server in the background:

    ```
    $ pyocd server -q -u0240 &
    ```

2. The new `--probe-server` argument on the `gdbserver` subcommand, which causes a probe server to run alongside the gdbserver. This example demonstrates the option:

    ```
    $ pyocd gdb -u0240 --probe-server
    ```

3. Commander has a `probeserver` command that accepts `start` and `stop` actions.

As usual, if there is more than one debug probe connected and you do not use `-u` to select one by its unique ID, then pyOCD will present a prompt asking you to choose a probe to use.

### Client
Connecting another pyOCD instance to a probe server is easy. This is done by selecting the `remote` debug probe and passing the server address using the `--uid` (or `-u`) argument to any pyOCD subcommand where you would normally use a local debug probe. The syntax is `remote:<server>[:<port>]`, where server is a domain name or IP and port is the TCP port number. If left off, the default port of 5555 is used.

Example to connect a commander to a running probe server on the default port 5555:
```
$ pyocd cmd -uremote:localhost:5555 -v
```

Remote probes do not appear in the list of available probes printed by `pyocd list --probes`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ The `pyocd` command line tool gives you total control over your device with thes
 - `pack`: Manage [CMSIS Device Family Packs](http://arm-software.github.io/CMSIS_5/Pack/html/index.html)
     that provide additional target device support.
 - `commander`: Interactive REPL control and inspection of the MCU.
+- `server`: Share a debug probe with a TCP/IP server.
 - `list`: Show connected devices.
 
 The API and tools provide these features:
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,7 @@
 - [Configuration](configuration.md)
 - [Available options](options.md)
 - [User scripts](user_scripts.md)
+- [Remote probe access](remote_probe_access.md)
 
 #### Advanced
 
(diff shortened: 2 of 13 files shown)


loverdeg:
🥳 one small step towards MS language server?

PR #935, opened 2020-08-17
Fixed bug in dap.py - local call of function overwritten the input pa…

Gargy007:
Hello,
 I found bad hadling of now parameter in dap.py read_ap_multiple method.

Hope that this will be useful.


flit:
Hi @Gargy007 , did you see an issue caused by the original code? It was intentionally written that way so that the call to `self.probe_read_ap_multiple()` always returns a result callback, which then is always invoked in `read_ap_multiple_cb()`. The point is to a) simplify the number of code paths, b) ensure that the returned read callback is wrapped in the exception handler in `read_ap_multiple_cb()` when it is executed (so `_handle_error()` is called on an exception).

PR #939, opened 2020-08-20
USB HID Detach Kernel Driver fix for FreeBSD.

cederom:
This change catches all exceptions during kernel driver detach try.
It was blocking pyOCD on FreeBSD where HID was attached to DAPLink.

More generic kernel driver detach exception handling fixes crashes
on systems where user is not permitted (or even welcome) to detach
HID driver on the whole system just to use DAPLink.
On FreeBSD user is not allowed to detach HID kernel drivers because
that would impact whole OS and the rest of the workstation services,
while DAPLink works fine even with those HID drivers attached :-)

I have added one Debug message before detach and Warning after it fails.

### DIFF
--- a/pyocd/probe/pydapaccess/interface/pyusb_backend.py
+++ b/pyocd/probe/pydapaccess/interface/pyusb_backend.py
@@ -91,10 +91,12 @@ class PyUSB(Interface):
         self.kernel_driver_was_attached = False
         try:
             if dev.is_kernel_driver_active(interface_number):
+                LOG.debug("Detaching Kernel Driver of Interface %d from USB device (VID=%04x PID=%04x).", interface_number, dev.idVendor, dev.idProduct)
                 dev.detach_kernel_driver(interface_number)
                 self.kernel_driver_was_attached = True
-        except NotImplementedError as e:
+        except (NotImplementedError,usb.core.USBError) as e:
             # Some implementations don't don't have kernel attach/detach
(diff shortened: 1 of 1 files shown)


flit:
Cool, thanks @cederom !

pyocd/probe/pydapaccess/interface/pyusb_backend.py
```diff
@@ -91,10 +91,12 @@ def open(self):
         self.kernel_driver_was_attached = False
         try:
             if dev.is_kernel_driver_active(interface_number):
+                LOG.debug("Detaching Kernel Driver of Interface %d from USB device (VID=%04x PID=%04x).", interface_number, dev.idVendor, dev.idProduct)
                 dev.detach_kernel_driver(interface_number)
                 self.kernel_driver_was_attached = True
-        except NotImplementedError as e:
+        except:
```
flit, line 97: Can we make the exception handler catch all the errors that could be raised by pyusb/libusb without making it a catchall handler? Reason being that pylint and other tools will flag that as problematic code. It also will not allow an `InterruptException` (Ctrl-C) to pass.

cederom: Is `except (NotImplementedError,usb.core.USBError) as e` okay? :-)

flit:
sorry!

flit:
test

cederom:
TANK U SIR =)

PR #942, opened 2020-09-01
Fixed a buf of HC32x120's flash algorithm. Added new targets HC32F4A0xG, HC32F4A0xI and HC32M423 of HDSC MCU.

lennvn:


flit:
Thanks for the contribution! :)

Issue #855, opened 2020-04-25
Error with simultaneous GDB and SWV on CMSIS-DAP probe


adamgreig:
Hi! I'm having a somewhat confusing problem with SWV and GDB and I can't tell how pyOCD is meant to resolve this issue, so I'd welcome any guidance. The summary is that CMSIS-DAP probe responses to the regular poll for SWO data by the SWVReader thread (`DAP_SWO_Data` commands) are being received by the transfer thread (expecting `DAP_Transfer` responses), or vice versa, so either way an error is thrown, which means I can't use GDB while SWV is active.

I'm using a custom CMSIS-DAP probe, so I worry an implementation issue is causing the fault, but I don't understand how it could be mitigated. My probe implements both v1 and v2 endpoints but does not support SWO streaming on its own endpoint, so all responses are via `DAP_SWO_Data` command polling. The issue is the same when using either v1 or v2 backends. I don't have access to another CMSIS-DAP probe that also has SWO, and I can't see anything obvious to deal with this in the DAPlink firmware.

I'm using SWV via `pyocd gdbserver -O enable_swv=true -O connect_mode=attach` on a target that's constantly streaming some ITM data via SWO. There's a very low data rate, maybe 50 bytes/second. After starting pyocd I can telnet to port 4444 and view the correct ITM data, no problems there. I use `connect_mode=attach` as otherwise the target is halted at this stage which prevents ITM output. At this point the SWVReader thread is constantly calling `swo_read()`, which sends new `DAP_SWO_Data` commands and receives responses.

If I then try and attach a GDB session to port 3333, pyOCD sends some `DAP_Transfer` commands in a separate Python thread. That thread then reads the same USB endpoint as the SWVReader is using, and pretty quickly either it receives a response to the `DAP_SWO_Data` command, or the SWVReader thread receives a response to the `DAP_Transfer` command. Either way this immediately causes an error. As far as I can tell there's no locking or synchronisation in pyOCD to prevent this, and from looking at the USB logs and also inserting extra logging into pyOCD, the probe appears to correctly responding in-order.

Has anyone got SWV working at the same time as GDB using a CMSIS-DAP probe without the separate SWO endpoint? Am I missing something obvious? Thanks!

Here's a log snippet with extra logging added to pyusb_backend.py to print out transmitted and received buffers:
<details>

[...]

PR #947, opened 2020-09-06
many fixes for ITM/SWO/SWV

bcantrill:
This work addresses a total of 9 issues that were preventing ITM + SWO from working on pyOCD + LPC55S69 + CMSIS-DAP.  Some of these are particular to our use case, but many are generic to pyOCD + SWO.  

1. When ITM is enabled, TPR is incorrectly modified to *disable* unprivileged writes when it intended the exact opposite, causing unprivileged stimulus writes to be silently dropped.  This work modifies TPR as intended:  to allow all unprivileged stimulus writes.  

2. The remote probe interface doesn't correctly return SWO data; the `swo_read` fails on a bad JSON encoding.  This work correctly returns a list of integers representing bytes.  

3. On CMSIS-DAP, when enabling SWO when SWO is already enabled, SWO is not first disabled, causing subsequent configuration to fail and the SWO reader to malfunction.  This work always disables SWO before configuring it.

4. On CMSIS-DAP, the size sent in a `DAP_SWO_DATA` request does not correctly adjust for the protocol overhead, potentially causing the last four bytes of the SWO packet to be left unread.  These trailing bytes will then be read on a subsequent read, causing an exception when the data payload does not constitute the expected bytes.  This work adjusts the size sent in `DAP_SWO_DATA` to reflect the protocol overhead.

5. On CMSIS-DAP, both `CommandError` and `DeviceError` are constructed without error messages, making it hard to debug failures from a stack trace.  This work modifies the construction of these errors to always specify an error message.

6. For reasons that are unclear, the SWO parser silently drops any ITM instrumentation packet for any stimulus port other than port 0.  This work modifies the SWO parser to not drop any ITM packets.  

7. The GDB server creates a separate SWO reader for each core (even though there is nothing core-specific with respect to SWO).  Because there is no locking between these threads, having multiple threads reading SWO causes more or less causes rampant corruption of the ITM stream on multicore targets.  This work creates only a single SWO reader thread, regardless of the number of cores.  

8. There is no serialization between the SWO reader thread and the GDB server.  As a result, if SWO and GDB are both simultaneously active, both the SWO reader and the GDB server are likely to experience protocol errors.  (This is likely the cause of issue #855.) This work adds a very coarse grain locking scheme between the two to assure that only one thread is talking to the underlying probe at at a time.

9. On the LPC55, resets with SWO enabled will cause ITM stimulus writes in the target to hang:  because ITM is enabled but `TRACECLKDIV` and `TRACECLKSEL` are reset, the stimulus port is not disabled but the FIFO never becomes ready.  This work explicitly sets `TRACECLKDIV` and `TRACECLKSEL` on a reset on the LPC55.

### DIFF
--- a/pyocd/gdbserver/gdbserver.py
+++ b/pyocd/gdbserver/gdbserver.py
@@ -152,6 +152,9 @@ class GDBServer(threading.Thread):
         # Read back bound port in case auto-assigned (port 0)
         self.port = self.abstract_socket.port
 
+        # Coarse grain lock to synchronize SWO with other activity
+        self.lock = threading.Lock()
+
         self.session.subscribe(self.event_handler, Target.Event.POST_RESET)
 
         # Init semihosting and telnet console.
@@ -173,14 +176,20 @@ class GDBServer(threading.Thread):
             semihost_console = semihost_io_handler
         self.semihost = semihost.SemihostAgent(self.target_context, io_handler=semihost_io_handler, console=semihost_console)
         
+        #
+        # If SWV is enabled, create a SWVReader thread. Note that we only do
+        # this if the core is 0: SWV is not a per-core construct, and can't
+        # be meaningfully read by multiple threads concurrently.
+        #
         self._swv_reader = None
-        if session.options.get("enable_swv"):
+
+        if session.options.get("enable_swv") and core == 0:
             if "swv_system_clock" not in session.options:
                 LOG.warning("Cannot enable SWV due to missing swv_system_clock option")
             else:
                 sys_clock = int(session.options.get("swv_system_clock"))
                 swo_clock = int(session.options.get("swv_clock"))
-                self._swv_reader = SWVReader(session, self.core)
+                self._swv_reader = SWVReader(session, self.core, self.lock)
                 self._swv_reader.init(sys_clock, swo_clock, console_file)
 
         # pylint: disable=invalid-name
@@ -377,14 +386,18 @@ class GDBServer(threading.Thread):
                 LOG.error("Unknown RSP packet: %s", msg)
                 return self.create_rsp_packet(b""), 0
 
+            self.lock.acquire()
             if msgStart == 0:
                 reply = handler()
             else:
                 reply = handler(msg[msgStart:])
+            self.lock.release()
+
             detach = 1 if msg[1:2] in self.DETACH_COMMANDS else 0
             return reply, detach
 
         except Exception as e:
+            self.lock.release()
             LOG.error("Unhandled exception in handle_message: %s", e, exc_info=self.session.log_tracebacks)
             return self.create_rsp_packet(b"E01"), 0
 
@@ -546,13 +559,18 @@ class GDBServer(threading.Thread):
                 self.packet_io.interrupt_event.clear()
                 return self.create_rsp_packet(val)
 
+            self.lock.release()
+
             # Wait for a ctrl-c to be received.
             if self.packet_io.interrupt_event.wait(0.01):
+                self.lock.acquire()
                 LOG.debug("receive CTRL-C")
                 self.packet_io.interrupt_event.clear()
                 self.target.halt()
                 val = self.get_t_response(forceSignal=signals.SIGINT)
                 break
+
+            self.lock.acquire()
 
             try:
                 if self.target.get_state() == Target.State.HALTED:
--- a/pyocd/probe/pydapaccess/cmsis_dap_core.py
+++ b/pyocd/probe/pydapaccess/cmsis_dap_core.py
@@ -135,7 +135,7 @@ class CMSISDAPProtocol(object):
         resp = self.interface.read()
         if resp[0] != Command.DAP_INFO:
             # Response is to a different command
-            raise DAPAccessIntf.DeviceError()
+            raise DAPAccessIntf.DeviceError("expected DAP_INFO")
 
         if resp[1] == 0:
             return
@@ -162,11 +162,11 @@ class CMSISDAPProtocol(object):
         resp = self.interface.read()
         if resp[0] != Command.DAP_LED:
             # Response is to a different command
-            raise DAPAccessIntf.DeviceError()
+            raise DAPAccessIntf.DeviceError("expected DAP_LED")
 
         if resp[1] != 0:
             # Second response byte must be 0
-            raise DAPAccessIntf.CommandError()
+            raise DAPAccessIntf.CommandError("DAP_LED failed")
 
         return resp[1]
 
@@ -179,11 +179,11 @@ class CMSISDAPProtocol(object):
         resp = self.interface.read()
         if resp[0] != Command.DAP_CONNECT:
             # Response is to a different command
-            raise DAPAccessIntf.DeviceError()
+            raise DAPAccessIntf.DeviceError("expected DAP_CONNECT")
 
         if resp[1] == 0:
             # DAP connect failed
-            raise DAPAccessIntf.CommandError()
+            raise DAPAccessIntf.CommandError("DAP_CONNECT failed")
 
         return resp[1]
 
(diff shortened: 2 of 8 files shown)


flit:
Many thanks!

pyocd/probe/pydapaccess/cmsis_dap_core.py
```diff
@@ -528,6 +534,6 @@ def vendor(self, index, data):
 
         if resp[0] != Command.DAP_VENDOR0 + index:
             # Response is to a different command
-            raise DAPAccessIntf.DeviceError()
+            raise DAPAccessIntf.DeviceError("expected DAP_VENDOR0")
```
flit, line 537: The error message should be `"expected DAP_VENDOR%i" % index`.

pyocd/trace/swv.py
```diff
@@ -49,10 +49,7 @@ def receive(self, event):
         """
         if not isinstance(event, TraceITMEvent):
             return
-        
-        if not event.port == 0:
-            return
-        
+
```
flit, line 52: Fwiw, this restriction to port 0 was present because only port 0 is defined as the SWV stdio-style logging port. Instead of fixing it to port 0, we could make the port configurable.

The problem is that SWO reading and SWV processing are combined into a single class. Ultimately the intention is to migrate to a better design with a generic SWO reader thread that accepts a data flow graph for processing events, and SWV processing is done in a graph sink class.

pyocd/probe/tcp_probe_server.py
```diff
@@ -410,6 +411,9 @@ def _request__get_memory_interface_for_ap(self, ap_address_version, ap_nominal_a
             handle = None
         return handle
     
+    def _request__swo_read(self):
+        return list(self._probe.swo_read())
```
flit, line 415: Should the client convert the list back into a bytearray? It's not technically necessary,  but might be nice for symmetry.

pyocd/gdbserver/gdbserver.py
```diff
@@ -152,6 +152,9 @@ def __init__(self, session, core=None, server_listening_callback=None):
         # Read back bound port in case auto-assigned (port 0)
         self.port = self.abstract_socket.port
 
+        # Coarse grain lock to synchronize SWO with other activity
+        self.lock = threading.Lock()
```
flit, line 156: The `DebugProbe` classes have a lock that should be used instead of creating a new one, along with `lock()` and `unlock()` methods.

flit (changes requested):

flit:
Questions/comments:

**Locking**
I'm really curious what interactions between the gdbserver and SWV thread caused issues. The existing locks seem like they should take care of it. Although, these locks were only merged in (ed54d26) on June 18, this year, and released in v0.27.0.

- All `DebugProbe` instances have a lock accessible by `lock()` and `unlock()` methods. This is used by the DP code to serialize all DP/AP register accesses.
- `MEM_AP` has its own lock for all memory transactions performed on that AP.
- `DAPAccessCMSISDAP` serializes the SWO commands with other CMSIS-DAP commands.

**Client/server**
At some point, I'd like to move to Base64 encoding for data transfers between the client and server.

**Session**
In the stuff I'm working on right now (refactoring commands and integrating with gdbserver: [refactor/commander](https://github.com/flit/pyOCD/tree/refactor/commander) in my fork) the `Session` now keeps track of all the server objects (gdbservers and probeserver). It makes sense to also track the `SWVReader` that is shared  by the gdbservers.

pyocd/gdbserver/gdbserver.py
```diff
@@ -173,14 +176,20 @@ def __init__(self, session, core=None, server_listening_callback=None):
...
         
+        #
+        # If SWV is enabled, create a SWVReader thread. Note that we only do
+        # this if the core is 0: SWV is not a per-core construct, and can't
+        # be meaningfully read by multiple threads concurrently.
+        #
         self._swv_reader = None
-        if session.options.get("enable_swv"):
+
+        if session.options.get("enable_swv") and core == 0:
```
flit, line 186: There's a minor issue with limiting to core 0. The gdbserver subcommand accepts a `--core` argument so it only starts a gdbserver for the specified core(s). So if you only start up a gdbserver for core 1, SWV would not work with this check.

For now this change is fine to keep since it fixes the more common problem.

Probably the ultimate solution is to move the creation of all the console, semihosting, and SWV infrastructure to outside the `GDBServer` class.

bcantrill:
> Questions/comments:
> 
> **Locking**
> I'm really curious what interactions between the gdbserver and SWV thread caused issues. The existing locks seem like they should take care of it. Although, these locks were only merged in ([ed54d26](https://github.com/mbedmicro/pyOCD/commit/ed54d2689410d2ba0051b56265aa41a37343d395)) on June 18, this year, and released in v0.27.0.
> 
> * All `DebugProbe` instances have a lock accessible by `lock()` and `unlock()` methods. This is used by the DP code to serialize all DP/AP register accesses.
> * `MEM_AP` has its own lock for all memory transactions performed on that AP.
> * `DAPAccessCMSISDAP` serializes the SWO commands with other CMSIS-DAP commands.

I did all of my testing on latest, for whatever it's worth. SWO commands do not, unfortunately, serialize with all other CMSIS-DAP commands:  because the locking is so fine-grained, there is nothing to prevent (say) `read_reg` from being interfered with by a `swo_read` in the (large) window between writing to and reading from the interface.

It should also be noted that this really didn't work **at all** (threads immediately die with data corruption as they start to interfere with one another) -- to the point that it's hard to believe that anyone has really tested SWO on CMSIS-DAP.  So this should be immediately reproducible for anyone connecting GDB to a target that has SWO enabled (as described by @adamgreig in #855), should you be interested in the specific failure modes.

pyocd/gdbserver/gdbserver.py
bcantrill: Yes, though the right fix is to have the `SWVReader` not have any notion of core whatsoever, and to handle core-based event handling at a different layer -- but I was trying to get this working with minimal change, not rearchitect it.

pyocd/trace/swv.py
bcantrill: The use of a particular port for a particular purpose is entirely based on convention, not definition.  Making it configurable would be fine, though I would leave that to subsequent work.

pyocd/probe/pydapaccess/cmsis_dap_core.py
bcantrill: Agreed; good catch.

pyocd/gdbserver/gdbserver.py
bcantrill: The `DebugProbe` class locks are too fine-grained, unfortunately.

pyocd/probe/tcp_probe_server.py
bcantrill: I was trying to keep with what appeared to be the intended signature in the comments of returning a list of integers.

pyocd/gdbserver/gdbserver.py
flit: Yes, exactly. And of course I don't expect you to rearchitect it! 😄

flit:
Thanks. It did actually work at one point, but I was primarily using it without the gdbserver in custom scripts. It was partially merged upstream just so it didn't completely rot on a branch somewhere, and so hopefully someone would find it useful and help improve. 😉 Your work here is very much appreciated.

bcantrill:
> Thanks. It did actually work at one point, but I was primarily using it without the gdbserver in custom scripts. It was partially merged upstream just so it didn't completely rot on a branch somewhere, and so hopefully someone would find it useful and help improve.  Your work here is very much appreciated.

Happy to help!  Glad that we were able to get it all working. For whatever it's worth, the multithreaded nature of pyOCD means that you definitely want to test the GDB server concurrently with the SWO reader, as that was the source of several of these issues.  But this is hard to test in any automated fashion, and definitely has a significant MCU- and probe-specific component; it's an impossibly big matrix, in many ways.

flit:
Agreed, the complexity of testing some of the features (like RTOS support) has been an issue. For a while I've been intending to support device-specific tests, so the code under test gets verified on at least one device and the required test firmware is limited.

flit:
Did you want to fix the "DAP_VENDOR0" log message before I merged?

bcantrill:
> Did you want to fix the "DAP_VENDOR0" log message before I merged?

Yes, fixed that -- thanks for spotting it!

flit (approved):

PR #959, opened 2020-09-18
Add support for NuMaker-PFM-M487 board

ychsu-tf:


flit (changes requested):
Looks good. And thanks for fixing the bug with chip erase!

The only request is that you change the copyright on the new files from Arm to your own (or Nuvoton's if appropriate). If you wish, you may also add your copyright below existing copyrights on the files you modified.

flit (approved):

PR #972, opened 2020-10-05
Pemicro

Gargy007:
Hi Chris,
 take a look on my PEMICRO implementation.

It passed all tests :-)
Petr


flit:
Awesome, thanks!

Gargy007:
Hello Chris / PyOCD reviers, 
 I have problem that in new supporting package for PEMICRO is use typing introduced in Python 3.5. What is the common solution in PyOCD to enable module like this for Python2.7? Or another question - how long you want to support P2.7 in PyOCD - thanks.

Petr

Gargy007:
Hi Chris,
 it looks awesome, but it isn't. I'm afraid that you doesn't want to accept
my PR, due to used "typing" in PEMICRO package. This doesn't pass the build
tests. ANd as I already wrote, I'm still newbie in Python , so I don't
now currently any elegance solution.

And second thing - this is not depended on PEMICRO (tested as with CMSIS
DAP as J-LINK) the Concurrent test fails since version 27.2 (in 27.1 it
works as expected). You can try it. But I think that this is general issue
in new version of PyOCD and this is no show stopper for my PR.

[image: image.png]

If you have any idea for my "typing" issue, I would be more than happy -
I'm not willing to remove it, because I spent a lot of time to add it and I
think that tis is really useful.

Thanks
Petr


po 5. 10. 2020 v 16:29 odesílatel Chris Reed <notifications@github.com>
napsal:

Gargy007:
Hi Chris,
 The Python2.7 issue I will solve by moving all typing from PEMicro package into interface. So next commit this solve.

flit:
Yeah… Python 2 support is becoming a big pain. I'd very much like to start using typing, too. In fact, I just announced that it is now deprecated and some future version of pyocd will drop Python 2 support (though I don't have a plan for which version yet).

But! I have a solution for you. Pyocd supports plugins now for debug probes. We can make the PEMicro plugin a separate package, separate repo in the pyocd org, where you're the owner. Then the main pyocd package can list the PEMicro plugin package as a dependency for Python >3.5.

This also resolves a concern I have about incorporating shared library binaries in the core pyocd. That may be a personal preference rather than a real technical concern, but plugins allows us to cleanly separate things like this.

Regarding the concurrency.py test—that runs on our CI system for every PR, on all 3 OSes, and for both Py 2.7 + 3.7, and I haven't seen it fail. So I'm curious what might be different in your setup that causes the failure. What board are you testing with? That could possibly affect it.

cederom:
> Yeah… Python 2 support is becoming a big pain. I'd very much like to start using typing, too. In fact, I just announced that it is now deprecated and some future version of pyocd will drop Python 2 support (though I don't have a plan for which version yet).

Here in FreeBSD Python 2 support from the core / base system will be removed at the end of the year. No software should be using Python2 except they can provide themselves. If this support is blocking then Python3 may become a mandatory requirement for new pyOCD releases? :-)

flit:
>  If this support is blocking then Python3 may become a mandatory requirement for new pyOCD releases? :-)

That's right! 😄 I'm not sure exactly what pyocd release will drop Python 2 support, but based on feedback I've heard so far it will be quite soon—nobody has yet said that dropping Python 2 support will be an issue. My major concern was Mbed Studio, but they moved to Python 3 earlier this year.d

Gargy007:
Hi Gentlemens - as we discussed the Peicro support has been moved to separate package (with non Python 2.7 support). Coul you check this pull request if it's possible to use it and merge to master ?
Thanks

flit:
Closing this PR. The PEMicro plugin was implemented in a separate [pyocd-pemicro](/pyocd/pyocd-pemicro) repo.

Once pyocd/pyocd-pemicro#1 is addressed, I will add the plugin as a dependency for pyocd, only for Python 3.6 and above.

Issue #980, opened 2020-10-12
Gdbserver reports "overlapping regions in memory map" to Gdb for stm32l4 Devices


DrTobe:
When I use `pyocd gdbserver --target stm32l432kcux` to connect to my nucleo-l432kc board and connect `arm-none-eabi-gdb` with `target extended :3333` afterwards, I get the "warning: Overlapping regions in memory map: ignoring" message. Then, it is impossible to use `gdb` to flash the device.

The same issue is present on the nucleo-l412-rb-p board but pyocd's gdbserver works flawlessly with an stm32l071kbu mcu which I have available for comparison here. Since pyocd uses DFPs to handle different targets and the stm32 DFPs are organized according to the device families, I suspect that this issue could be present on (all) stm32l4 devices but not on stm32l0 devices, caused by the stm32l4 DFP.

I have skimmed the DFP for the L0 and the L4 families and found memory information in the two `.pdsc` files. Below, some lines that looked interesting to me:

`Keil.STM32L0xx_DFP.pdsc`
```xml
    <subFamily DsubFamily="STM32L071">
        [...]
        <!-- *************************  Device 'STM32L071KBUx'  ***************************** -->
        <device Dname="STM32L071KBUx">
          <memory id="IROM1"                                   start="0x08000000" size="0x00020000" startup="1" default="1"/>
          <memory id="IRAM1"                                   start="0x20000000" size="0x00005000" init   ="0" default="1"/>
          [...]
        </device>
```

`Keil.STM32L4xx_DFP.pdsc`
```xml
      <subFamily DsubFamily="STM32L432">
        [...]
        <memory id="IRAM1"                                 start="0x20000000" size="0x0000C000" init="0" default="1"/>
        <memory id="IRAM2"                                 start="0x10000000" size="0x00004000" init="0" default="0"/>
        [...]
        <!-- *************************  Device 'STM32L432KCUx'  ***************************** -->
        <device Dname="STM32L432KCUx">
          <memory name="Main_Flash"  access="rx"   id="IROM1"     start="0x08000000" size="0x00040000" startup="1" default="1"/>
          <memory name="SRAM"        access="rwx"  id="IRAM1"     start="0x20000000" size="0x0000C000" init="0"    default="1"/>
          [...]
        </device>
```

[...]

PR #984, opened 2020-10-15
Identify DFP memory regions by address space in addition to names

DrTobe:
This approach seems to be the most reliable according to
https://www.keil.com/pack/doc/CMSIS/Pack/html/pdsc_family_pg.html#element_memory
because only 'start' and 'size' attributes are required for memory
elements. Furthermore, it is allowed that inner memory elements
supersede outer memory elements, so they have to be identified reliably.

This commit solves issue #980 from the main pyocd repo.

### DIFF
--- a/pyocd/target/pack/cmsis_pack.py
+++ b/pyocd/target/pack/cmsis_pack.py
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2019 Arm Limited
 # Copyright (c) 2020 Men Shiyun
+# Copyright (c) 2020 Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -214,8 +215,8 @@ class CmsisPack(object):
                 prev_start, prev_size = get_start_and_size(prev_elem)
                 # Overlap: start or end between previous start and previous end
                 end = start+size
-                if ((start >= prev_start and start <= prev_start+prev_size) or
-                        (end >= prev_start and end <= prev_start+prev_size)):
+                prev_end = prev_start+prev_size
+                if (prev_start <= start < prev_end) or (prev_start <= end < prev_end):
                     if not self._warned_overlapping_memory_regions:
                         LOG.warning("Overlapping memory regions in file "+str(self._pack_file)+","
                                     " deleting outer region. Further warnings will be suppressed"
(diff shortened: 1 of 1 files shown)


flit:
Thanks for the fix!

I think the logic probably need to be more complex.

- We still need to keep the overriding by name, as according to the CMSIS team that is how overriding is supposed to work.
- What about partially overlapping regions?

pyocd/target/pack/cmsis_pack.py
```diff
@@ -173,27 +177,52 @@ def _extract_items(self, state_info_name, filter):
...
+            for k in list(map.keys()):
+                prev_pname = k[1]
+                if prev_pname != pname:
+                    continue
+                prev_elem = map[k]
+                prev_start, prev_size = get_start_and_size(prev_elem)
+                # Overlap: start or end between previous start and previous end
+                end = start+size
+                if ((start >= prev_start and start <= prev_start+prev_size) or
+                        (end >= prev_start and end <= prev_start+prev_size)):
```
flit, line 218: I believe the end addresses are off by 1. This will cause back to back regions, e.g. start=0+size=0x1000 followed by start=0x1000+size=0x100, to be considered overlapping. You can use a `<` instead of `<=` to solve this.

I'd write it like this to take advantage of Python's range comparisons:

```py
end = start + end
prev_end = prev_start + prev_size
if (prev_start <= start < prev_end) or (prev_start <= end < prev_end):
```

flit (changes requested):
Two requests and it's ready to go! 

- Fix the off by 1.
- Please add your copyright below the existing copyright in the file header.

Thanks for fixing this! 😄

pyocd/target/pack/cmsis_pack.py
DrTobe: Thanks for spotting this error and nice proposal using those "range comparisons", I didn't know those before. I will upgrade the PR as requested.

pyocd/target/pack/cmsis_pack.py
```diff
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2019 Arm Limited
 # Copyright (c) 2020 Men Shiyun
+# Copyright (c) 2020 Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V.
```
flit, line 4: Unfortunately Python 2.7 doesn't like the non-ASCII `ö` character.

If you add `# -*- coding: utf-8 -*-` as the first line it will be fixed.

DrTobe: Well ... Python2 :smile:

flit: Yeah… I'll be a happy person when it's time to completely drop Python 2 from pyOCD!

flit (approved):

PR #1014, opened 2020-11-28
Publish raw SWV stream to TCP port

### LINKED ISSUE #1009: Enhancement Request: Publish SWO/SWV stream to TCP port
Hi,

For a CMSIS-DAP v2 interface, would it be possible to publish the SWO output (the second USB IN stream) to a TCP port? That would allow utilities such as Orbuculum to connect directly to it (indeed, if it was on port 3443 it would be transparent and wouldn't even need any command line options). I think OpenOCD is supporting this in development builds now.

Regards
DAVE

patrislav1:
This patch implements a [raw SWO stream server](https://github.com/pyocd/pyOCD/issues/1009) for tools like [orbuculum](https://github.com/orbcode/orbuculum).

I introduced a new option `tpiu_port` that is set to 3443 by default. The TPIU raw server is started implicitly, when `enable_swv` is activated. (If this is not desirable, then the default could also be changed to zero and the user would have to set it explicitly to 3443 or whatever to enable it).

I tested it on a STM32 Nucleo board with `pyocd gdbserver --target stm32l433rctxp -O enable_swv=true -O swv_system_clock=16000000` and a `orbcat -c 1,"0x%08x\n"` on the other side.

Fixes #1009

### DIFF
--- a/pyocd/trace/swv.py
+++ b/pyocd/trace/swv.py
@@ -1,5 +1,6 @@
 # pyOCD debugger
 # Copyright (c) 2019-2020 Arm Limited
+# Copyright (c) 2020 Patrick Huesmann
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -27,6 +28,7 @@ from ..coresight.tpiu import TPIU
 from ..core.target import Target
 from ..core import exceptions
 from ..probe.debug_probe import DebugProbe
+from ..utility.server import StreamServer
 
 LOG = logging.getLogger(__name__)
 
@@ -155,6 +157,10 @@ class SWVReader(threading.Thread):
         if self._lock:
             self._lock.acquire()
 
+        tpiu_port = self._session.options.get('tpiu_port')
+        tpiu_server = StreamServer(tpiu_port, name="TPIU", is_read_only=True) \
+                      if tpiu_port else None
+
         # Stop SWO first in case the probe already had it started. Ignore if this fails.
         try:
             self._session.probe.swo_stop()
@@ -165,6 +171,8 @@ class SWVReader(threading.Thread):
         while not self._shutdown_event.is_set():
             data = self._session.probe.swo_read()
             if data:
+                if tpiu_server:
+                    tpiu_server.write(data)
                 self._parser.parse(data)
         
             if self._lock:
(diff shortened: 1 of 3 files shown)


pyocd/trace/swv.py
```diff
@@ -155,6 +157,10 @@ def run(self):
         if self._lock:
             self._lock.acquire()
 
+        tpiu_port = self._session.options.get('tpiu_port')
+        tpiu_server = StreamServer(tpiu_port, name="TPIU", is_read_only=True) \
+                      if tpiu_port else None
```
flit, line 162: The other servers in pyocd support setting the port to 0, which causes the OS to select any available high-numbered port. The actual port is then read back and printed in the "server started" log message. To do this for the raw SWV server might require adding a separate `swv_raw_server_enable`.

flit (commented):
Awesome, thanks!

Starting the raw server by default is ok.

patrislav1:
Thanks for the quick review. Updated the PR according to your suggestions and also added to the `options.md` (It looks a bit like it could have been auto generated from the options dict, however I didn't find generator code, so I updated it manually anyway).

By the way, there's still some additional magic that `openocd "tpiu config internal ..."` is doing to the target to enable SWO that pyocd is not doing yet. If I cold start the target and enable SWV from pyocd, then the STM32 gets stuck at `while (ITM->PORT[TRACE_CHANNEL].u32 == 0)` - if I run `openocd "tpiu ..."` then that is resolved.
Not sure if this SWO enable magic is something that is in the scope of pyocd or not, but I can imagine it creating some confusion if openocd handles it one way and pyocd handles it differently. WDYT?

40Grit:
Everything is finally coming together. The life of the embedded developer should soon be less frustrating.

flit:
Thanks. The `options.md` file is still manually updated at the moment (whereas `command_reference.md` is auto-generated).

Regarding the STM32, that's a good question. It probably should be handled by pyocd. There's a `trace_start()` method in `SoCTarget` that can be overridden. However, it's also a delegate method, so it can be implemented in a user script (`pyocd_user.py`). And I believe the STM32 CMSIS-Packs provide a [debug sequence](https://arm-software.github.io/CMSIS_5/Pack/html/debug_description.html) that enables the trace clock. Pyocd doesn't support debug sequences yet, but I'm [working on it](https://github.com/flit/pyOCD/tree/feature/debug_sequences) in the background. But handling this doesn't need to be part of this PR.

flit (approved):

PR #1022, opened 2020-12-08
Added code to unlock AP for lpc55s69

ttt-t3r:
Hi.
For the LPC55S69 mcu, core APs are locked and unavailable when running in the boot ROM. This happens for example when there is nothing in the flash ROM. 

This code detects if the the core APs are missing and attempts to unlock them.

It also includes code to check if the flash page containing the reset vector is blank, and omits the flash read and reset operations in that case (to avoid the core APs becoming locked again).

### DIFF
--- a/pyocd/target/family/target_lpc5500.py
+++ b/pyocd/target/family/target_lpc5500.py
@@ -1,5 +1,6 @@
 # pyOCD debugger
 # Copyright (c) 2019-2020 Arm Limited
+# Copyright (C) 2020 Ted Tawara
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -17,6 +18,8 @@
 from time import sleep
 import logging
 
+from ...utility.sequencer import CallSequence
+from ...core import exceptions
 from ...core.target import Target
 from ...coresight.coresight_target import CoreSightTarget
 from ...core.memory_map import (FlashRegion, RamRegion, RomRegion, MemoryMap)
@@ -43,8 +46,29 @@ FLASH_DATAW0            = 0x00034080
 FLASH_INT_STATUS        = 0x00034FE0
 FLASH_INT_CLR_STATUS    = 0x00034FE8
 FLASH_CMD_READ_SINGLE_WORD = 0x3
+FLASH_CMD_BLANK_CHECK      = 0x5
 
 BOOTROM_MAGIC_ADDR      = 0x50000040
+
+DM_AP                   = 2
+
+# Control and Status Word (CSW) is used to control the Debug Mailbox communication
+DM_CSW = 0x00
+# Debugger will set this bit to 1 to request a resynchronrisation
+DM_CSW_RESYNCH_REQ_MASK =    (1<<0)
+# Write only bit. Once written will cause the chip to reset (note that the DM is not
+# reset by this reset as it is only resettable by a SOFT reset or a POR/BOD event)
+DM_CSW_CHIP_RESET_REQ_MASK = (1<<5)
+
+# Request register is used to send data from debugger to device
+DM_REQUEST = 0x04
+
+# Return register is used to send data from device to debugger
+# Note: Any read from debugger side will be stalled until new data is present.
+DM_RETURN = 0x08
+
+## Debugger mailbox command to start a debug session (unlock debug).
+DM_START_DBG_SESSION = 7
 
 LOG = logging.getLogger(__name__)
 
@@ -52,19 +76,38 @@ class LPC5500Family(CoreSightTarget):
 
     VENDOR = "NXP"
 
+    # Minimum value for the 'adi.v5.max_invalid_ap_count' option.
+    _MIN_INVALID_APS = 3
+
     def create_init_sequence(self):
         seq = super(LPC5500Family, self).create_init_sequence()
         
-        seq.wrap_task('discovery',
-            lambda seq: seq
-                    .wrap_task('find_components', self._modify_ap1) \
-                    .replace_task('create_cores', self.create_lpc55xx_cores) \
-                    .insert_before('create_components',
-                        ('enable_traceclk', self._enable_traceclk),
-                        )
-            )
-        
+        seq.wrap_task('discovery', self.modify_discovery)
         return seq
+
+    def modify_discovery(self, seq):
+        seq.insert_before('find_aps',
+                          ('set_max_invalid_aps', self.set_max_invalid_aps)) \
+           .insert_before('find_components',
+                          ('check_locked_state', lambda : self.check_locked_state(seq))) \
+           .wrap_task('find_components', self._modify_ap1) \
+           .replace_task('create_cores', self.create_lpc55xx_cores) \
+           .insert_before('create_components',
+                          ('enable_traceclk', self._enable_traceclk),
+                        ) \
+           .append(('restore_max_invalid_aps', self.restore_max_invalid_aps))
+        return seq
+    
+    def set_max_invalid_aps(self):
+        # Save current option and make sure it is set to at least 3.
+        self._saved_max_invalid_aps = self.session.options.get('adi.v5.max_invalid_ap_count')
+        if self._saved_max_invalid_aps < self._MIN_INVALID_APS:
+            self.session.options.set('adi.v5.max_invalid_ap_count', self._MIN_INVALID_APS)
+    
+    def restore_max_invalid_aps(self):
+        # Only restore if we changed it.
+        if self._saved_max_invalid_aps < self._MIN_INVALID_APS:
+            self.session.options.set('adi.v5.max_invalid_ap_count', self._saved_max_invalid_aps)
     
     def _modify_ap1(self, seq):
         # If AP#1 exists we need to adjust it before we can read the ROM.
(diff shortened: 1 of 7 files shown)


pyocd/target/family/target_lpc5500.py
```diff
@@ -158,27 +220,48 @@ def reset_and_halt(self, reset_type=None):
...
-                 # Read the reset vector address.
-                reset_vector = self.read32(0x00000004)
+                LOG.info("required flash area is erased")
+                halt_only = True
+
+            # Use the flash programming model to check if the first flash page is readable, since
+            # attempted accesses to erased pages result in bus faults. The start and stop address
+            # are both set to 0x0 to probe the sector containing the reset vector.
+            self.write32(base + FLASH_STARTA, 0x00000000) # Program flash word start address to 0x0
+            self.write32(base + FLASH_STOPA, 0x00000000) # Program flash word stop address to 0x0
```
flit, line 248: Does using sector 0 also work if the device boots in Secure mode, with the initial VTOR_S set to 0x1000_0000?

pyocd/target/family/target_lpc5500.py
```diff
@@ -75,6 +83,29 @@ def _modify_ap1(self, seq):
         
         return seq
 
+    def check_locked_state(self, seq):
+        # Attempt to unlock cores if they are locked (flash is empty etc.)
+        self.was_core_locked = False
```
flit, line 88: Is `was_core_locked` used outside of this method? I mean, does it need to be an instance attribute?

pyocd/target/family/target_lpc5500.py
```diff
@@ -75,6 +83,29 @@ def _modify_ap1(self, seq):
...
+        if not DM_AP in self.aps:
+            exists = coresight.ap.AccessPort.probe(self.dp, DM_AP)
+            if not exists:
+                LOG.error("Error: no DMAP exists")
+                return
+            ap_addr = coresight.ap.APv1Address(DM_AP)
+            ap = coresight.ap.AccessPort.create(self.dp, ap_addr)
+            self.unlock()
```
flit, line 96: Can you pass this `ap` object to `unlock()` so it is used instead of doing low level AP/DP register accesses in `unlock()`?

pyocd/target/family/target_lpc5500.py
```diff
@@ -132,13 +163,44 @@ def trace_start(self):
...
+        self.dp.probe.write_ap(addr=addr + 0,   data=0x21)
+        self.dp.probe.flush()
+        sleep(0.1)
+        retval = self.dp.probe.read_ap(addr=addr + 0)
+        if retval != 0:
+            LOG.error("did not reset")
+            return
+        self.dp.probe.write_ap(addr=addr + 4, data=7)
+        self.dp.probe.flush()
+        retval = self.dp.probe.read_ap(addr=addr + 8) & 0xffff
```
flit, line 191: The AP/DP register accesses should be done using the `DebugPort` methods rather than going through the probe.  You can directly use the AP register access methods on `DebugPort` with a "full" AP register address, such as 0x0200_00FC to read the IDR. The appropriate DP registers will be updated if necessary to select the AP (at offset 24 in the addr) and bank (offset 4).

It's also not necessarily safe to directly call the probe methods, since that bypasses the register cache in `DebugPort`, and it can thus get out of sync.

flit (commented):
Well, here's the review anyway… 😅

flit:
I hope you can finish this, it would be most appreciated! Feel free to reuse this PR, just reopen and force push to update if necessary.

flit:
One more comment on the changes: please add your copyright below the existing copyrights in the `target_lpc5500.py` file.

pyocd/target/family/target_lpc5500.py
ttt-t3r: I'm not using secure mode so I'm not sure what should happen. The original code is reading the vector from hardcoded 0 address, that's what I'm using here. I think secure mode support would be out of scope for this issue.

pyocd/target/family/target_lpc5500.py
ttt-t3r: This has been removed

pyocd/target/family/target_lpc5500.py
ttt-t3r: Please see updated commit

pyocd/target/family/target_lpc5500.py
ttt-t3r: Thanks, I've passed in the AP created by the caller, but there's no flush() operation so that's still called directly.

flit:
Thanks for the update. I'm testing this right now.

pyocd/target/family/target_lpc5500.py
flit: Secure mode being out of scope for this PR is totally reasonable.

flit:
Do you mind if I push to your branch for this PR?

flit:
Fyi, you can see my changes on [my pull/1022 branch](/flit/pyOCD/tree/pull/1022).

If you're ok with me pushing to your branch, then I'll do that tomorrow and get the changes merged in. Or you can cherry pick my commits yourself. Then I'll kick off CI. Either way, please re-test on your side, and then I think it's ready to go.

I had to correct the logic in `check_locked_state()` for it to work. I also added some code to the `MEM_AP` class to be able to test whether the AP is enabled, then used that in the LPC55xx family. On my device, the APs were detected but disabled (`CSW.DEVICEEN` == 0) when the device was locked.

Thanks very much for putting this together! It was actually very timely for me, since I _just_ ran into a case where I needed it like 10 minutes after you created the PR. 😀

ttt-t3r:
> Do you mind if I push to your branch for this PR?

Hi no problems, thanks for looking into this. I just started looking into the code, so apologies if the code wasn't quite right.

ttt-t3r:
> Fyi, you can see my changes on [my pull/1022 branch](/flit/pyOCD/tree/pull/1022).
> 
> If you're ok with me pushing to your branch, then I'll do that tomorrow and get the changes merged in. Or you can cherry pick my commits yourself. Then I'll kick off CI. Either way, please re-test on your side, and then I think it's ready to go.
> 
> I had to correct the logic in `check_locked_state()` for it to work. I also added some code to the `MEM_AP` class to be able to test whether the AP is enabled, then used that in the LPC55xx family. On my device, the APs were detected but disabled (`CSW.DEVICEEN` == 0) when the device was locked.
> 
> Thanks very much for putting this together! It was actually very timely for me, since I _just_ ran into a case where I needed it like 10 minutes after you created the PR. 😀

Hi I'll take a look at it in a moment.
Glad it was useful!

pyocd/target/family/target_lpc5500.py
```diff
@@ -85,54 +103,63 @@ def _modify_ap1(self, seq):
...
+            return
+        
+        # The debugger mailbox should always be present.
         if not DM_AP in self.aps:
-            exists = coresight.ap.AccessPort.probe(self.dp, DM_AP)
-            if not exists:
-                LOG.error("Error: no DMAP exists")
-                return
-            ap_addr = coresight.ap.APv1Address(DM_AP)
-            ap = coresight.ap.AccessPort.create(self.dp, ap_addr)
```
ttt-t3r: Quick note: unless I'm missing some other changes, this doesn't work for me as DM_AP does not get created (`self.aps` is empty).

I'm using the CMSIS-DAP embedded in the LPC55S69-EVK (revision A2) board.

```
(venv3) [ttt@localhost pyocd-mbedmicro]$ python3 -m pyocd reset -t lpc55s69 -m hw
(venv3) [ttt@localhost pyocd-mbedmicro]$ python3 -m pyocd gdbserver -t lpc55s69 
0000464:INFO:board:Target type is lpc55s69
0000557:INFO:dap:DP IDR = 0x6ba02477 (v2 rev6)
0000571:ERROR:target_lpc5500:cannot request debug unlock; no debugger mailbox AP was found
0000571:ERROR:target_lpc5500:AP#0 was not found, unable to create core 0
0000637:CRITICAL:__main__:No cores were discovered!
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/__main__.py", line 398, in run
    self._COMMANDS[self._args.cmd](self)
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/__main__.py", line 742, in do_gdbserver
    with session:
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/core/session.py", line 343, in __enter__
    self.open()
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/core/session.py", line 461, in open
    self._board.init()
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/board/board.py", line 85, in init
    self.target.init()
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/core/soc_target.py", line 120, in init
    seq.invoke()
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/utility/sequencer.py", line 214, in invoke
    resultSequence = call()
  File "/home/ttt/contribs/pyocd/pyocd-mbedmicro/pyocd/coresight/coresight_target.py", line 220, in check_for_cores
    raise exceptions.DebugError("No cores were discovered!")
pyocd.core.exceptions.DebugError: No cores were discovered!
```

ttt-t3r:
I tried also using the scan_all_aps option, 
but the scan stops at the second invalid AP address, so if both cores are out of commission the loop does not get to the DM_AP (`discovery.py/_find_aps()`).

Should the code:
* have the invalid count be higher? 
* make the invalid count value an option? 
* force create the DM_AP if not found?

Something like
```
        # The debugger mailbox should always be present.
        dm_ap = None
        try:
            dm_ap = self.aps[DM_AP]
        except KeyError as error:
            dm_ap = coresight.ap.AccessPort.create(self.session.target.dp, coresight.ap.APv1Address(DM_AP))

        if dm_ap == None:
            LOG.error("cannot request debug unlock; no debugger mailbox AP was found")
            return
        
        # Perform the unlock procedure using the debugger mailbox.
        self.unlock(dm_ap)
```
would make the changes local to this target

pyocd/target/family/target_lpc5500.py
flit: That's strange, I wonder why the AHB-APs are completely invisible on your device but on mine they are visible but with `CSW.DEVICEEN` set to 0. My LPC55S69 is the controller on one of the new NXP MCU-Link probes (though it's a prototype build of the probe from several months ago, not brand new).

I'll handle this by adding something else on my todo list: control over the number of not-present APs found in a row that cause the AP scan to stop. Right now it's either a fixed 2 unless `scan_all_aps` is enabled. The LPC5500 family code will set this number to 3 so the DM-AP is always found.

ttt-t3r: Cool, I didn't know the debug probe had that chip. 
The chip on the EVK is a 100pin QFP and the date code says 19271B, so it is a 1B revision chip.

Having the not present APs configurable would help reduce the number of options to set on the command line. Thanks!

PR #1031, opened 2020-12-17
Bugfix in pyusb_backend: serial_number returned by pyusb may be None

mentha:
Hi. I was trying to use a GD-Link with pyocd 0.28.3, and found that `pyocd list` always crash in `cmsis_dap_probe.py` when trying to truncate `unique_id` at line 99. This GD-Link doesn't have a usb serial number string, so pyusb returns None as `unique_id`.

I found a related issue pyusb/pyusb#295. This behavior is expected but not well documented.

### DIFF
--- a/pyocd/probe/pydapaccess/interface/pyusb_backend.py
+++ b/pyocd/probe/pydapaccess/interface/pyusb_backend.py
@@ -169,9 +169,9 @@ class PyUSB(Interface):
             new_board = PyUSB()
             new_board.vid = board.idVendor
             new_board.pid = board.idProduct
-            new_board.product_name = board.product
-            new_board.vendor_name = board.manufacturer
-            new_board.serial_number = board.serial_number
+            new_board.product_name = board.product or ""
+            new_board.vendor_name = board.manufacturer or ""
+            new_board.serial_number = board.serial_number or ""
             boards.append(new_board)
 
         return boards
@@ -356,6 +356,8 @@ class FindDap(object):
         if cmsis_dap_interface is None:
             return False
         if self._serial is not None:
+            if self._serial == "" and dev.serial_number is None:
+                return True
             if self._serial != dev.serial_number:
                 return False
         return True
--- a/pyocd/probe/pydapaccess/interface/pyusb_backend.py
+++ b/pyocd/probe/pydapaccess/interface/pyusb_backend.py
@@ -21,6 +21,8 @@ import six
 from time import sleep
 import platform
 import errno
+from hashlib import sha1
+from base64 import b32encode
 
 from .interface import Interface
 from .common import (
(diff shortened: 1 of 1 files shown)


flit:
Thanks for the fixes!

How do they expect you to distinguish multiple connected GD-Links if it doesn't have a serial number string?

I think we need to generate a unique ID of some sort. Could we do something like generate a hash of the device path? That would at least allow selecting a specific GD-Link if multiple are connected, or selecting a GD-Link if there are more than one CMSIS-DAP probe connected.

Do you happen to know what hidapi and/or pywinusb return for serial/vendor/product string when not present? It would be nice to ensure GD-Link is supported on all OSes.

mentha:
Sorry for the late reply.

I tested GD-Link on a Windows 10 PC with version 0.28.3 from pip. It works but the unique id is an empty string.

pyocd/probe/pydapaccess/interface/pyusb_backend.py
```diff
@@ -169,9 +169,9 @@ def get_all_connected_interfaces():
             new_board = PyUSB()
             new_board.vid = board.idVendor
             new_board.pid = board.idProduct
-            new_board.product_name = board.product
-            new_board.vendor_name = board.manufacturer
-            new_board.serial_number = board.serial_number
+            new_board.product_name = board.product or ""
+            new_board.vendor_name = board.manufacturer or ""
+            new_board.serial_number = board.serial_number or ""
```
flit, line 174: I'd like to have some default serial number value here.

How about:

```py
base64.b32encode(
  struct.pack('Q', (
    hash(str(board.bus)) ^ hash(str(board.address))) ^ 
    hash(str(board.idVendor)) ^ hash(str(board.idProduct))
    )
  ).strip(b'=').decode('ascii')
```

This will generate a consistent serial number for a probe plugged into the same port and bus. For a random device I have plugged in, it generates "PWY3DCDKBUVTK".

Probably total overkill. 😅

flit (changes requested):

pyocd/probe/pydapaccess/interface/pyusb_backend.py
mentha: I have added this to pyusb backend, using SHA1. The builtin `hash` function is randomized on python startup.

flit: Much better than mine. 💯

flit:
It looks great, thanks for the update. One more request that I missed last time and it's ready to merge: please add your copyright below the existing copyrights in `pyusb_backend.py`.

flit:
@mentha Fyi, I took your changes from this PR and generalized them to be applied to pyusb_v2_backend as well. See #1077 for the result.

Please let me know what name you'd like your copyright to be under. For now it's just "mentha".

mentha:
Thanks. That's fine.

Issue #1056, opened 2021-01-20
No cores were discovered with LPC845


jpmzometa:
I'm getting the error 'No cores were discovered'. This happens after successfully flashing a valid program.
If I erase the chip (using MCUXpresso, the official IDE for LPC845), the pyocd is able to detect the cores again.

This might be related to issue #743 . But this an NXP LPC845, so the solution proposed there does not work here. 

```
pyocd erase -vv --chip -t lpc845 -O reset_type=hw
0000439:DEBUG:pyusb_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=1bcf PID=2c94). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000441:DEBUG:pyusb_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=05f3 PID=0007). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000441:DEBUG:pyusb_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=145f PID=0263). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000442:DEBUG:pyusb_v2_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=1bcf PID=2c94). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000443:DEBUG:pyusb_v2_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=05f3 PID=0007). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000443:DEBUG:pyusb_v2_backend:[Errno 13] Access denied (insufficient permissions) while trying to interrogate a USB device (VID=145f PID=0263). This can probably be remedied with a udev rule. See <https://github.com/mbedmicro/pyOCD/tree/master/udev> for help.
0000768:DEBUG:session:Project directory: /home/pablo/pyocd2
[...]

PR #1060, opened 2021-01-23
Add OKDO E1 lpc55s69 based board

newbrain:
One liner in board_ids.py to add support for [OKDO E1](https://www.okdo.com/p/okdo-e1-development-board/), a cheap board based on lpc55s69.

As the board is 'compatible' with the NXP EVK, in the sense that LEDs are connected to the same pins (though with different colours), there's no need for a new binary.

```
(venv3) ~\Git\pyOCD [OKDOE1 ≡]> pyocd list
  #   Probe                         Unique ID  
-----------------------------------------------
  0   OKDO-E1 LPC55S69 [lpc55s69]   15034033  
```

### DIFF
--- a/pyocd/board/board_ids.py
+++ b/pyocd/board/board_ids.py
@@ -195,6 +195,7 @@ BOARD_ID_TO_INFO = {
     "1304": BoardInfo(  "NuMaker-PFM-M487",     "m487jidae",        None,                   ),
     "1309": BoardInfo(  "NuMaker-M252KG",       "m252kg6ae",        None,                   ),
     "1310": BoardInfo(  "NuMaker-IoT-M263A",    "m263kiaae",        None,                   ),
+    "1503": BoardInfo(  "OKDO-E1 LPC55S69",     "lpc55s69",         "lpcxpresso55s69.bin",  ),
     "1549": BoardInfo(  "LPC1549",              "lpc1549jbd100",    None,                   ),
     "1600": BoardInfo(  "Bambino 210",          "lpc4330",          "l1_lpc4330.bin",       ),
     "1605": BoardInfo(  "Bambino 210E",         "lpc4330",          "l1_lpc4330.bin",       ),


flit:
Are the first 4 digits of the E1 unique ID always `1503`?

Either way, maybe we can add a name-based board identification mechanism in addition to the current ID-based method.

newbrain:
On all my boards, yes as I have only one!
I followed this instructions in the docs:
> This section only applies if your board has an on-board debug probe that either:
> * Uses the Arm DAPLink firmware. DAPLink presents the board ID as the first 4 characters of the USB serial number.

The integrated probe is a vanilla CMSIS-DAP v1.0.7 on  an LPC11U35.

Name based == passed as command line argument/in option file? Or something smarter?

flit:
> On all my boards, yes as I have only one!

I have an E1 too so I can check mine (if I can find it).

We might also see if the OKdo folks can confirm. Unless they are actively doing something in the firmware to guarantee the digits are consistent, it's probably not reliable enough to use.

> The integrated probe is a vanilla CMSIS-DAP v1.0.7 on an LPC11U35.

Only the DAPLink firmware reports the board ID in the first 4 digits of the USB serial number. (On-board STLinkV2-1 and V3 report the ID in an .htm file on the USB mass storage volume.)

What's weird is that the USB VID/PID matches the DAPLink VID/PID, otherwise your patch wouldn't work. Here: https://github.com/pyocd/pyOCD/blob/7be53bf551fe47636adcb770396d31f78752a14f/pyocd/probe/cmsis_dap_probe.py#L141

But the serial number format isn't a DAPLink serial number, which is derived from the probe controller's silicon unique ID. For the LPC11U35, [DAPLink uses the ReadUID ISP command](https://github.com/ARMmbed/DAPLink/blob/b22fe39fa42a37a64211be4140dfbc443c4dc2a5/source/hic_hal/nxp/lpc11u35/read_uid.c#L25) that returns a 128-bit value (which is actually a 16-character ASCII string according to the LPC11U35 RM). DAPLink then [formats this unique ID as hex and adds some IDs](https://github.com/ARMmbed/DAPLink/blob/b22fe39fa42a37a64211be4140dfbc443c4dc2a5/source/daplink/info.c#L178). So the 8-digit serial number `15034033` of your device doesn't match the serial number generated by DAPLink. Dunno.

> Name based == passed as command line argument/in option file? Or something smarter?

Something more automatic! Pyocd could match against the USB vendor and product strings. A fair number on-board debug probes have strings that identify the target.

A possible design would be a list in `board_ids.py` similar to `FAMILIES` in `pyocd/target/family/__init__.py`, with two regexes for the vendor and product strings. (Vendor might not need a regex, as in `FAMILIES`.) The test should probably be performed if no explicit target type was given, here in `board.py` https://github.com/pyocd/pyOCD/blob/903d429973739fb8838684024e2300cb39fa645d/pyocd/board/board.py#L34

flit:
Just looking at #1056, the log shows the serial number for the on-board probe for the LPC854 dev board, which has the same 8-character serial format as the E1 probe. This reminds me both that the NXP CMSIS-DAP firmware uses that serial number format, and that OKdo used the NXP firmware for the E1. I still don't get why the VID/PID is DAPLink's; NXP uses their own VID of 0x1FC9. Maybe I'm missing something and the VID/PID isn't DAPLink's.

newbrain:
I was about to address just what you put in https://github.com/pyocd/pyOCD/pull/1060#issuecomment-766243483:
The VID/PID for the E1 is 0x1FC9/0x0132 -> NXP / something (I found only one reference to this combiantion, for a different board).

I admit that my patch stems from a misunderstanding.
For some reason, I never found the github repo while googling for DAPLINK, so I naïvely assumed it was another name for the CMSIS-DAP!

In case it might help, here are the results for the only CMSIS-DAP probes I have:
- MIMXRT1020-EVK (removing the board_ids matching line) - It's got DAPLINK VID/PID
- LPCxpresso43S67 - 0x1FC9/0x0090
- OKDO E1 - 0x1FC9/0x0132
```
(venv3) ~\Git\pyOCD [ThreadX ≡ +0 ~1 -0 !]> pyocd list               
  #   Probe           Unique ID
------------------------------------------------------------------------
  0   ARM CMSIS-DAP   0226000047784e4500499004d9170010e561000097969900  
(venv3) ~\Git\pyOCD [ThreadX ≡ +0 ~1 -0 !]> pyocd list
  #   Probe                                           Unique ID  
-----------------------------------------------------------------
  0   NXP Semiconductors LPC-LINK2 CMSIS-DAP V5.183   JRARCQCR   
(venv3) ~\Git\pyOCD [ThreadX ≡ +0 ~1 -0 !]> pyocd list
  #   Probe                                          Unique ID  
----------------------------------------------------------------
  0   NXP Semiconductors LPC11U3x CMSIS-DAP v1.0.7   15034033   
```


Ah well, it's not a big deal!
Of course,feel free to close this PR, if we cannot guarantee a consistent result.

flit:
Thanks for the additional info. That clears up my misconceptions!

> The VID/PID for the E1 is 0x1FC9/0x0132 -> NXP / something (I found only one reference to this combiantion, for a different board).

I'm still totally confused by how your patch works given that the VID/PID don't match DAPLink's 0x0d28/0x0204. (Like 141 in `cmsis_dap_probe.py` in the comment above.) Oh well. 🤷🏽 

> For some reason, I never found the github repo while googling for DAPLINK, so I naïvely assumed it was another name for the CMSIS-DAP!

Don't feel bad, the names definitely get confusing.

Since it's unlikely to reliable, I will close this PR. And unfortunately the name-matching approach I suggest won't work in this case since the actual USB strings don't include the "OKDO E1" bit; I misunderstood your log which showed the board name from your patch.

newbrain:
And...as a cherry on top I managed to brick the OKDO E1 probe: it does not come up any longer as a drive  when holding reset during power on. 😒
So, I soldered a couple of wires to SWCLK and SWDIO, and I'm going to try to directly flash a DAPLINK image on it. 🤞

flit:
That's strange. Good luck! If you erase flash it should boot into the ROM MSD bootloader as if you held reset on power-up.

Issue #1050, opened 2021-01-16
Test test_no_timeout will always fail on Windows


newbrain:
On Windows, the monotonic clock used for sleep() by Python platform has about 15 ms resolution, as described in [PEP564](https://www.python.org/dev/peps/pep-0564/#id23).

The test below is bound to fail as each iteration will consume at least 15 ms in sleep(0.01), for a total of at least 60 ms.
The test can pass if a reasonably higher than 50 ms timeout is used.

https://github.com/pyocd/pyOCD/blob/b0ab779c3e347a29adf1f5e9f8b2d676c3c9cf05/test/unit/test_timeout.py#L23-L33

As demonstrated by the following simple script, a possible value should probably not be much lower than 100 ms.
```
(venv3) ~\Git\pyOCD [master ≡ +0 ~1 -0 !]> python.exe
Python 3.7.9 (tags/v3.7.9:13c94747c7, Aug 17 2020, 16:30:00) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from time import time
>>> from time import sleep
>>> Max = 0
>>> for i in range(10000):
...     a=time()
...     sleep(0.01); sleep(0.01); sleep(0.01); sleep(0.01)
...     a= time()-a
...     if a > Max:
...             Max = a
...
>>> Max 
0.09961342811584473
```

flit:
Thanks! I'll try to get a fix in soon. Actually, I've been wanted to rewrite the test to mock `sleep()` so it doesn't actually sleep.

newbrain:
Success!
```
test\unit\test_timeout.py ....                                                                                         [100%]

===================================================== warnings summary ====================================================== 
venv\lib\site-packages\pyreadline\py3k_compat.py:8
  C:\Users\feder\Git\pyOCD\venv\lib\site-packages\pyreadline\py3k_compat.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working        
    return isinstance(x, collections.Callable)

-- Docs: https://docs.pytest.org/en/stable/warnings.html

----------- coverage: platform win32, python 3.9.1-final-0 -----------
Coverage HTML written to dir htmlcov

============================================== 859 passed, 1 warning in 20.73s ============================================== 
(venv) ~\Git\pyOCD [(v0.30.0)]> 
```
[...]

PR #1064, opened 2021-01-24
Support for ThreadX/AzureRTOS

newbrain:
Here is, as promised (threatened?) the PR for a ThreadX/AzureRTOS plug-in.

### General info
- The plug-in is named threadx.
- The code is, as one can expect, heavily modelled on the existing RTOS plug-ins.
- It has been autopep8'ed
- I'm a C programmer, mostly, this might show here and there.
- The plug-in has been tested with a NUCLEO-STM32F411RE and an OKDO-E1 (LPC55S69) and some test programs, using also hard FP but not including use of the secure features on LPC55S69.
- Tests have been done both with command line debugging and from inside VS Code.
- The stack frames look correct (registers and variables as expected), but they differ from ThreadX docs, see azure-rtos/threadx#70
- I have noticed that if I leave pyocd gdbserver running, quit gdb and then start a new gdb session, the plug-in is correctly loaded but threads are disabled in the new session until I issue a monitor thread enable. After that, the expected behaviour resumes.

### Running the internal test suite
- All unit tests on Windows 10 Pro 20H2 and Linux Mint 20.1 passes, with the exception of #1050 on Windows.
- Automatic tests always fail on Windows, with all the following boards:
  - NUCLEO STM32F411RE 
  - NUCLEO STM32F401RE 
  - NUCLEO STM32L476RG 
  - OKDO E1
  - Some others I've forgotten...
  - NXP MIMXRT1020-EVK
- The same happens on Linux, with the exception of the NUCLEO STM32L476RG, which passes all test.
- The failing tests (both systems) are  usually "Gdb Test" and "Flash Loader Test". Not 100% repeatable.
- MIMXRT1020-EVK behaves differently, see below:
```
------ TEST RESULTS ------
Target         Test                 Result         Time

mimxrt1020     Basic Test           Fail           2.302
mimxrt1020     Json Lists Test      Pass           10.825
mimxrt1020     Connect Test         Pass           9.743
mimxrt1020     Speed Test           Fail           2.163
mimxrt1020     Cortex Test          Fail           12.349
mimxrt1020     Concurrency Test     Fail           2.170
mimxrt1020     Flash Test           Pass           2718.288
mimxrt1020     Flash Loader Test    Pass           18.418
mimxrt1020     Debug Context Test   Pass           1.898
mimxrt1020     Gdb Test             Pass           69.294
mimxrt1020     Commands Test        Pass           8.540
mimxrt1020     Commander Test       Pass           0.852
```

This should not have anything to do with the plug-in itself, though, as a clean master version behaves the same.

### A possible improvement
To generate the threads' list, the plug-in walks the (doubly) linked list of created threads (so it's able to report also completed and terminated threads - of course not deleted ones).
Walking this list, and reading name, state and priority attributes of the thread relies on offsets inside the TX_THREAD structure.
Unfortunately, if the user has defined some thread extensions (using TX_THREAD_EXTENSION_0..3) these offsets will be wrong.

https://github.com/azure-rtos/threadx/blob/32e3b3b25f4b17ff51ce8a6bad8d9eb32729a0f6/common/inc/tx_api.h#L410-L418

To have some degree of confidence, the plug-in checks that each thread structure reached has the right id ('THRD' chars at offset 0)  and that the total number of threads matches the created threads count (yes, there's a small window in thread creation and deletion  where this will fail, maybe a + or - 1 tolerance should be added).

The ideal solution would be to read the offsets from the .elf, I don't know whether that is possible.
A thread is not guaranteed to exist, as ThreadX uses no Idle thread and the timer thread can be done away with and tucked in an ISR.

### DIFF
--- a/pyocd/rtos/threadx.py
+++ b/pyocd/rtos/threadx.py
@@ -14,8 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from enum import Enum
-
 from .provider import (TargetThread, ThreadProvider)
 from .common import (read_c_string, HandlerModeThread,
                      EXC_RETURN_EXT_FRAME_MASK)
@@ -31,7 +29,7 @@ import logging
 TX_THREAD_ID = 0x54485244  # 'THRD'
 THREAD_ID_OFFSET = 0
 THREAD_STACK_POINTER_OFFSET = 8
-# All the following offset may be meesed up if thread extensions are defined.
+# All the following offset may be messed up if thread extensions are defined.
 # They should be read somehow from the elf.
 THREAD_NAME_OFFSET = 40
 THREAD_PRIORITY_OFFSET = 44
@@ -177,9 +175,14 @@ class ThreadXThreadContext(DebugContext):
         super(ThreadXThreadContext, self).__init__(parent)
         self._thread = thread
         self._has_fpu = self.core.has_fpu
-        self.nofpu_register_offsets = self.NOFPU_REGISTER_OFFSETS
-        if self.core.architecture == CoreArchitecture.ARMv6M:
-            self.nofpu_register_offsets.update(self.NOFPU_REGISTER_OFFSETS_V6M)
+        if self.core.architecture != CoreArchitecture.ARMv6M:
+            # Use the default offsets for this istance
+            self._nofpu_register_offsets = self.NOFPU_REGISTER_OFFSETS
+        else:
+            # Use a copy with the Cortex-M0 specific offsets for this istance
+            self._nofpu_register_offsets = self.NOFPU_REGISTER_OFFSETS.copy()
+            self._nofpu_register_offsets.update(
+                self.NOFPU_REGISTER_OFFSETS_V6M)
 
     def read_core_registers_raw(self, reg_list):
         reg_list = [index_for_reg(reg) for reg in reg_list]
@@ -205,7 +208,7 @@ class ThreadXThreadContext(DebugContext):
         # Determine which register offset table to use and the offsets past the saved state.
         hwStacked = 0x20
         swStacked = 0x24
-        table = self.NOFPU_REGISTER_OFFSETS
+        table = self._nofpu_register_offsets
         if self._has_fpu:
             try:
                 if inException and self.core.is_vector_catch():
@@ -257,23 +260,28 @@ class ThreadXThreadContext(DebugContext):
 class ThreadXThread(TargetThread):
     """! @brief A ThreadX task."""
 
-    class ThreadState(Enum):
-        Ready = 0
-        Completed = 1
-        Terminated = 2
-        Suspended = 3
-        Sleep = 4
-        Queue = 5
-        Semaphore = 6
-        EventFlag = 7
-        BlockMemory = 8
-        ByteMemory = 9
-        IoDriver = 10
-        File = 11
-        TcpIp = 12
-        Mutex = 13
-        PriorityChange = 14
-        Unknown = 99
+    STATE_NAMES = {
+        0: "Ready",
+        1: "Completed",
+        2: "Terminated",
+        3: "Suspended",
+        4: "Sleep",
+        5: "Queue",
+        6: "Semaphore",
+        7: "EventFlag",
+        8: "BlockMemory",
+        9: "ByteMemory",
+        10: "IoDriver",
+        11: "File",
+        12: "TcpIp",
+        13: "Mutex",
+        14: "PriorityChange",
+        99: "Unknown"
+    }
+
+    READY = 0
+    PRIORITYCHANGE = 14
+    UNKNOWN = 99
 
     def __init__(self, targetContext, provider, base):
         super(ThreadXThread, self).__init__()
@@ -307,8 +315,8 @@ class ThreadXThread(TargetThread):
                 self._base + THREAD_PRIORITY_OFFSET)
             self._state = self._target_context.read32(
                 self._base + THREAD_STATE_OFFSET)
-            if not self.ThreadState.Ready.value <= self._state <= self.ThreadState.PriorityChange.value:
-                self._state = self.ThreadState.Unknown.value
+            if not self.READY <= self._state <= self.PRIORITYCHANGE:
+                self._state = self.UNKNOWN
         except exceptions.TransferError:
             LOG.debug("Transfer error while reading thread info")
 
@@ -335,7 +343,7 @@ class ThreadXThread(TargetThread):
     @property
     def description(self):
         # return "%s; Priority %d" % (self.STATE_NAMES[self.state], self.priority)
-        return "%s; Priority %d" % (self.ThreadState(self.state).name, self.priority)
+        return "%s; Priority %d" % (self.STATE_NAMES[self.state], self.priority)
 
     @property
     def is_current(self):
@@ -353,18 +361,17 @@ class ThreadXThread(TargetThread):
 
 
 class ThreadXThreadProvider(ThreadProvider):
-    """! @brief Thread provider for ThreadX."""
+    """! @brief Thread provider for ThreadX.
+
+        To successfully initialize, the following ThreadX symbols are needed:
+            _tx_thread_created_ptr:     pointer to list of created processes
+            _tx_thread_created_count:   count of created processes
+            _tx_thread_current_ptr:     current thread
+            _tx_thread_system_state:    ThreadX state: initializing, run, interrupt
+    """
 
     # Scheduler not yet up
     TX_INITIALIZE_IN_PROGRESS = 0xF0F0F0F0
-
-    '''
-    Required ThreadX symbols:
-        _tx_thread_created_ptr:     pointer to list of created processes
-        _tx_thread_created_count:   count of created processes
-        _tx_thread_current_ptr:     current thread
-        _tx_thread_system_state:    ThreadX state: initializing, run, interrupt
-    '''
 
     def __init__(self, target):
         super(ThreadXThreadProvider, self).__init__(target)
@@ -441,8 +448,8 @@ class ThreadXThreadProvider(ThreadProvider):
 
         # Is the number of threads correct?
         if len(newThreads) != threadCount:
-            LOG.warning("ThreadX: thread count mismatch, expected: {0}, found: {1}".format(
-                threadCount, len(newThreads)))
+            LOG.warning("ThreadX: thread count mismatch, %d expected, %d found",
+                        threadCount, len(newThreads))
 
         # Create fake handler mode thread.
         if self._target_context.read_core_register('ipsr') > 0:
@@ -466,9 +473,15 @@ class ThreadXThreadProvider(ThreadProvider):
 
     @property
     def is_enabled(self):
+        # The _tx_thread_system_state global is used to determine whether
+        # the kernel is running. Before the kernel starts, it'll contain
+        # TX_INITIALIZE_IN_PROGRESS and possibly TX_INITIALIZE_IN_PROGRESS+1.
+        # On cortex-m ports it should otherwise be 0.
+        # As it's used in other ports to indicate the interrupt nesting level, it's
+        # safer to compare it with TX_INITIALIZE_IN_PROGRESS.
         if self._system_state is None:
             return False
-        return self._target_context.read32(self._system_state) < ThreadXThreadProvider.TX_INITIALIZE_IN_PROGRESS
+        return self._target_context.read32(self._system_state) < self.TX_INITIALIZE_IN_PROGRESS
 
     @property
     def current_thread(self):
@@ -498,6 +511,7 @@ class ThreadXThreadProvider(ThreadProvider):
         if not self.is_enabled:
             return None
         return self._target_context.read32(self._current_ptr)
+
 
 class ThreadXPlugin(Plugin):
     """! @brief Plugin class for ThreadX."""
--- a/pyocd/rtos/threadx.py
+++ b/pyocd/rtos/threadx.py
@@ -1,0 +1,516 @@
+# pyOCD debugger
+# Copyright (c) 2020-2021 Federico Zuccardi Merli
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+
+from .provider import (TargetThread, ThreadProvider)
+from .common import (read_c_string, HandlerModeThread,
+                     EXC_RETURN_EXT_FRAME_MASK)
+from ..core import exceptions
+from ..core.target import Target
+from ..core.plugin import Plugin
+from ..debug.context import DebugContext
+from ..coresight.cortex_m_core_registers import index_for_reg
+from ..coresight.core_ids import CoreArchitecture
+
+import logging
+
+TX_THREAD_ID = 0x54485244  # 'THRD'
+THREAD_ID_OFFSET = 0
+THREAD_STACK_POINTER_OFFSET = 8
+# All the following offset may be meesed up if thread extensions are defined.
+# They should be read somehow from the elf.
+THREAD_NAME_OFFSET = 40
+THREAD_PRIORITY_OFFSET = 44
+THREAD_STATE_OFFSET = 48
+THREAD_NEXT_OFFSET = 136
+
+# Create a logger for this module.
+LOG = logging.getLogger(__name__)
+
+
+class TargetList(object):
+    def __init__(self, context, ptr):
+        self._context = context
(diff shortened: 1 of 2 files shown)


flit:
> I have noticed that if I leave pyocd gdbserver running, quit gdb and then start a new gdb session, the plug-in is correctly loaded but threads are disabled in the new session until I issue a monitor thread enable. After that, the expected behaviour resumes.

The gdbserver includes some logic to disable reporting threads until the first resume after a connect or reset. This is to work around the problem of stale thread data after you load updated code but RAM still contains partially valid RTOS data structures. `monitor thread enable` simply overrides this logic. (Jeez I need to document all this…)

> Automatic tests always fail on Windows

I'm not too worried. Running the functional tests on STM32 and i.MX RT boards has always been problematic (though will have to be addressed in the new CI system!) And of course the functional tests don't test RTOS plugins (yet).

> A thread is not guaranteed to exist…

Is a "live" fallback thread created? Or put another way, what does the user see when there is no thread and they halt in gdb?

newbrain:
Thanks for the explanation, it makes sense.

> Is a "live" fallback thread created? Or put another way, what does the user see when there is no thread and they halt in gdb?

For the case we are still in initialization phase:
ThreadX has an internal variable (_tx_thread_system_state) that tells you whether:
* It's initializing (in two steps).
* It's in the ready state (0).
* On some ports, if It's inside a ThreadX interrupt handler (for cortex-m it's always 0).

In the plug-in, this is checked by the is_enabled @property in ThreadXThreadProvider.
In gdb, it appears as a single non-RTOS thread.
When inside an exception handler, an extra (fictitious) thread is reported.

Not having any threads is a bit of a corner case, but it is indeed possible:
* Have the timer task served in IRS (it's compile time #define) 
* Do not create any thread, and register a timeout during the initialization phase, with threads creation in its callback function.
* Until the timer expires, there'll be 0 threads, the code will spin in a "wait here" loop.

And you probably got me here - I did not test this scenario so I cannot answer.
I will do ASAP (need to put together a test program), and in case add a check for 0 threads with a default action.

flit:
Since it takes a bit of an unrealistic scenario to have no threads, I'm not really worried about it at this point. Thanks for the explanation of the required sequence.

What about when there are no running or ready threads? If I remember correctly, ThreadX stays in a loop in the scheduler until a thread becomes ready. But it's been _years_ since I used ThreadX, so I guess it may have changed.

newbrain:
Oh, for me it's a new thing: I noticed Microsoft acquired it, so I was curious.
I don't dislike it, the naming is less confusing than FreeRTOS, it does not have FreeRTOS's limitation of special APIs for IRS (and the one on IRQ priority).

A thread in ThreadX can be in a number of states, they will be reported in gdb along the other thread info.
See below:
```
(gdb) tb main
Temporary breakpoint 1 at 0x800041c: file Src/sample_threadx.c, line 62.
(gdb) c
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.

Temporary breakpoint 1, main () at Src/sample_threadx.c:62
62          tx_kernel_enter();
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 1          main () at Src/sample_threadx.c:62
(gdb) c
Continuing.
[New Thread 536871640]
[New Thread 536882300]
[New Thread 536881408]
[New Thread 536871044]
[New Thread 536872108]
[New Thread 536871456]
[New Thread 536881748]
[New Thread 536871820]
[New Thread 536871264]

Thread 2 "thread 1" received signal SIGINT, Interrupt.
[Switching to Thread 536871640]
0x08003b48 in _tx_queue_send (queue_ptr=0x200029b4 <queue_0>, source_ptr=0x200029f0 <thread_1_messages_sent>,
    wait_option=4294967295) at ThreadX/common/src/tx_queue_send.c:147
147                 TX_QUEUE_MESSAGE_COPY(source, destination, size)
(gdb) info threads
  Id   Target Id                                                      Frame
* 2    Thread 536871640 "thread 1" (Ready; Priority 16)               0x08003b48 in _tx_queue_send (
    queue_ptr=0x200029b4 <queue_0>, source_ptr=0x200029f0 <thread_1_messages_sent>, wait_option=4294967295)
    at ThreadX/common/src/tx_queue_send.c:147
  3    Thread 536882300 "System Timer Thread" (Suspended; Priority 0) 0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  4    Thread 536881408 "thread 0" (Sleep; Priority 1)                0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  5    Thread 536871044 "thread 2" (Ready; Priority 16)               0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  6    Thread 536872108 "thread 3" (Semaphore; Priority 8)            0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  7    Thread 536871456 "thread 4" (Sleep; Priority 8)                0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  8    Thread 536881748 "thread 5" (EventFlag; Priority 4)            0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  9    Thread 536871820 "thread 6" (Mutex; Priority 8)                0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  10   Thread 536871264 "thread 7" (Sleep; Priority 8)                0x08004a12 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
```
So, "thread 1" (2) was the current one, interrupted and is in _Ready_ state, the "System Timer Thread" is _Suspended_ (waiting for a service) and others are waiting for _EventFlags_, _Sleep_ing etc.
After all, with only one core, only one thread is really executing at any time.

As a quick test for the zero threads issue:
I patched the plug-in to always give an empty thread list, and the result is that the current position is shown, it is possible to step etc. BUT **warning: Remote failure reply: E01** is reported:
```
(gdb) c
Continuing.
warning: Remote failure reply: E01

Program stopped.
0x080007e0 in thread_2_entry (thread_input=2) at Src/sample_threadx.c:248
248             if ((status != TX_SUCCESS) || (received_message != thread_2_messages_received))
(gdb) bt
#0  0x080007e0 in thread_2_entry (thread_input=2) at Src/sample_threadx.c:248
#1  0x08004392 in _tx_thread_shell_entry () at ThreadX/common/src/tx_thread_shell_entry.c:114
#2  0xfffffffe in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
[...]
(gdb) n
warning: Remote failure reply: E01

Program stopped.
_txe_queue_receive (queue_ptr=0x200029b4 <queue_0>, destination_ptr=0x20001138 <memory_area+3032>,
    wait_option=4294967295) at ThreadX/common/src/txe_queue_receive.c:102
102         else if (queue_ptr -> tx_queue_id != TX_QUEUE_ID)
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 1          _txe_queue_receive (queue_ptr=0x200029b4 <queue_0>,
    destination_ptr=0x20001138 <memory_area+3032>, wait_option=4294967295)
    at ThreadX/common/src/txe_queue_receive.c:102
```

flit:
I used ThreadX extensively for nearly a decade, though never on Cortex-M, and it's one of my favourite RTOSes. In some ways, I based my [Argon](/flit/argos-rtos) RTOS on it. I was happy to see Microsoft buying and open sourcing it.

The thread status reporting looks great, just what I would expect.

Really, don't worry about the no threads issue right now. You can add that in a follow-up PR when you get a chance. My recommendation is to focus instead on the case where there are no ready threads; this is happen in almost all real applications. (They may have the same or similar solution, though.)

newbrain:
Scratch the quick test part!
It seems to be much better than that: the reported warning was due to using _false_ instead of _False_ (eh, I warned I'm a C programmer...).

The result seems to be quite OK:
```
(gdb) mon reset halt
Resetting target with halt
Successfully halted device on reset
(gdb) tb main
Temporary breakpoint 2 at 0x800041c: file Src/sample_threadx.c, line 62.
(gdb) c
Continuing.

Temporary breakpoint 2, main () at Src/sample_threadx.c:62
62          tx_kernel_enter();
(gdb) c
Continuing.

Program received signal SIGINT, Interrupt.
0x080013dc in _txe_queue_receive (queue_ptr=0x200029b4 <queue_0>, destination_ptr=0x20001138 <memory_area+3032>,
    wait_option=4294967295) at ThreadX/common/src/txe_queue_receive.c:140
140                     if (current_thread == &_tx_timer_thread)
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 1          0x080013dc in _txe_queue_receive (queue_ptr=0x200029b4 <queue_0>,
    destination_ptr=0x20001138 <memory_area+3032>, wait_option=4294967295)
    at ThreadX/common/src/txe_queue_receive.c:140
(gdb) n
152         if (status == TX_SUCCESS)
(gdb) n
156             status =  _tx_queue_receive(queue_ptr, destination_ptr, wait_option);
(gdb) n
160         return(status);
(gdb) n
161     }
(gdb) n
thread_2_entry (thread_input=2) at Src/sample_threadx.c:248
248             if ((status != TX_SUCCESS) || (received_message != thread_2_messages_received))
(gdb) n
252             thread_2_messages_received++;
(gdb) n
241             thread_2_counter++;
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 1          thread_2_entry (thread_input=2) at Src/sample_threadx.c:241
(gdb) bt
#0  thread_2_entry (thread_input=2) at Src/sample_threadx.c:241
#1  0x08004392 in _tx_thread_shell_entry () at ThreadX/common/src/tx_thread_shell_entry.c:114
#2  0xfffffffe in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```
But, to make sure, I still will set up a proper test.

newbrain:
The case when no task is Ready is handled by the logic in  [ThreadXThreadContext::read_core_registers_raw](https://github.com/newbrain/pyOCD/blob/f4cb0c8410bf340521b05c6ba026499644891717/pyocd/rtos/threadx.py#L184-L254).
If the thread is not current, the stacked frame will be read, otherwise the live registers.
As this is done thread by thread, there is no need for a special case when all threads are suspended/waiting.

flit:
Perfect! 👍🏽

newbrain:
As a proof, I put all threads to sleep (it's almost midnight here, and they are quite young):
```
 (gdb) info threads
  Id   Target Id                                                      Frame
* 2    Thread 2 "Handler mode" (PendSV)                               __tx_ts_wait ()
    at ThreadX/ports/cortex_m4/gnu/src/tx_thread_schedule.S:251
  3    Thread 536882300 "System Timer Thread" (Suspended; Priority 0) 0x08004916 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  4    Thread 536881408 "thread 0" (Sleep; Priority 1)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  5    Thread 536871640 "thread 1" (Sleep; Priority 16)               0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  6    Thread 536871044 "thread 2" (Sleep; Priority 16)               0x08004916 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  7    Thread 536872108 "thread 3" (Sleep; Priority 8)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  8    Thread 536871456 "thread 4" (Sleep; Priority 8)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  9    Thread 536881748 "thread 5" (Sleep; Priority 4)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  10   Thread 536871820 "thread 6" (Sleep; Priority 8)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
  11   Thread 536871264 "thread 7" (Sleep; Priority 8)                0x08004a42 in __get_ipsr_value ()
    at ThreadX/ports/cortex_m4/gnu/inc/tx_port.h:371
```

Note the the idle loop in ThreadX is, as you said, inside the scheduler (PendSV ISR), i.e. in the "Handler Mode" thread.
Were it outside, some special handling might have been needed.

newbrain:
After some documentation perusing and test code abusing, the corner case I hypothesized above (scheduler started but zero threads) is not possible: in fact, threads and other objects _cannot_ be created from a timer callback.
If error checking is enabled during compilation, TX_CALLER_ERROR is returned by the likes of tx_*primitive*_create()
So, not something we should worry about.

flit:
All very good, thanks for double checking.

pyocd/rtos/threadx.py
```diff
@@ -0,0 +1,514 @@
...
+        if not self.is_enabled:
+            return None
+        self.update_threads()
+        return self._threads.get(threadId, None)
+
+    @property
+    def is_enabled(self):
+        if self._system_state is None:
+            return False
+        return self._target_context.read32(self._system_state) < ThreadXThreadProvider.TX_INITIALIZE_IN_PROGRESS
```
flit, line 471: Is a less-than comparison correct? Just odd to see < a pattern like 0xF0F0F0F0. If it's correct, maybe add comment on how the comparison works?

pyocd/rtos/threadx.py
```diff
@@ -0,0 +1,514 @@
...
+                    t = ThreadXThread(self._target_context, self, threadBase)
+                LOG.debug("Thread 0x%08x (%s)", threadBase, t.name)
+                newThreads[t.unique_id] = t
+            except exceptions.TransferError:
+                LOG.debug(
+                    "TransferError while examining thread 0x%08x", threadBase)
+
+        # Is the number of threads correct?
+        if len(newThreads) != threadCount:
+            LOG.warning("ThreadX: thread count mismatch, expected: {0}, found: {1}".format(
```
flit, line 444: Use `%` formatting instead of `format()` method, like the other log calls. The logger stores the string formatting arguments in the log message object and only formats the string if it gets printed.

setup.py
```diff
@@ -123,6 +123,7 @@
             'freertos = pyocd.rtos.freertos:FreeRTOSPlugin',
             'zephyr = pyocd.rtos.zephyr:ZephyrPlugin',
```
flit, line 124: Would you mind moving `zephyr` to the end of this list? I should have kept it in sorted order… Thanks.

flit (changes requested):
Looks really good. Just a few small things to change and it's ready to merge.

pyocd/rtos/threadx.py
```diff
@@ -0,0 +1,514 @@
...
+        0x4f:   192,    # s15
+        33:     196,    # fpscr
+        # (reserved word: 200)
+    }
+
+    def __init__(self, parent, thread):
+        super(ThreadXThreadContext, self).__init__(parent)
+        self._thread = thread
+        self._has_fpu = self.core.has_fpu
+        self.nofpu_register_offsets = self.NOFPU_REGISTER_OFFSETS
```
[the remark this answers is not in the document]
newbrain: Good catch, thanks - the case you make is not really far fetched, as e.g. NXP LPC43S37 has two M0 (+?) and one M4 and if a soft abi is used for FP, it will appear just as you envisioned.

pyocd/rtos/threadx.py
```diff
@@ -0,0 +1,514 @@
...
+        Semaphore = 6
+        EventFlag = 7
+        BlockMemory = 8
+        ByteMemory = 9
+        IoDriver = 10
+        File = 11
+        TcpIp = 12
+        Mutex = 13
+        PriorityChange = 14
+        Unknown = 99
```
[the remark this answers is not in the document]
newbrain: I see, on the other hand I was using them also as the printed out name for the thread state (via enum.name).
I will revert to a dictionary (my first implementation, not pushed), if this is against the rules.

pyocd/rtos/threadx.py
newbrain: I will add a comment, the system state takes TX_INITIALIZE_IN_PROGRESS and TX_INITIALIZE_IN_PROGRESS+1 during initialization, and 0 (for cortex-m ports). In other ports, it contains interrupt nesting count.

pyocd/rtos/threadx.py
newbrain: Agreed. Did not know it would make a difference.

setup.py
newbrain: Of course! In fact, in my first version I did just that, then thought of not introducing spurious changes.

newbrain:
Thanks for the insightful comments, I'll work on this tonight - hope to have it ready soon.

newbrain:
I don't think I need to rebase on master, as the PR can be merged cleanly.
Let me know in case it's preferred.

flit (approved):
Perfect!

newbrain:
Thanks for the great support!

flit:
Thanks for the contribution! 😄

PR #1070, opened 2021-02-03
add NXP LPC845 target

Hoohaha:
Tested in my local with lpc845breakout, works well.
But i do not how to get the board id, so do not add board info.


flit:
Hi @Hoohaha 😄 

It looks like the on-board CMSIS-DAP on the LPC845-BRK is the generic NXP CMSIS-DAP firmware. Since this firmware is generic, it doesn't report any ID that can be used to recognize the board.

Do you happen to have any idea what might be causing #1056?

PR #1080, opened 2021-02-16
Dependency updates

flit:
- Move capstone from extras_require to install_requires, since installing works on all platforms now, even Raspberry Pi.
- Drop pywinusb in favour of hidapi on Windows.
- Update version requirements for several packages.
- Add version constraints for all packages without one, even for packages with only pre-1.0 versions.
- Add pyocd_pemicro probe plugin.
- Removed long-since deprecated `pyocd-flash` and `pyocd-tool` console executables. I'm keeping `pyocd-gdbserver`, primarily because the cortex-debug plugin for VSCode still relies on it. (Eclipse Embedded CDT née GNU MCU Eclipse uses the combined `pyocd` tool since the recent version 6.1.)


newbrain:
Check the comment from @marus on [pyocd-gdbserver](https://github.com/Marus/cortex-debug/issues/322#issuecomment-783003871), maybe with a little synchronization this can move forward!

flit:
@newbrain Thanks for the pointer to the cortex-debug issue!

PR #1082, opened 2021-02-20
Picoprobe support in pyOCD

newbrain:
Here is a PR to support the ultra simple [Picoprobe](https://github.com/raspberrypi/picoprobe) in pyOCD.
The adapter receives commands to write or read sequence of bits on the SWD interface, and has no intelligence.
Requests are sent and received using bulk EPs in a vendor specific HID interface.

- Supports HW reset in Picoprobe (if available), see this [PR in Picoprobe](https://github.com/raspberrypi/picoprobe/pull/10)
- Safe (slow) or bulk (faster) SWD modes.
In bulk mode, commands are sent in blocks of max. 256, and ACKs and parities are checked in bulk at the end of the transaction.
In safe mode, each SWD command checks for ACK & parity.
Safe mode is significantly slower as there's an unavoidable ~1 ms turnaround time for each phase of an SWD command (cmd, ACK, data).
A new boolean option is provided `picoprobe.safeswd`, defaulting to False.
The size of the block has been chosen to fit Picoprobe's internal 8 kB read and write buffers, so that a single EP read or write is needed
- Tested with a Discovery F429 and a Nucleo L476, using:
Command line, commander mode, VS Code, and Visual Studio Community + VisualGDB.
Flashing, erasing, breakpoints, reading and writing RAM memory and registers.
Windows10 Pro 20H2, libusb0 (v1.2.6.0).
I need to test this on Linux.
- Probably not working in python 2.x, not tested and not written with that in mind.

Questions:
- Is it needed to write the DP SELECT register in `read_ap()` and `write_ap()`?
Or does the upper layer always calls `write_dp()` beforehand?
- Is there the need to do something special to support ADI versions later than 5?
- Am I right in thinking that python 2 support is not needed?
- Is it really worth having this in pyOCD (honest question)?
I agree with @flit that porting DAPlink on the Pico board migth be a better option. (maybe I'll do that next, now that I have a slightly better grasp of how al this works).
An advantage of having the support in pyOCD is that there's no need to compile the special version of openocd, which brings in yet another toolchain, and is not so painless.

### DIFF
--- a/pyocd/probe/picoprobe.py
+++ b/pyocd/probe/picoprobe.py
@@ -274,7 +274,9 @@ class Picoprobe(DebugProbe):
 
     @ classmethod
     def get_probe_with_id(cls, unique_id, is_explicit=False):
-        pass
+        for dev in PicoLink.enumerate_picoprobes():
+            if dev.get_unique_id() == unique_id:
+                return cls(dev)
 
     def __init__(self, picolink):
         super(Picoprobe, self).__init__()
--- a/pyocd/probe/picoprobe.py
+++ b/pyocd/probe/picoprobe.py
@@ -1,0 +1,634 @@
+# pyOCD debugger
+# Copyright (c) 2021 Federico Zuccardi Merli
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from array import array
+
+from time import sleep
+from usb import core, util
+
+from .debug_probe import DebugProbe
+from ..core import exceptions
+from ..core.options import OptionInfo
+from ..core.plugin import Plugin
+
+
+class PicoLink(object):
+    """! @brief Wrapper to handle picoprobe USB.
+
+    Just to hide details of USB and Picoprobe command layer
+    """
+
+    VID = 0x2E8A
+    PID = 0x0004
+
+    CLASS = 0xFF    # Vendor Specific
+
+    CMD_HDR_LEN = 6  # do not include pico packet header
+    PKT_HDR_LEN = 4  # pico packet header
+    HDR_LEN = PKT_HDR_LEN + CMD_HDR_LEN
+
+    PROBE_INVALID = 0       # Invalid command
+    PROBE_WRITE_BITS = 1    # Host wants us to write bits
+    PROBE_READ_BITS = 2     # Host wants us to read bits
+    PROBE_SET_FREQ = 3      # Set TCK
+    PROBE_RESET = 4         # Reset all state: it's a no-op!
+    PROBE_TARGET_RESET = 5  # Reset target (Hardware nreset)
+
+    BUFFER_SIZE = 8192      # Size of buffers in the picoprobe
+
+    def __init__(self, dev):
+        self._dev = dev
+        self._probe_id = dev.serial_number
+        self._vend = dev.manufacturer
+        self._prod = dev.product
+        # USB intefrface and endpoints
+        self._if = None
+        self._wr_ep = None
+        self._rd_ep = None
+        # Progressive command id
+        self._id = 0
+        # Probe command queue
+        self._queue = array('B', (0, 0, 0, 0))
+        self._qulen = self.PKT_HDR_LEN
+        # Buffer for endpoint reads
+        self._bits = array('B', (0 for _ in range(self.BUFFER_SIZE)))
+
+    # ------------------------------------------- #
+    #          Picoprobe Access functions
+    # ------------------------------------------- #
+    def open(self):
+        # Only one configuration considered
+        self._dev.set_configuration()
+        # Search the Vendor Specific interface in first configuration
+        for i in self._dev[0]:
+            if i.bInterfaceClass == PicoLink.CLASS:
+                self._if = i
+                break
+        # Check for a missing device interface
+        if self._if is None:
+            raise exceptions.ProbeError()
+        # Scan and assign Endpoints
+        for e in self._if:
+            if util.endpoint_direction(e.bEndpointAddress) == util.ENDPOINT_OUT:
+                self._wr_ep = e
+            else:
+                self._rd_ep = e
+        # Something is missing from this probe!
+        if self._wr_ep is None or self._rd_ep is None:
+            raise exceptions.ProbeError("Unrecognized Picoprobe interface")
+
+    def close(self):
+        self._if = None
+        self._wr_ep = None
+        self._rd_ep = None
+
+    @classmethod
+    def enumerate_picoprobes(cls):
+        """! @brief Find and return all Picoprobes """
+        return [PicoLink(probe) for probe in core.find(idVendor=PicoLink.VID, idProduct=PicoLink.PID, find_all=True)]
+
+    def q_read_bits(self, bits):
+        """! @brief Queue a read request for 'bits' bits to the probe """
+        # Cannot be called with bits = 0
+        self._queue_cmd_header(self.PROBE_READ_BITS, bits)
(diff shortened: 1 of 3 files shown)


flit:
> Questions:
> 
> * Is it needed to write the DP SELECT register in `read_ap()` and `write_ap()`?
>   Or does the upper layer always calls `write_dp()` beforehand?

This is controlled by the `MANAGED_AP_SELECTION` capability:
 https://github.com/pyocd/pyOCD/blob/7dccf5c5da825d36cb947a7b06c17f6a56852189/pyocd/probe/debug_probe.py#L53-L57

It's for probes like J-Link or STLink that take a full AP register address. If you don't report this capability, the `DebugProbe` class will manage `DP_SELECT` for you.

> * Is there the need to do something special to support ADI versions later than 5?

No for SWD.
If JTAG were supported then there is a small change. But pyocd doesn't fully support JTAG yet so there's nothing to do.

> * Am I right in thinking that python 2 support is not needed?

Technically pyocd still supports Python 2, in that CI runs on by 2 and 3, and `setup.py` indicates Py2 support. However, I'm almost certainly going to drop Py2 support in the next release, v0.30.0. Just need to update everything to reflect that.

> * Is it really worth having this in pyOCD (honest question)?
>   I agree with @flit that porting DAPlink on the Pico board migth be a better option. (maybe I'll do that next, now that I have a slightly better grasp of how al this works).
>   An advantage of having the support in pyOCD is that there's no need to compile the special version of openocd, which brings in yet another toolchain, and is not so painless.

Probably? I'm certainly not opposed to it and will be happy to merge it in.

You do have another option though. Pyocd supports probe plugins. So, if you preferred, you could release picoprobe support as a standalone plugin package on PyPI. This is how the PEMicro probe support is handled. It's contained in the [pyocd-pemicro](/pyocd/pyocd-pemicro) repo and released separately on PyPI, and will soon be added as a dependency for pyocd (there was a bug that prevented it from being added in v0.29). It gives the author more freedom since they own the repo. (And pyocd-pemicro was originally separated because it only supports Python 3, which is much easier to manage when a separate package.)

Totally up to you. I'll be happy to create a pyocd-org repo if you like.

pyocd/probe/picoprobe.py
```diff
@@ -0,0 +1,636 @@
...
+        """! @brief Only valid after connecting."""
+        return DebugProbe.Protocol.SWD if self._is_connected else None
+
+    @ property
+    def is_open(self):
+        return self._is_open
+
+    @ property
+    def capabilities(self):
+        return {DebugProbe.Capability.SWJ_SEQUENCE}
```
flit, line 321: Let's coordinate so you can add the `SWD_SEQUENCE` capability and `swd_sequence()` API that I've added in my RP2040 branch:

https://github.com/flit/pyOCD/blob/a9b1e3824d1c809ac27c57328e943507d6cbbd99/pyocd/probe/debug_probe.py#L231

This will enable using picoprobe to debug the RP2040.

pyocd/probe/picoprobe.py
```diff
@@ -0,0 +1,636 @@
...
+        # Use the bulk or safe read and write functions according to option
+        if self.session.options.get(self.SAFESWD_OPTION):
+            self.read_ap_multiple = self._safe_read_ap_multiple
+            self.write_ap_multiple = self._safe_write_ap_multiple
+        else:
+            self.read_ap_multiple = self._bulk_read_ap_multiple
+            self.write_ap_multiple = self._bulk_write_ap_multiple
+        # Subscribe to option change events
+        self.session.options.subscribe(self._change_options, [self.SAFESWD_OPTION])
+        # Do I need to do anything else here?
```
flit, line 355: No, nothing else is required.

For most probes, you need to send some command to switch them from Hi-Z to output. This is required so you can have multiple probes connected to a target without conflicts (as long as only one is active at a time, of course).

pyocd/probe/picoprobe.py
newbrain: SWD sequences:
I'll do that ASAP - probably, in less than 24h.
It seems to fit quite well with the primitives I have, so should be no great deal.
What is the max length expected?

Should I rebase the PR onto the RP2040 branch, then?
(So I get the new base class and capabilities)

As for the other points:
- SELECT: Now I understand what that capabilities meant, I was not sure there was something else to support.
I use a separate write (cached so it's probably always been skipped anyway!) to DP.SELECT, even for 'bulk' mode. I'll simply remove it. Less code, less bugs. J-link implementation, in fact, does exactly what I'm doing - well, the other way round, TBH.
- ADI: good, thanks.
- Python 2 and separate repo: If you are Ok with it, **and** support for python 2 is going to be removed, I'll just have this merged. If any of the two does not hold, I'll go for the separate repo.

Thanks for your support!

flit:
I'm finding a few bugs in the existing code as I'm reviewing your changes. 😅  I'll update everything in a moment.

pyocd/probe/picoprobe.py
flit: > What is the max length expected? 

Max length is 64 cycles per sequence. There isn't a max for the number of sequences, though I need to update the CMSIS-DAP implementation to actually support this (right now it's limited to what fits in a packet, which is fine because the only user of the API is the RP2040 target for the short target selection sequence).

> Should I rebase the PR onto the RP2040 branch, then?

Yup. The other option would be to wait until I merge in the branch, then rebase on `master`.

> SELECT

Well, I messed up. 😓  The J-Link implementation doesn't actually need to deal with DP_SELECT at all. I missed removing that code when I refactored the DP_SELECT support. It still works fine, but I'm in the process of removing it right now. (This is the bug I mentioned in another comment.)

I'll add some review comments just to be thorough.

> ADIv6

I was actually sort of wrong here. The `APv2_ADDRESSES` probe capability declares support for the wider AP addresses used with ADIv6. But it only applies if the probes manages AP selection, i.e., the `MANAGED_AP_SELECTION` capability (although I declare it anyway for `CMSISDAPProbe` and soon for `JLinkProbe`, for completeness).

So once you remove the management of DP_SELECT, ADIv6 will be supported.

> Python 2

Yes, I'm happy to merge it in. It'll force me to not be lazy about removing Python 2 support! 😄

pyocd/probe/picoprobe.py
```diff
@@ -0,0 +1,636 @@
...
+            return val
+        return val if now else read_dp_result_callback
+
+    def write_dp(self, addr, value):
+        # Check whether we need to write a new SELECT value
+        if addr == self.SELECT:
+            if value == self._select:
+                return
+            else:
+                self._select = value
```
flit, line 403: Reminder that you can remove `SELECT` management.

newbrain:
A first attempt at swd_sequence is [here](https://github.com/newbrain/pyOCD/commit/f2a4a85e577ecb8bdf565e77c06b4244a83a3755).
Not (yet) tested. Need to solder some pin headers on the second Pico...

pyocd/probe/picoprobe.py
```diff
@@ -0,0 +1,636 @@
...
+    def write_ap(self, addr, value):
+        self.write_ap_multiple(addr, (value,))
+
+    def _safe_read_ap_multiple(self, addr, count=1, now=True):
+        # Extract the APSEL and APBANKSEL fields
+        sel = addr & self.APSEL_APBANKSEL
+        # Write it to SELECT register on DP
+        # It would appear that a write to SELECT is always done first,
+        # so this is possibly not needed. Oh well, it's cached anyway.
+        self.write_dp(self.SELECT, sel)
```
flit, line 422: Reminder that you can remove `SELECT` management. Same for other AP methods, of course.

pyocd/probe/picoprobe.py
newbrain: > >  Should I rebase the PR onto the RP2040 branch, then?

> Yup. The other option would be to wait until I merge in the branch, then rebase on master.

Ok, no probs, and I've created a branch ([Picoprobe2040](https://github.com/newbrain/pyOCD/tree/PicoprobeRP2040)) from your RP2040 one, adding another remote locally to keep in synch. I've also added an implementation of swd_sequence, exercising my comprehension of list comprehension.

I'll also follow up the review here (on Picoprobe), and cherry pick the commit(s) with the review changes over PicoprobeRP2040 before making a PR there, so nothing should be lost, and you are free to point  me in whichever direction.

Once again, thanks for your great support.

flit:
Closing this PR. The changes have been merged into my [feature/rp2040](/flit/pyOCD/tree/feature/rp2040) branch, which will soon be merged upstream.

PR #1083, opened 2021-02-21
Raspberry Pi Pico (RP2040) support

flit:
This PR includes a number of changes and additions.

Major changes:
- RP2040 targets: `rp2040_core0`, `rp2040_core1`, `rp2040` (alias for core 0). Only one core can be accessed at a time for now, pending full multi-drop SWD support.
- [picoprobe](https://github.com/raspberrypi/picoprobe) support thanks to @newbrain.

Other changes:
- The CMSIS-DAP protocol version is read.
- Added `jtag_sequence()` debug probe API.
- Added `swd_sequence()` debug probe API.
- Refactored the `SWJSequenceSender` class to allow sending fragments of sequences and perform any single step of the SWD <—> dormant <—> JTAG transitions.
- Flash algo generation improvements, used to generate the RP2040 flash algo from  https://github.com/rp-rs/flash-algo.

### DIFF
--- a/pyocd/probe/picoprobe.py
+++ b/pyocd/probe/picoprobe.py
@@ -1,0 +1,636 @@
+# pyOCD debugger
+# Copyright (c) 2021 Federico Zuccardi Merli
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from array import array
+
+from time import sleep
+from usb import core, util
+
+from .debug_probe import DebugProbe
+from ..core import exceptions
+from ..core.options import OptionInfo
+from ..core.plugin import Plugin
+
+
+class PicoLink(object):
+    """! @brief Wrapper to handle picoprobe USB.
+
+    Just to hide details of USB and Picoprobe command layer
+    """
+
+    VID = 0x2E8A
+    PID = 0x0004
+
+    CLASS = 0xFF    # Vendor Specific
+
+    CMD_HDR_LEN = 6  # do not include pico packet header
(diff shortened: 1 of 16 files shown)


pyocd/probe/picoprobe.py
```diff
@@ -0,0 +1,633 @@
...
+        self._qulen = self.PKT_HDR_LEN
+        # Buffer for endpoint reads
+        self._bits = array('B', (0 for _ in range(self.BUFFER_SIZE)))
+
+    # ------------------------------------------- #
+    #          Picoprobe Access functions
+    # ------------------------------------------- #
+    def open(self):
+        # Only one configuration considered
+        self._dev.set_configuration()
```
newbrain, line 75: Please see flit#4

newbrain:
Ah, nice to see it all coming together.
I have unfortunately found a problem on Linux with Picoprobe, see my comment above.

udev/50-picoprobe.rules
```diff
@@ -0,0 +1,8 @@
+# 2e8a:0004 Raspberry Pi picoprobe
+# https://github.com/raspberrypi/picoprobe
+SUBSYSTEM=="usb", ATTR{idVendor}=="2e8a", ATTR{idProduct}=="0004", MODE:="666"
```
newbrain, line 3: My preference really goes to using a group, rather than the satanic MODE.
But as long as the safer alternatives are suggested (line 7-8) I'm fine with this, as it's simpler for naïve users.
This is more a general comment, as all the .rules files here follow the same pattern.

flit: Yeah, that makes sense. I just followed the STLink rules that were created by ST. We could change all of the rules to use groups? Just need to add some instructions in comments on how to set up the group for users that don't know, and maybe the MODE based rules, commented out, as an alternative. This would be a separate PR, though.

newbrain: The good thing with the rules as they are now is that zero knowledge or extra actions are needed for the user, just drop them in and they work.
A group (or username) based rule needs the user to tailor it according their identity.

Probably, an username rule is easier - as it would not involve:
* Creating an ad-hoc group or selecting an existing one (but hey vary across distributions)
* Making sure the user belongs to the group

Only this step would be needed:
* put the username in the right place in the rule

I might try to write a template this WE, depending on available time - no guarantee.

flit: I'm sure that's why ST chose MODE based (and it made sense for me).

The ultimate would be to have all 3 variants: "just works", user based, group based. Separated into different folders under `./udev/`. That's probably overkill, though. 😁  If you find time, I will happily add your additions!

PR #1127, opened 2021-03-31
Board IDs: update few stm32 name

jeromecoutant:
Hi @flit 

I propose few changes for some stm32 boards

Thx


flit:
Hi @jeromecoutant, Thanks for the PR!

A few of the changed boards had the target type set to a built-in STM32 target:

- `stm32f412xg`
- `stm32l475xg`
- `stm32f429xi`
- `stm32f439xi`

STM32 support from built-in targets is currently a little better than from the CMSIS-Packs. Mostly because the DBGMCU registers are configured automatically. (Once debug sequences are implemented in pyocd, this will change.)

jeromecoutant:
OK, so impact of this PR is now quite minor :-)

flit:
Still useful! 😄

Issue #1126, opened 2021-03-31
command "pyocd list", error occurred


lennvn:
Run command "pyocd list", error occurred, pyocd version >= 0.30:
Successfully installed pyocd-0.30.0

```
PS C:\Users\wuze> pyocd list
0000671:WARNING:common:STLink and CMSIS-DAPv2 probes are not supported because no libusb library was found.
0000775:CRITICAL:__main__:No backend available
Traceback (most recent call last):
  File "d:\program files\python\python39\lib\site-packages\pyocd\__main__.py", line 402, in run
    self._COMMANDS[self._args.cmd](self)
    ConnectHelper.list_connected_probes()
    allProbes = ConnectHelper.get_all_connected_probes(blocking=False)
  File "d:\program files\python\python39\lib\site-packages\pyocd\core\helpers.py", line 82, in get_all_connected_probes
    allProbes = DebugProbeAggregator.get_all_connected_probes(unique_id=unique_id)
  File "d:\program files\python\python39\lib\site-packages\pyocd\probe\aggregator.py", line 64, in get_all_connected_probes
    probes += cls.get_all_connected_probes(unique_id, is_explicit)
  File "d:\program files\python\python39\lib\site-packages\pyocd\probe\picoprobe.py", line 335, in get_all_connected_probes
    return [cls(dev) for dev in PicoLink.enumerate_picoprobes()]
  File "d:\program files\python\python39\lib\site-packages\pyocd\probe\picoprobe.py", line 108, in enumerate_picoprobes
    return [PicoLink(probe) for probe in core.find(find_all=True, custom_match=FindPicoprobe(uid))]
  File "d:\program files\python\python39\lib\site-packages\usb\core.py", line 1299, in find
    raise NoBackendError('No backend available')
usb.core.NoBackendError: No backend available
```

newbrain:
Please check DPR #1128.

flit:
@newbrain Thanks for responding so quickly!

PR #1128, opened 2021-03-31
Fix for missing usb backend in picoprobe

newbrain:
<ins>Preliminary</ins> Fix for #1126:

Catch the unhandled exception NoBackendError from pyusb.

It seems that I'm unable to eradicate libusb-win32 from my machines (tried some different methods).
So to test this I'll have to setup a clean VM/physical machine.
I will update here when done.

@lennvn: if not asking too much, could you please try it?

### DIFF
--- a/pyocd/probe/picoprobe.py
+++ b/pyocd/probe/picoprobe.py
@@ -26,6 +26,7 @@ import logging
 from typing import List
 
 from .debug_probe import DebugProbe
+from .common import show_no_libusb_warning
 from ..core import exceptions
 from ..core.options import OptionInfo
 from ..core.plugin import Plugin
@@ -108,9 +109,8 @@ class PicoLink(object):
             # Use a custom matcher to make sure the probe is a Picoprobe and accessible.
             return [PicoLink(probe) for probe in core.find(find_all=True, custom_match=FindPicoprobe(uid))]
         except core.NoBackendError:
-            LOG.debug(" No usb backend found")
+            show_no_libusb_warning()
             return []
-
 
     def q_read_bits(self, bits):
         """! @brief Queue a read request for 'bits' bits to the probe """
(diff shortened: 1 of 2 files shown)


pyocd/probe/picoprobe.py
```diff
@@ -104,8 +104,13 @@ def close(self):
     @classmethod
     def enumerate_picoprobes(cls, uid=None) -> List["PicoLink"]:
         """! @brief Find and return all Picoprobes """
-        # Use a custom matcher to make sure the probe is a Picoprobe and accessible.
-        return [PicoLink(probe) for probe in core.find(find_all=True, custom_match=FindPicoprobe(uid))]
+        try:
+            # Use a custom matcher to make sure the probe is a Picoprobe and accessible.
+            return [PicoLink(probe) for probe in core.find(find_all=True, custom_match=FindPicoprobe(uid))]
+        except core.NoBackendError:
+            LOG.debug(" No usb backend found")
```
flit, line 111: Change to make use of `pyocd.probe.common.show_no_libusb_warning()`, and update the message in that function.

flit:
This is the same approach used by both the STLink and CMSIS-DAPv2 plugins, so I'm sure it will work fine.

I really need to follow through with creating a Python package that wraps up the libusb library, so no separate install step is required.

pyocd/probe/picoprobe.py
newbrain: I had noticed that, and was wondering whether it was a better choice.
In the end I decided not change an unrelated file, but it's a better solution so I'll go for it.

Would you suggest adding picoprobe explicitly or just changing it to a generic message such as "Libusb is not installed: some probe types cannot be supported" and adding PP to the documentation?

newbrain: Ok, I'll make this last change!

lennvn:
> Preliminary Fix for #1126:
> 
> Catch the unhandled exception NoBackendError from pyusb.
> 
> It seems that I'm unable to eradicate libusb-win32 from my machines (tried some different methods).
> So to test this I'll have to setup a clean VM/physical machine.
> I will update here when done.
> 
> @lennvn: if not asking too much, could you please try it?

I installed the latest dev version and no error occurred with command 'pyocd list'

newbrain:
Thanks for following it up.

PR #1141, opened 2021-04-16
Docs: Add '--' to pyocd pack install and find example

sambhurst:


flit:
Thanks for the PR. The problem with the `pyocd pack` command docs is actually an unfortunate consequence of the difference between the released version of pyocd and what's on the master branch… Recently I extended `pyocd pack` to have subcommands of its own for the different operations (trying out a new and hopefully better command style). This hasn't been released yet, but I went ahead and updated the docs so I didn't forget. 

To help resolve this, I've been thinking about switching to a main/develop branch strategy, so what's on main always reflects the latest release. In addition, I'm working on a pyocd.io site that will host docs separately from the repo.

PR #1151, opened 2021-05-12
Added pudb3 + wrapper to debug pyOCD in Unix shell.

cederom:

### DIFF
--- a/pudb-pyocd.sh
+++ b/pudb-pyocd.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+# This script allows pyOCD debug directly in Unix shell.
+# 
+# You can insert breakpoint with `import pudb; pudb.set_trace()`.
+#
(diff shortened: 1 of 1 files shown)


cederom:
Whoops. Internals of pyOCD changed a lot recently. I have updated the script. With this kind of reorganization why not a major number bump (i.e. 1.0 instead 0.30)? :-)

flit:
(Sorry it's taken me a while to get to your PRs.)

Imo this seems to be more of a personal utility than something most pyocd users would benefit from. It's really just prefixing `pudb3` on the pyocd command-line, and the name of this script itself is almost the same amount of typing. Also, the choice of which Python command-line debugger to use is quite personal.

Probably a better option would be to extend the developer docs with a "Debugging pyOCD" section that describes this. (And how to debug in VSCode, which more people use, including me. I've been planning to document this for some time…)

cederom:
Allrgith, this will land into the documentation then :-)

PR #1154, opened 2021-05-12
Updated error message on USB Interface backend autodetection.

cederom:
New message is more informative and does not mislead to pyusb errors.

### DIFF
--- a/pyocd/probe/pydapaccess/interface/__init__.py
+++ b/pyocd/probe/pydapaccess/interface/__init__.py
@@ -59,6 +59,6 @@ if not USB_BACKEND:
     elif system == "Linux":
         USB_BACKEND = "pyusb"
     else:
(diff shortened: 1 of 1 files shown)


pyocd/probe/pydapaccess/interface/__init__.py
```diff
@@ -59,6 +59,6 @@
     elif system == "Linux":
         USB_BACKEND = "pyusb"
     else:
-        raise DAPAccessIntf.DeviceError("No USB backend found")
+        raise DAPAccessIntf.DeviceError("Unknown platform! No USB Interface backend autoselected.")
```
flit, line 62: Please change "Interface" to "interface" (not capitalised), and "autoselected" to "auto-selected" (the dictionary says the version without a dash is not a valid word).

flit (changes requested):

PR #1156, opened 2021-05-12
Fallback to pyusb as the default backend.

cederom:
This will try to use `pyusb` as the default USB Interface backend.
So far only selected list of OS has this parameter set.
This prevents pyOCD working on a platforms that are not on the list.
`pyusb` may be used as "fallback" in this case as most versatile solution.
If `pyusb` is not available or fails to work we will know that explicitly.

### DIFF
--- a/pyocd/probe/pydapaccess/interface/__init__.py
+++ b/pyocd/probe/pydapaccess/interface/__init__.py
@@ -59,6 +59,7 @@ if not USB_BACKEND:
     elif system == "Linux":
         USB_BACKEND = "pyusb"
     else:
-        raise DAPAccessIntf.DeviceError("No USB backend found")
+        #raise DAPAccessIntf.DeviceError("No USB backend found")
+        USB_BACKEND = "pyusb"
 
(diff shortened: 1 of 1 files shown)


flit:
Not a bad idea. 😄

pyocd/probe/pydapaccess/interface/__init__.py
```diff
@@ -59,6 +59,7 @@
     elif system == "Linux":
         USB_BACKEND = "pyusb"
     else:
-        raise DAPAccessIntf.DeviceError("No USB backend found")
+        #raise DAPAccessIntf.DeviceError("No USB backend found")
```
flit, line 62: Should check the `PyUSB.isAvailable` class attribute first. If that is False, then raise the exception, otherwise set the backend and continue.

flit (changes requested):

flit:
Also need to synchronize with #1154.

cederom:
Allright need get back to this as in new venv this blocks pyocd operations :-)

PR #1166, opened 2021-06-09
ST board Id: Alignment with mbed-os-tools v1.8.6

jeromecoutant:
Alignment with mbed-os-tools v1.8.6


jeromecoutant:
Same action done in mbed-os-tools: https://github.com/ARMmbed/mbed-os-tools/pull/262

@JeanMarcR

flit (approved):

PR #1172, opened 2021-06-24
Session: allow project_dir as an env var

g-arjones:
It's a very common use case to have env files that setup a workspace for a given project (i.e virtualenv). This feature can be very handy in these situations

### DIFF
--- a/pyocd/core/session.py
+++ b/pyocd/core/session.py
@@ -149,7 +149,7 @@ class Session(Notifier):
         
         # Init project directory.
         if self.options.get('project_dir') is None:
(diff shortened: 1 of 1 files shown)


flit:
Thanks!

g-arjones:
@flit No problem. If I may, how often are new versions released?

flit:
Well, there's not a defined, regular schedule for releases, but I try to get a new minor (feature) version out every 2-3 months or so, more often for bug fix releases. It's actually approaching time for a new release, and I'm starting to think about when the right time is.

All that means a new version will be coming pretty soon. 😄

PR #1174, opened 2021-07-07
Add NXP ke17z and lpc55s36 target

Hoohaha:


pyocd/target/builtin/target_LPC55S36.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+}
+
+
+class LPC55S36(LPC5500Family):
+
+    MEMORY_MAP = MemoryMap(
+        FlashRegion(name='nsflash',     start=0x00000000, length=0x040000, access='rx',
+            page_size=0x200,
+            sector_size=0x8000,
```
flit, line 93: Fyi,  the sector size is actually 0x200 in the hardware, and the value in the DFP flash algo is incorrect. I verified this with the RM and testing. However, because the flash algo always erases 0x8000 we have to leave the sector size at that value.

And in fact, the flash algo, or at least the LPC55S69 version, has a bug that it will erase 0x8000 bytes started at whatever address you give it, even if that address is not aligned to 0x8000.

(I reported this to the SDK team via Jason Martin a while ago.)

pyocd/target/builtin/target_LPC55S36.py
```diff
@@ -0,0 +1,119 @@
...
+        RamRegion(  name='scoderam',    start=0x14000000, length=0x4000, access='srwx',
+            alias='nscoderam',
+            default=False),
+        RamRegion(  name='nsram',       start=0x20000000, length=0x01c000, access='rwx'),
+        RamRegion(  name='sram',        start=0x30000000, length=0x01c000, access='srwx',
+            alias='nsram'),
+        )
+
+    def __init__(self, session):
+        super(LPC55S36, self).__init__(session, self.MEMORY_MAP)
```
flit, line 118: For future reference, you no longer need to pass parameters to `super()` because pyocd no longer needs to support Python 2.

flit (approved):

flit:
(Fyi, I had to manually re-do the merge of this PR. The PR was originally targeted to the `main` branch, but all new feature PRs should now be targeted at `develop`. I changed this, but unfortunately git pulled in a copy (4b68539) of a readme change commit (6a04d06) that I didn't want.)

pyocd/target/builtin/target_LPC55S36.py
Hoohaha: I will check with our team to fix it. Thanks flit!

flit: Thanks!

Issue #1143, opened 2021-04-19
Invalid firmware version on CMSIS-DAP


kongr45gpen:
I am trying to interface with an ATSAMV71-XULT, and getting this error:
```
INFO:pyocd.board.board:Target type is atsamv71q21b
Traceback (most recent call last):
  File "./test.py", line 22, in <module>
    with ConnectHelper.session_with_chosen_probe(unique_id=args.unique_id) as session:
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/core/session.py", line 339, in __enter__
    self.open()
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/core/session.py", line 453, in open
    self._probe.open()
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/probe/cmsis_dap_probe.py", line 170, in open
    self._link.open()
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/utility/concurrency.py", line 28, in _locking
    return func(self, *args, **kwargs)
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py", line 644, in open
    self._read_protocol_version()
  File "/home/kongr45gpen/.local/lib/python3.7/site-packages/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py", line 624, in _read_protocol_version
    patch = int(fw_version[2] if len(fw_version) > 2 else 0)
ValueError: invalid literal for int() with base 10: '01B6'
```

The reported firmware version for CMSIS-DAP is:
```
fw_version_str = "03.25.01B6"
```

I'm not sure if the documentation supports a hexadecimal string as the patch version or not, and not sure how to go around this issue. Also note that working with this board worked perfectly before commit bbf361e3b289f19d1ed354fa017a460b119d9cd8.

phsommer:
I have observed a similar problem with the Particle Debugger running firmware version [v0254.2](https://github.com/particle-iot/DAPLink/releases/tag/v0254.2):

```
File "/Users/xyz/venv/lib/python3.9/site-packages/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py", line 645, in open
    self._read_protocol_version()
[...]

PR #1201, opened 2021-08-23
Add work-around for incorrectly reported DAP version in microchip EDBG

mcbridejc:
Addresses issue #1143, by adding an exception for the Microchip EDBG firmware (found on xplained pro boards) which incorrectly reports some number -- I assume a firmware version for EDBG -- instead of the DAP protocol version in the DAP_Info command.

### DIFF
--- a/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py
+++ b/pyocd/probe/pydapaccess/dap_access_cmsis_dap.py
@@ -25,7 +25,7 @@ from .dap_settings import DAPSettings
 from .dap_access_api import DAPAccessIntf
 from .cmsis_dap_core import CMSISDAPProtocol
 from .interface import (INTERFACE, USB_BACKEND, USB_BACKEND_V2)
-from .interface.common import ARM_DAPLINK_ID
+from .interface.common import ARM_DAPLINK_ID, MICROCHIP_EDBG
 from .cmsis_dap_core import (
     Command,
     Pin,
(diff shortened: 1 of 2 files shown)


pyocd/probe/pydapaccess/dap_access_cmsis_dap.py
```diff
@@ -602,6 +602,8 @@ def _read_protocol_version(self):
         # based on unclear documentation.
         elif (self._vidpid == ARM_DAPLINK_ID) and (fw_version_str in ("0254", "0255", "0256")):
             self._cmsis_dap_version = CMSISDAPVersion.V2_0_0
+        elif (self._vidpid == MICROCHIP_EDBG) and (fw_version_str == "03.25.01B6"):
```
flit, line 605: Does this version string change with different EDBG firmware releases? It might be necessary to parse the version and use a more complex matching expression.

flit:
Thanks! It's interesting to see someone else misinterpret the CMSIS-DAP specification for that DAP_Info value. In CMSIS-DAP v2.1, recently released, we added a [new Info value](https://arm-software.github.io/CMSIS_5/develop/DAP/html/group__DAP__Info.html) for the firmware version and made it much more clear that the original is the protocol version.

flit:
Would it be possible for you to rebase your branch onto `develop`? I tried changing the base branch for this PR to `develop` but that brings in some unwanted commits. (I'm still working out the new branch policy for pyocd, so it might be ok to merge this to `main` and then I can cherry-pick to `develop`.)

mcbridejc:
I can definitely re-base. I'll wait to hear from you if you want to proceed with this narrow approach or do something more general (#1143).

flit:
As per the discussion on other issue, instead of this narrow change let's switch to wrapping the version parse with an exception handler and check for valid CMSIS-DAP protocol versions. No worries if you can't get to it, I'll be happy to implement it.

flit:
Closing in favour of #1203 that implements the generic checks.

Thanks for creating this PR that reminded me of the issue!

Issue #1148, opened 2021-05-05
memory transfer fault when resetting STM32H743ZI2


rotu:
I'm using Mbed studio to try and debug on a STM32H743ZI2, and it fails with a read error. "0000760:ERROR:rom_table:Error attempting to probe CoreSight component referenced by ROM table entry #0: Memory transfer fault (read) @ 0xe000efbc-0xe000f043".

A minimal command that causes this is:
`pyocd cmd -c reset`

Repeatedly trying to reproduce this shows the fault happens at particular addresses: `0xe00f0fbc-0xe00f1043` or `0xe000efbc-0xe000f043`

According to STM32H7 programming manual, there is a "Vendor-specific memory" block from 0x0000000 to 0xE0100000. There is a note that suggests naive access to this memory can cause a crash:

> Using the Peripheral, Private peripheral bus, or Vendor-specific memory areas can lead to unpredictable behavior in some systems. This is because the processor uses a
different interfaces for load/store instructions and vector fetch in these memory areas. If the vector table is located in a region of memory that is cacheable, core must treat any load or store to the vector as self-modifying code and use cache maintenance instructions to synchronize the update to the data and instruction caches

See logs:
[mbed-studio.log](https://github.com/pyocd/pyOCD/files/6430775/mbed-studio.log)
[minimal.log](https://github.com/pyocd/pyOCD/files/6430784/minimal.log)

rotu:
According to STM32H7 programming manual, there is a "Vendor-specific memory" block from 0x0000000 to 0xE0100000. There is a note that suggests naive access to this memory can cause a crash, which I suspect is the root cause:

> Using the Peripheral, Private peripheral bus, or Vendor-specific memory areas can lead to unpredictable behavior in some systems. This is because the processor uses a
different interfaces for load/store instructions and vector fetch in these memory areas. If the vector table is located in a region of memory that is cacheable, core must treat any load or store to the vector as self-modifying code and use cache maintenance instructions to synchronize the update to the data and instruction caches

rotu:
More from the manual

[...]

PR #1213, opened 2021-09-19
Don't read CoreSight registers for non-CoreSight components

j4cbo:
The CoreSight spec (IHI0029D B2.3) says "All component classes require the implementation
of the Component and Peripheral Identification registers", but non-CoreSight components
may sometimes fault on reading registers outside this space. In particular, a Cortex-M7
SCS, while the core is in reset, will fault on an attempt to read 0xe000efbc, which relates
to the data cache and not component identification.

To fix, we should avoid reading registers outside PIDR and CIDR unless the component
has the CoreSight component class.

This is a partial fix for pyocd/pyOCD#1148.

### DIFF
--- a/pyocd/coresight/rom_table.py
+++ b/pyocd/coresight/rom_table.py
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2015-2020 Arm Limited
 # Copyright (c) 2021 Chris Reed
+# Copyright (c) 2021 Jacob Berg Potter
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -36,10 +37,7 @@ class CoreSightComponentID(object):
     registers are made available as attributes.
     """
 
-    # CoreSight identification register offsets.
-    DEVARCH = 0xfbc
-    DEVID = 0xfc8
-    DEVTYPE = 0xfcc
+    # Component identification register offsets.
     PIDR4 = 0xfd0
     PIDR0 = 0xfe0
     CIDR0 = 0xff0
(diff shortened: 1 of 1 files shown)


j4cbo:
Anything needed on this? I think this change will be necessary even with pack debug sequence support - the pack will take care of setting DBGMCU but this is still needed for the connect-under-reset issue mentioned in #1148.

flit:
Oops, sorry, I thought I had replied!

I need to review the CoreSight specification in more detail and ask some questions of the CoreSight and CPU architects. There's always the possibility and even likelihood that this is required as a workaround, but I'd like to know the full context and document it correctly in the code. I'll try to look into it this week!

I agree that it sounds like components with a class that is not 0x9 may fault on accesses to undefined registers within the CoreSight management register space (offsets 0xF00-0xFFF), although it's actually not specified and the intention may be that any component that adheres to the CoreSight programmers' model should behave like this.

Also, the M-profile architecture specification states that all privileged accesses to undefined registers in the PPB space are RES0 (like RAZ/WI but reserved for future use). v7-M ARM § C1.2.1 says, "A reserved register or bit field has the value UNK/SBZP.". It's much more clearly defined in v8.1-M ARM rule R_GLNG: "Privileged accesses to unimplemented registers [in the SCS] are RES0.". But of course there are complications not defined by the architecture related to things like disabled clocks. 😄 

Sorry for being pedantic here… this is an area that I've had on my work todo list for some time to get more explicitly documented.

j4cbo:
Makes sense! Definitely agreed with getting the exact context right for documentation purposes. FWIW, I see faults reading `0xe000efbc` when using `--connect=under-reset` with both STM32F750 (Cortex-M7 r0p1) and STM32H750 (Cortex-M7 r1p1).

I agree with your reading that those registers should all "normally" be readable (RES0), so the question is whether the fact that they _aren't_ readable while the core is in reset is working-as-intended or a potential erratum?

claymation:
#1304 fixes a related issue, where the debug probe would remain locked after a memory transfer fault, preventing the GDB server thread from being able to use the probe at all.

j4cbo:
Bump :)

flit:
Hi @j4cbo, I haven't forgotten about this! 😄 Sorry for moving slowly.

I've been preparing the 0.33 release for release soon. I was trying to get the CMSIS-Pack debug sequences feature finished up, but haven't had time because of fixing other bugs.

My concern right now is that breaking up the ID read into two parts will slow down connect performance. Naturally, that will be most noticeable on probes with FS USB. I'll try some quick performance tests to check the difference. Deferred reads might be a workaround here, always performing the second read but just ignoring any faults.

Anyway, I'm looking to get this fix into 0.33!

j4cbo:
Thanks! I understand the speed concerns - it still feels a little better to me to avoid reading the potentially-unreadable addresses, but if the performance hit is too much, ignoring the fault would work too (#1304 plus some extra recovery in the ROM table code?)

flit:
Turns out the performance impact is negligible! And depending on the device and probe, there may actually be a small improvement from not reading the extra registers from Type 0x1 ROM tables. (ADIv6 Type 0x9 ROM tables are CoreSight components.)


 Branch                    | STM32H743 (STLinkV3, HS)    | MIMXRT1052 (DAPLink, FS)
-------------------|----------------|-----
develop e2adf0c    |  1.671 | 1.670
pull/1213/merge 69dc193 | 1.738 | 1.535

Basically, the change is almost below the noise level. 😄

flit (approved):

PR #1253, opened 2021-11-30
RTT host-to-target communication

ccattuto:
This is an initial implementation of RTT host-to-target communication for pyOCD's `rtt` subcommand.
 - It is meant to provide bidirectional, terminal-like RTT communication with the target, with the assumption that the user will type on `stdin` short commands for the target
 - `stdin` is read one character at a time, in a non-blocking fashion. Characters are echoed to the terminal and accumulated in a buffer until we see a CR of LF. At that point, if there is sufficient space in the RTT down-buffer (host-to-target), the buffer is copied character-by-character to the RTT ring buffer.
 - the code processes backspace characters from `stdin`, allowing the user to make corrections before hitting CR/LF. ESC terminates the session.
 - all the modifications were carried out within `rtt_cmd.py`, including the addition of a class (copied from [here](https://stackoverflow.com/questions/2408560/non-blocking-console-input)) for cross-platform, non-blocking read of 1 character from stdin.

### DIFF
--- a/pyocd/subcommands/rtt_cmd.py
+++ b/pyocd/subcommands/rtt_cmd.py
@@ -26,9 +26,8 @@ from pyocd.subcommands.base import SubcommandBase
 from pyocd.utility.cmdline import convert_session_options, int_base_0
 from ctypes import Structure, c_char, c_int32, c_uint32, sizeof
 
-# The KBHit code below is based on the file https://simondlevy.academic.wlu.edu/files/software/kbhit.py ,
-# re-licensed under an MIT License by its author Simon D. Levy (https://github.com/simondlevy)
-# for inclusion in the pyOCD project.
+# The KBHit code below is based on https://github.com/simondlevy/kbhit by Simon D. Levy,
+# published under the MIT License.
 
 import os
 
@@ -90,7 +89,7 @@ class KBHit:
             dr,dw,de = select([sys.stdin], [], [], 0)
             return dr != []
 
-# end of the KBHit code re-licensed by its author Simon D. Levy under an MIT License 
+# end of the KBHit code
 
 
 LOG = logging.getLogger(__name__)
--- a/pyocd/subcommands/rtt_cmd.py
+++ b/pyocd/subcommands/rtt_cmd.py
@@ -24,72 +24,8 @@ from pyocd.core.memory_map import MemoryMap, MemoryRegion, MemoryType
 from pyocd.core.soc_target import SoCTarget
 from pyocd.subcommands.base import SubcommandBase
 from pyocd.utility.cmdline import convert_session_options, int_base_0
+from pyocd.utility.kbhit import KBHit
 from ctypes import Structure, c_char, c_int32, c_uint32, sizeof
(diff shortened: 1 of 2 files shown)


pyocd/subcommands/rtt_cmd.py
```diff
@@ -24,6 +24,70 @@
...
+            pass
+        else:
+            termios.tcsetattr(self.fd, termios.TCSAFLUSH, self.old_term)
+
+    def getch(self):
+        ''' Returns a keyboard character after kbhit() has been called.
+            Should not be called in the same program as getarrow().
+        '''
+
+        s = ''
```
flit, line 74: This is the unused local variable identified by LGTM.

flit (changes requested):
Since this change is pretty significant, would you mind adding your copyright to the `rtt_cmd.py` header comment?

Aside from that, just please fix the unused local variable warning from LGTM.

If you feel like it, type annotations would be nice but aren't strictly required yet.

pyocd/subcommands/rtt_cmd.py
flit: Ooh, good catch on the license! Thanks for following up on that!

ccattuto: @simondlevy kindly agreed to re-license the KBHit class under an MIT License (which is Apache 2.0 compatible) for inclusion in pyOCD. I'll go ahead and add this information to the copyright notice, adding his name too.

ccattuto: All should be OK now.

ccattuto: @simondlevy made available the code on GitHub under the MIT License. I've updated the comments accordingly.

flit: Exellent! 😄 

We should probably put KBHit in its own `pyocd.utility.kbhit.py` module, so the whole file is MIT-licensed and has his copyright. (I'll also add this comment on the `KBHit` class in the updated `rtt_cmd.py` file.)

pyocd/subcommands/rtt_cmd.py
```diff
@@ -24,6 +26,72 @@
 from pyocd.utility.cmdline import convert_session_options, int_base_0
 from ctypes import Structure, c_char, c_int32, c_uint32, sizeof
 
+# The KBHit code below is based on https://github.com/simondlevy/kbhit by Simon D. Levy,
+# published under the MIT License.
```
flit, line 30: KBHit should probably be moved to its own `pyocd.utility.kbhit.py` module, so the whole file is MIT-licensed.

pyocd/subcommands/rtt_cmd.py
ccattuto: Done.

flit (approved):

pyocd/subcommands/rtt_cmd.py
flit: Thanks!

PR #1261, opened 2021-12-09
Add new micro:bit board IDs

mbrossard:


flit:
Thanks! Could you please retarget this to `develop` (since it's not a bug fix)?

flit:
Thanks!

PR #1265, opened 2021-12-11
Exit gracefully if RTT block is not found

ccattuto:
Minor fix to avoid chocking on undefined `kb` if RTT control block is not found.


flit (approved):

flit:
Thanks 😄

PR #1280, opened 2021-12-25
ELF files: clear T-bit on symbol addresses

flit:


flit:
Closing PR because this is a backwards-incompatible change that will have to wait til later.

PR #1305, opened 2022-01-12
Gdbserver: clear shutdown event after disconnect

claymation:
Otherwise, the listening socket is closed, preventing reconnection.


flit:
Thanks for all the fixes!

PR #1318, opened 2022-01-21
Switch from naturalsort to natsort

### LINKED ISSUE #1317: Replace naturalsort dependency with natsort or remove altogether
Hi! I'm about to package this project for Arch Linux (due to other dependencies).

Unfortunately the [naturalsort](https://github.com/xolox/python-naturalsort) upstream seems very dead, [does conflict with natsort](https://github.com/xolox/python-naturalsort/issues/4) (which is actively maintained and already packaged on Arch Linux), has less features and thus seems less fit for being included in any upstream project.
It would be very cool if you could switch to using natsort, or remove the necessity of naturalsort altogether, which seems to be only required for this command:
https://github.com/pyocd/pyOCD/blob/7d164d99e816c4ef6c99f257014543188a0ca5a6/pyocd/commands/commands.py#L119-L120

dvzrv:
pyocd/commands/commands.py:
Use natsort's `natsorted()` instead of naturalsort's `natsort()`.

setup.cfg:
Replace naturalsort with natsort.

Fixes #1317

### DIFF
--- a/pyocd/commands/commands.py
+++ b/pyocd/commands/commands.py
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2015-2020 Arm Limited
 # Copyright (c) 2021 Chris Reed
+# Copyright (c) 2022 David Runge
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


flit (approved):
Since it's only a 2 line change in the .py file, I won't make it a strict requirement that you add your copyright to the file header, but it would be nice.

dvzrv:
> 

Sure, I can add it, no problem.

I hope my change works as intended!

flit:
Thanks!

I just did a quick test, and it works as expected. 👍🏽

Issue #1322, opened 2022-01-28
v0.33.0 USB Error: Operation timed out


BlokhinDima:
Instantly have this exception when trying to use v0.33.0. For e.g. if i try `pyocd list` command. My proggramer is ST-link. OS Windows 10. (As i understand i don't need libusb-1.0.dll in Python folder, but i've tried to add it anyway. Doesn't work.) 

Everything is good with v0.32.3 and libusb-1.0.dll installed in Python 3.10.1 folder.

```
0007003 C USB Error: [Errno 10060] Operation timed out [__main__]
```

Please ask, if you need more information.

flit:
Thanks for reporting the issue! The change to use a copy of libusb automatically provisioned as a dependency (via [libusb-package](https://github.com/pyocd/libusb-package)) of pyocd is new in v0.33, so I'm not totally surprised there are issues even with all the testing I did. Thanks also for confirming that v0.32.3 works for you.

First questions:
- What version of libusb did you install that works? (libusb-package currently contains version libusb 1.0.24.)
- Is your Python installation 32-bit or 64-bit?
- Would you mind running `pyocd list -vv -L "*trace=debug"` and providing the output here?

Thanks! I will also try to reproduce.

vacagonzalo:
> Thanks for reporting the issue! The change to use a copy of libusb automatically provisioned as a dependency (via [libusb-package](https://github.com/pyocd/libusb-package)) of pyocd is new in v0.33, so I'm not totally surprised there are issues even with all the testing I did. Thanks also for confirming that v0.32.3 works for you.
> 
> First questions:
> 
>     * What version of libusb did you install that works? (libusb-package currently contains version libusb 1.0.24.)
> 
>     * Is your Python installation 32-bit or 64-bit?
> 
>     * Would you mind running `pyocd list -vv -L "*trace=debug"` and providing the output here?
> 
> 
> Thanks! I will also try to reproduce.

Hello, I cannot run pyocd list.
Note: I cp udev rules.

USB Error: [Errno 110] Operation timed out

My groups:

administrador : administrador adm dialout cdrom sudo dip plugdev lpadmin lxd sambashare docker

Debug msg:

[...]

PR #1325, opened 2022-02-06
probe: stlink: disable usage of command to get the board ID.

flit:
The `JTAG_GET_BOARD_IDENTIFIERS` command that is disabled by this change will fail to respond for at least some versions of STLinkV2 firmware, thus causing a USB timeout error.

See #1322 for details of the issue fixed by this change.


flit:
Note: This commit from this PR was force-pushed off of `main` only a few minutes after being merged, as a proper fix for #1322 was identified.

Issue #1327, opened 2022-02-09
v0.33.1 reset type not honoured


Hoohaha:
Tested v0.33.1 version on evkbimxrt1050, the mass erase always failed.

- OS: Windows 10
- Python 3.9

Added `print(reset_type)` in `family/target_imxrt.py line:99`, found that reset_type always is `ResetType.SW_SYSRESETREQ` no matter i changed the default reset type.

If force set `reset_type=self.ResetType.SW` in `def set_reset_catch(self, reset_type=None):` seems it works.
Do not know how to handle this properly.

```
"C:\Python39\python.exe" -m pyocd erase --mass -v --no-wait -u 02380000082e00d800000000000000000000000097969905 -Oresume_on_disconnect=false -Oallow_no_cores=true
0001264 I Target type is mimxrt1024 [board]
0001297 I DP IDR = 0x0bd11477 (v1 MINDP rev0) [dap]
0001328 I AHB-AP#0 IDR = 0x04770041 (AHB-AP var4 rev0) [ap]
0001375 I AHB-AP#0 Class 0x1 ROM table #0 @ 0xe00fd000 (designer=00e part=88c) [rom_table]
0001375 I [0]<e00fe000:ROM class=1 designer=43b part=4c8> [rom_table]
0001375 I   AHB-AP#0 Class 0x1 ROM table #1 @ 0xe00fe000 (designer=43b part=4c8) [rom_table]
0001394 I   [0]<e00ff000:ROM class=1 designer=43b part=4c7> [rom_table]
0001394 I     AHB-AP#0 Class 0x1 ROM table #2 @ 0xe00ff000 (designer=43b part=4c7) [rom_table]
0001398 I     [0]<e000e000:SCS v7-M class=14 designer=43b part=00c> [rom_table]
0001400 I     [1]<e0001000:DWT v7-M class=14 designer=43b part=002> [rom_table]
0001403 I     [2]<e0002000:FPB v7-M class=14 designer=43b part=00e> [rom_table]
0001403 I     [3]<e0000000:ITM v7-M class=14 designer=43b part=001> [rom_table]
0001403 I   [1]<e0041000:ETM M7 class=9 designer=43b part=975 devtype=13 archid=4a13 devid=0:0:0> [rom_table]
0001403 I   [2]<e0042000:CTI class=9 designer=43b part=906 devtype=14 archid=0000 devid=40800:0:0> [rom_table]
[...]

PR #1328, opened 2022-02-09
imxrt:force use SW reset for set_reset_catch

Hoohaha:
Signed-off-by: Haley Guo <hui.guo@nxp.com>

#1327


Hoohaha:
The change cannot fix the issue. closed

PR #1330, opened 2022-02-09
Increase supported python version to >=3.7.0

dvzrv:
Increase supported python version to >= 3.7.0

### DIFF
--- a/.github/workflows/basic_test.yaml
+++ b/.github/workflows/basic_test.yaml
@@ -27,7 +27,6 @@ jobs:
         - macos-latest
         - windows-latest
         python-version:
-        - "3.6"
         - "3.7"
         - "3.8"
         - "3.9"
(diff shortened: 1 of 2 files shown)


flit:
Is there a reason to not support Python 3.6? I actually need it for one of the projects I work on due to the OS version and available packages. So I suspect other users also still need it.

flit:
Btw, sorry for the slow replies this week, I've been moving into a new house.

dvzrv:
> Is there a reason to not support Python 3.6? I actually need it for one of the projects I work on due to the OS version and available packages. So I suspect other users also still need it.

Python 3.6 is not supported  by upstream anymore, so there is no reason to support it in downstreams.
FWIW, prettytable now also only supports >= 3.7

dvzrv:
It appears that you have bumped the supported version of prettytable to < 4.0.0 in https://github.com/pyocd/pyOCD/pull/1359, so I guess this can be closed.

Dropping support for python 3.6 is nonetheless a good idea and should be done as soon as possible, as it prevents using newer language features, etc.

dvzrv:
Ping?

flit:
Fyi, I've created a new PR #1477 based on this one, with a few minor changes like removing the `dataclasses` dependency.

Issue #1331, opened 2022-02-09
Using libusb-package is a problem for system packaging


dvzrv:
Hi! I would like to package this project for Arch Linux.
Unfortunately it requires libusb-package, which vendors a version of libusb. However, on Linux we already provide libusb and it does not make sense to vendor these shared libraries within the context of a python package.

The change introduced in ce05b146b3ec01fe4855c2f0e89b5e01f3a0bb1d is problematic as it would require downstream distributions to package a vendored libusb that can not be used by anything else.

It would be very awesome if the import of the module could be made optional.

dvzrv:
@flit I have created a PR (#1332) that makes the use of `libusb_package` optional and falls back to `usb.core` if it can not use it.
This would at least allow distributions (because frankly, no Linux distribution will want to use libusb-package, as there is already [pyusb](https://github.com/pyusb/pyusb) interfacing with [libusb](https://libusb.info/)) to simply remove the requirement for `libusb-package` in setup.py and plainly rely on pyusb.
There should probably a note about this though.

There were instances of guards against failing imports of `usb.core` and `usb.util` in some of the files, which were not in others.
I didn't dig much deeper and am not sure whether they are really required.

While working on the pull request I realized that libusb-package is solely used for the `find()` call and pyusb is used anyways. Is the overhead of shipping your own version of libusb in the context of a python package really worth it?

jjakob:
Patch to replace all uses of libusb_package with usb.core and remove the libusb_package dependency
https://github.com/jjakob/gentoo-overlay/blob/master/dev-python/pyocd/files/0001-Use-pyusb-instead-of-libusb_package-remove-libusb-pa.patch

PR #1332, opened 2022-02-11
Make the use of libusb-package optional

dvzrv:
pyocd/probe/*:
Attempt to import libusb-package's `find()` function, else fall back to using pyusb's `find()` function.
This allows to use pyusb by just removing the project's requirement on libusb-package, as python-pyusb and libusb
packaging go hand-in-hand on Linux distributions.

Related to #1331

### DIFF
--- a/pyocd/probe/picoprobe.py
+++ b/pyocd/probe/picoprobe.py
@@ -19,7 +19,10 @@ from array import array
 
 from time import sleep
 from usb import core, util
-import libusb_package
+try:
+    from libusb_package import find as usb_find
+except ImportError:
+    from usb.core import find as usb_find
 
 import platform
 import errno
@@ -108,7 +111,7 @@ class PicoLink(object):
         """@brief Find and return all Picoprobes """
         try:
             # Use a custom matcher to make sure the probe is a Picoprobe and accessible.
-            return [PicoLink(probe) for probe in libusb_package.find(find_all=True, custom_match=FindPicoprobe(uid))]
+            return [PicoLink(probe) for probe in usb_find(find_all=True, custom_match=FindPicoprobe(uid))]
         except core.NoBackendError:
             show_no_libusb_warning()
             return []
--- a/pyocd/probe/pydapaccess/interface/pyusb_backend.py
+++ b/pyocd/probe/pydapaccess/interface/pyusb_backend.py
@@ -41,13 +41,17 @@ TRACE = LOG.getChild("trace")
 TRACE.setLevel(logging.CRITICAL)
 
 try:
-    import libusb_package
     import usb.core
     import usb.util
+    try:
+        from libusb_package import find as usb_find
+    except ImportError:
+        from usb.core import find as usb_find
 except ImportError:
     IS_AVAILABLE = False
 else:
     IS_AVAILABLE = True
+
 
 class PyUSB(Interface):
     """@brief CMSIS-DAP USB interface class using pyusb for the backend."""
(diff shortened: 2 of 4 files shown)


elfmimi:
Any idea for promoting users to install libusb upon the user doing `pip install pyocd` ? @dvzrv

flit:
@elfmimi That's basically what libusb-package was intended to solve. Users no longer need to worry about installing libusb.

@dvzrv For distro packages, I totally understand the requirement to use the distro-installed libusb. It only makes sense.

How to handle distros was actually something under consideration when designing libusb-package, at least at a high level. The distro version of libusb-package should return the distro libusb. (And if the user separately installs libusb-package from PyPI, the user gets the prebuilt wheel containing libusb.) This is the path I'd much prefer to take.

libusb-package supports being installed without a contained copy of libusb, a so-called "empty install". In this case, it will always fall back to asking for the system libusb via `ctypes.find_library()`.

The part that's missing is a way to force an empty install. But is this even needed? How do you generate a distro package of a Python package? If you can just take the libusb-package source from the repo, then there's nothing additional needed—it will be an empty install by default.

A more advanced option would be to install a symlink to the distro libusb .so in order to explicitly pin libusb-package to that libusb version.

elfmimi:
I know. I know, kind of, I asked for it.

dvzrv:
> The part that's missing is a way to force an empty install. But is this even needed?

Yes, we do not want to rely on precompiled shared libraries, but on system libraries that can be built reproducibly and be used by more than one application.

> How do you generate a distro package of a Python package? If you can just take the libusb-package source from the repo, then there's nothing additional needed—it will be an empty install by default.

We usually go the PEP517 route these days. Packaging guidelines for python on Arch Linux can be found in the [wiki](https://wiki.archlinux.org/title/Python_package_guidelines).
Can you elaborate what you mean by "empty install"?

diggit:
>  Can you elaborate what you mean by "empty install"?

When I tried to install `libusb-package` using PKGBUIULD according to linked wiki page, resulting package did not carry any pre-built shared library and pyocd was happy. [Here is PKGBUILD](https://gist.github.com/diggit/9b61882d5520d34b0e303838317f1773) I tested with.

It would mean new dependency for `python-pyocd` package, which would have to go directly into `community` repo. Despite that, maintaining wrapper package looks like least intrusive solution.

dvzrv:
As there is only limited time to deal with packages, I wonder what there is left to do with this PR.
From a distribution standpoint it makes no sense at all to package `libusb-package`, as with this patch we can directly make use of pyusb and everything just works.

Operating systems which require the bundling of libusb to make this work can still rely on `libusb-package` if they want to (this PR just falls back to `pyusb` if it `libusb-package` is not available).

@flit This means libusb-package could become a dependency only for specific operating systems (e.g. windows and/or macOS). However, at this point I am still asking myself why this is needed at all. Have you tried to reach out to https://github.com/pyusb/pyusb, as that project has been around for a long time and libusb-package looks like duplicated effort if pyusb could just add wheels bundling libusb.

> It would mean new dependency for python-pyocd package, which would have to go directly into community repo. Despite that, maintaining wrapper package looks like least intrusive solution.

@diggit But this is precisely the point. I maintain over 700 packages for Arch Linux. Why would I want to package yet another wrapper around libusb if there is already pyusb?

dvzrv:
@flit I have rebased on latest default branch. *Please* have a look

a-detiste:
please

ithinuel (commented):
woops, I forgot to give the ✅ before merging.

PR #1346, opened 2022-03-15
pyDAPAccess: fix search for devices without UID

martjoly:
Probes without UID get a unique ID generated during device listing.
Attempting to connect to a probe with a generated UID will fail as
its synthesized UID will be compared against the device empty UID during
device search.


flit:
Thanks for the fix!

PR #1374, opened 2022-04-25
Update docs to cover resolution to 1364

jeffreyabecker:
I've found that hardware modifications to my test board fixed this issue. I've updated the docs to reflect the hardware requirements and how they differ from the chip documentation.


flit:
Thanks very much!

flit (approved):

PR #1375, opened 2022-04-27
RTT: search for RTT control block in smaller chunks

te-johan:
recently started using pyOCD and noticed that RTT searched for the control block rather slow.
when testing on my device to took 0.3 seconds to find it compared to 4.3 seconds before this commit.


flit (changes requested):
Would you mind adding your copyright at the end of the existing copyrights in the file header comment? No other requested changes—looks great! 😄

flit (approved):

PR #1392, opened 2022-05-15
probe: block signals around USB transfers

flit:
This patch blocks all signals from being sent to the current thread before performing any USB transfer through pyusb.
For the separate receive threads created by some CMSIS-DAP USB backends, signals are permanently blocked.

The main reason to do this is that a SIGINT can cause a `KeyboardInterrupt` exception to be raised in the middle of pyusb and/or pyocd USB protocol related code. On occasion, this will cause the protocol to become out of sequence, resulting in all future transfers failing wildly.


flit:
Going to pause this one. More work is needed. A `KeyboardInterrupt` could still be raised from within (for instance) the CMSIS-DAP internal packet management code, which can break the sequencing of protocol transfer commands.

A real question is whether pyocd should even be doing this. Really the application should install its own SIGINT handler that completely avoids using `KeyboardInterrupt` (which pyocd does not do for its own gdb server, etc, right now!). Perhaps pyocd should just provide a utility to make this easy.

PR #1405, opened 2022-06-02
Add NXP KW45 target support

Hoohaha:


flit (changes requested):

flit:
Does this device need to re-enter the debug session via DM-AP after hardware reset? (I hope not… debug logic being reset on hw reset is a Bad Thing imo! 😉 )

flit:
Btw, if you have a CMSIS-Pack with debug sequences for this device, you can try out the [feature/debug_sequences](https://github.com/flit/pyOCD/tree/feature/debug_sequences) on my work. It's nearly complete and going through extensive testing and debugging (eg, with the RT595 and LPC55xx packs, and many others). The plan is to integrate this for the next release, version 0.35.

This branch also allows use of pre-expanded .pack files (as directories) with the `--pack` argument, which is very convenient for pack development.

If you do try it out, you can turn on debug sequence tracing with `-Lpyocd.debug.sequences.sequences.trace=debug`.

PR #1407, opened 2022-06-02
[semihosting] ensure directories are exists before open file

Hoohaha:
Target maybe open a filename in a none existing directory. 
This change can avoid IOError exception.

And BTW J-Link also support to automatically create directories if they are not exists.

### DIFF
--- a/pyocd/debug/semihost.py
+++ b/pyocd/debug/semihost.py
@@ -1,5 +1,6 @@
 # pyOCD debugger
 # Copyright (c) 2015-2020 Arm Limited
+# Copyright (c) 2022 NXP
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


pyocd/debug/semihost.py
```diff
@@ -184,6 +185,9 @@ def open(self, fnptr, fnlen, mode):
             return fd
 
         try:
+            # ensure directories are exists
+            pathlib.Path(filename).parent.mkdir(parents=True, exist_ok=True)
```
flit, line 189: This should probably only be done if `mode` is writable, eg `('w' in mode) or ('a' in mode)`.

flit (changes requested):
Did you want to add an NXP copyright?

pyocd/debug/semihost.py
Hoohaha: thanks for your comments.
updated.

flit (approved):

Issue #1417, opened 2022-06-12
RTT settings not available as session options


flit:
Need to add session options for the RTT subcommand's settings:

- Control block search range start address.
- Control block search range size.

More controls are needed, too, such as a fixed address for the RTT CB, number of channels, etc.

DavidLesnjak:
Hi @flit,
In the [latest release](https://github.com/pyocd/pyOCD/releases/tag/v0.44.0), we added the `rtt` session option as a YAML node. Please see the [updated documentation](https://pyocd.io/docs/options.html) for the available configuration details.
At the moment, this configuration is used only by the `pyocd run` command. Support for using it with `gdbserver` is planned for upcoming releases.

PR #1418, opened 2022-06-12
Add rtt control block into YAML config

crides:
This follows the way in the `jlink_probe` module of getting configs.

Fixes the first 2 requirements in #1417

### DIFF
--- a/pyocd/subcommands/rtt_cmd.py
+++ b/pyocd/subcommands/rtt_cmd.py
@@ -119,13 +119,17 @@ class RTTSubcommand(SubcommandBase):
 
                 memory_map: MemoryMap = target.get_memory_map()
                 ram_region: MemoryRegion = memory_map.get_default_region_of_type(MemoryType.RAM)
-
-                if self._args.address is None or self._args.size is None:
-                    rtt_range_start = ram_region.start
-                    rtt_range_size = ram_region.length
-                elif ram_region.start <= self._args.address and self._args.size <= ram_region.length:
-                    rtt_range_start = self._args.address
-                    rtt_range_size = self._args.size
+                def valid_range(start, size):
+                    return start != None and size != None \
+                        and ram_region.start <= start and size + start <= ram_region.start + ram_region.length
+                rtt_range_start, rtt_range_size = None, None
+                config_start, config_size = session.options.get('rtt.start'), session.options.get('rtt.size')
+                if valid_range(self._args.address, self._args.size):
+                    rtt_range_start, rtt_range_size = self._args.address, self._args.size
+                elif valid_range(config_start, config_size):
+                    rtt_range_start, rtt_range_size = config_start, config_size
+                else:
+                    rtt_range_start, rtt_range_size = ram_region.start, ram_region.length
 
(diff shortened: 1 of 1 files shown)


pyocd/subcommands/rtt_cmd.py
```diff
@@ -119,13 +119,17 @@ def invoke(self) -> int:
...
-                    rtt_range_start = ram_region.start
-                    rtt_range_size = ram_region.length
-                elif ram_region.start <= self._args.address and self._args.size <= ram_region.length:
-                    rtt_range_start = self._args.address
-                    rtt_range_size = self._args.size
+                def valid_range(start, size):
+                    return start != None and size != None \
+                        and ram_region.start <= start and size + start <= ram_region.start + ram_region.length
+                rtt_range_start, rtt_range_size = None, None
+                config_start, config_size = session.options.get('rtt.start'), session.options.get('rtt.size')
```
flit, line 126: Could you please add the new options to `pyocd/core/options.py` so they have help?

flit (changes requested):
Aside from the `options.py` additions, would you mind adding your copyright to the top of the file below the others? Thanks!!

pyocd/subcommands/rtt_cmd.py
```diff
@@ -120,27 +124,48 @@ def invoke(self) -> int:
...
+                else:
+                    elf_file = None
+
+                if elf_file != None:
+                    elf = ELFFile(open(self._args.elf, "rb"))
+                    symtab = elf.get_section_by_name(".symtab")
+                    if symtab:
+                        rtt_entry = symtab.get_symbol_by_name("_SEGGER_RTT")
+                        if len(rtt_entry) > 0:
+                            rtt_cb_addr = rtt_entry[0].entry["st_value"]
```
flit, line 143: Fyi, pyocd has some built-in support for ELF symbol extraction (and other things) in the `pyocd.debug.elf` sub-package. It's based on pyelftools, so for cases like this, where there is not much advantage, it's fine to directly use pyelftools.

For reference, this is what this code would look like using pyocd's ELF tools:

```python
sym_info = ELFBinaryFile(self._args.elf).symbol_decoder.get_symbol_by_name('_SEGGER_RTT')
if sym_info:
    rtt_cb_addr = sym_info.value
```

flit (changes requested):
Thank you for the changes! If you remove the change to `setup.cfg` it should be ready to go!

DavidLesnjak:
Thanks for the contribution, and apologies that this PR has been left open for so long.

I’m closing it at this point because pyOCD now has a newer RTT configuration approach in this area, and that has already been documented and released.

Please see the [latest release](https://github.com/pyocd/pyOCD/releases/latest) and the [updated documentation](https://pyocd.io/docs/options.html).

When cbuild-run configuration is used, RTT control block detection is now supported via the _SEGGER_RTT symbol as well.

At the moment, the rtt YAML configuration is used only by the pyocd run command. Broader support in gdbserver is planned for upcoming releases.

If you have any additional ideas or requirements around RTT configuration, please open a new issue/PR.

Thanks again for the work on this.

PR #1428, opened 2022-07-05
probe: stlink: fix broken non-word-aligned 8-bit transfers

### LINKED ISSUE #1409: Breakpoint causes unhandled exeception
Setup: ST-Link v2 with an STM32F411CE. Setup works with Mbed Studio (which uses pyOCD). However, trying to use pyOCD without studio -- either the version included or the latest version -- causes a failure when gdb sets a breakpoint.  I have tried both the stripped down pack included with Mbed Studio and the full pack. I've tried remote and extended remote. I've try gdb-multiarch and the arm noabi gdb. All the same results.

The exact same setup works fine with OpenOCD so I think wiring etc isn't the issue.

When I try to set a breakpoint I get:
```
0012712 E Unhandled exception in handle_message (b'm'): result size (6) != requested size (4) [gdbserver]
  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/gdbserver/gdbserver.py", line 423, in handle_message

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/gdbserver/gdbserver.py", line 844, in get_memory
    mem = self.target_context.read_memory_block8(addr, length)

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/debug/cache.py", line 42, in read_memory_block8
    return self._memcache.read_memory_block8(addr, size)

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/cache/memory.py", line 258, in read_memory_block8
    assert len(result) == size, "result size ({}) != requested size ({})".format(len(result), size)

AssertionError: result size (6) != requested size (4)

0012712 E Unexpected exception: a bytes-like object is required, not 'tuple' [gdbserver]

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/gdbserver/gdbserver.py", line 390, in _run_connection
    self.packet_io.send(resp)

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/gdbserver/packet_io.py", line 83, in send
    self._write_packet(packet)

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/gdbserver/packet_io.py", line 146, in _write_packet
    written = self._abstract_socket.write(packet)

  File "/home/alw/.local/lib/python3.8/site-packages/pyocd/utility/sockets.py", line 58, in write
    return self.conn.send(data)

TypeError: a bytes-like object is required, not 'tuple'
```

I feel like I must be doing something wrong. I've tried simple invocations and trying to mimic the Mbed Studio invocation. So:

`pyocd gdbserver -t stm32f411ce`

Along with explicit pack definitions. Also 

`python3 -mpyocd gdbserver --target STM32F411CE --erase=chip --frequency 1800000 --pack '/home/alw/.config/Mbed Studio/mbed-studio-tools/cmsis-packs/Keil.STM32F4xx_DFP.2.15.0-small.pack' -O connect_mode=under-reset   -O enable_swv=1 -O swv_system_clock=96000000 -O swv_raw_port=3443 -O swv_clock=2000000`

And many variations (e.g., no swv, etc.).

flit:
Fix for an embarrassing bug introduced when STLink HPROT/HNONSEC support was added.

Fixes #1409

### DIFF
--- a/pyocd/probe/stlink_probe.py
+++ b/pyocd/probe/stlink_probe.py
@@ -306,7 +306,7 @@ class STLinkMemoryInterface(MemoryInterface):
         res = []
 
         # read leading unaligned bytes
-        unaligned_count = 4 - (addr & 3)
+        unaligned_count = addr & 3
         if (size > 0) and (unaligned_count > 0):
             res += self._link.read_mem8(addr, unaligned_count, self._apsel, csw)
             size -= unaligned_count
@@ -332,7 +332,7 @@ class STLinkMemoryInterface(MemoryInterface):
         idx = 0
 
         # write leading unaligned bytes
-        unaligned_count = 4 - (addr & 3)
+        unaligned_count = addr & 3
         if (size > 0) and (unaligned_count > 0):
             self._link.write_mem8(addr, data[:unaligned_count], self._apsel, csw)
             size -= unaligned_count


DariusBabrauskas (commented):
Note. It seems mistake. Because :
1. First need read8 until aligned address. There is mistake - prewios code was ok. But need check size with unaligned_count
2. Then read32 until last aligned adress.
3. Read8 last not aligned  adress data.

PR #1438, opened 2022-07-23
target: family: LPC5500: excludes LPC54 from regex, Fix #1437

### LINKED ISSUE #1437: Family: Wrong regex for LPC55 breaks support for LPC54
The target family LPC5500 has a regex `r'LPC55?[0-9]{2}.*'`, which accidently matches all LPC54 series MCUs. While using CMSIS-Pack targets for these MCUs, the sequence for LPC55 will be used instead of default behavior.

Ref: https://github.com/pyocd/pyOCD/blob/main/pyocd/target/family/__init__.py#L41

This causes `reset_and_halt` fails, and here are the details:

> On LPC55, the `AP#2` is the `DM-AP`, but on LPC54, this AP does exist with a different name called `ISP-AP`, which lacks the "Start Debug Session" or debug authentication related commands. Since the `DM-AP` does not exist, the MCU will not be reset since the correct sequence is never executed as the LPC55 unlock sequence has no effect on ISP-AP.

This is not easy to find by looking into the logs (as these MCUs does have `AP#2` with compatible registers to `DM-AP`), and the only warning is the forced halt after polling has timed out.
```
...
0011500 D halting core 0 [cortex_m]
0011502 W had to forcibly halt core after reset [target_lpc5500]
0011506 I Done. [reset_cmd]
...
```

imi415:


flit:
Thanks for the patch! I ended up merging #1440 that uses a slightly different regex than your patch: `LPC55.?[0-9]{2}.*` instead of `LPC55S?[0-9]{2}.*`, so it'll catch all of the LPC55xx part numbers. But it was your PR that pointed out the issue there. 😄

PR #1447, opened 2022-09-07
commands: fix typo in rmwatch

kkitayam:
Fix typo


kkitayam:
`self.size` is not a member variable of `RemoveWatchpointCommand`. So, accessing `self.size` rises a exception every time and `rmwatch` command doesn't work properly.

flit (approved):

PR #1449, opened 2022-09-07
coresight: Fix JTAG initialization sequence for CMSIS-DAP

kkitayam:
This patch fixes JTAG with CMSIS-DAP. CMSIS-DAP must be in Run-Test/Idle state after JTAG initialization.

The PR requires #1448

### DIFF
--- a/pyocd/coresight/dap.py
+++ b/pyocd/coresight/dap.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2015-2020 Arm Limited
 # Copyright (c) 2021 Chris Reed
 # Copyright (c) 2022 Clay McClure
+# Copyright (c) 2022 Toshiba Electronic Devices & Storage Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


kkitayam:
Current version pyOCD doesn't work with JTAG on CMSIS-DAP. Because pyOCD set JTAG state to `Test-Logic-Reset` after initialized. 

We can see that [JTAG_Transfer](https://github.com/ARM-software/CMSIS_5/blob/51263182d16c92649a48144ba56c0945f9fce60e/CMSIS/DAP/Firmware/Source/JTAG_DP.c#L167) (DAP_Transfer invokes the function) sets JTAG state to `Select-DR-Scan` at first. It seems that CMSIS-DAP expects that JTAG state is in `Run-Test/Idle` after initialization.
See also [arm document](https://developer.arm.com/documentation/ddi0201/d/debug-support/the-jtag-state-machine). It is shown that all JTAG states.

This PR adds a state transition to `Run-Test/Idle`. I had confirmed that this PR fixes the issue and JTAG works with our board.

flit (approved):

flit:
Thanks a lot for these JTAG fixes! 😄

PR #1462, opened 2022-10-25
align second double buffer to size of one sector

ttt-t3r:
The MAX32660 has a flash sector size of 8kB but the flash RAM buffer for double buffering assumes each buffer to be 1kB.
During flash writing, the first sector has incorrect data starting from 0x400 (1kB), comparing with the source file this incorrect data matches data that should be at 0x2000.
All MAX32660 variants should have at least 32kB so I think there shouldn't be a problem increasing the RAM size.

### DIFF
--- a/pyocd/target/builtin/target_MAX32660.py
+++ b/pyocd/target/builtin/target_MAX32660.py
@@ -58,7 +58,7 @@ FLASH_ALGO = {
     'page_size' : 0x400,
     'analyzer_supported' : False,
     'analyzer_address' : 0x00000000,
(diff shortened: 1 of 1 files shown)


flit:
Interesting, I guess flashing failed?

flit (approved):

ttt-t3r:
> Interesting, I guess flashing failed?

yes, the program was hardfault-ing before reaching main(). I figured out that flashing worked if I forced `enable_double_buffering` to False in builder.py and traced the issue to a potential problem in the flash buffer size.

PR #1469, opened 2022-11-06
partial support for SAML11D16A

ttt-t3r:
partial support for Microchip/Atmel SAML11D16A

SAML11 devices contain security features that block debugger access according to level (DAL).
To reset DAL level a chip erase command must be issued with the correct key. 

* Added new target
* Add svd file from cmsis-pack
* Add support for mass erase when
   - key is accessible before mass erase command
   - key is the default value
* Override reset_and_halt function


flit:
Thanks for the addition!

Regarding the assert-reset-with-clock-low API: On my [feature/debug_sequences](https://github.com/flit/pyOCD/tree/feature/debug_sequences) branch, there is a new debug probe API that provides full pin control (required for the `DAP_SWJ_Pins()` debug sequence function). Since this is a generic API available for all probes, currently implemented for CMSIS-DAP and JLink, it would be best to rebase this PR after the new API is merged in. I'll extract those changes into a separate PR so they can be merged before the rest of the debug sequence support is finalised (it's in final test and debug).

Fwiw, the SAML11 CMSIS-Pack does provide debug sequences for handling the security states and special flash programming requirements. In fact, the standard flash programming sequences were added specifically to support the SAML11. However, pyocd won't support the flash programming sequences in the first release of debug sequence support, so this PR will still be required.

flit:
@ttt-t3r Please see #1478 for the new debug probe pins API. If you have any issue with it or suggestions for improvement, please let me know!

ttt-t3r:
> 

I had a feeling the upcoming debug_sequences feature would allow support for debugging a wider range of MCUs, but I had an immediate need due to development of a client board so I went ahead and added the necessary code. 

I haven't looked at the flash sequences support in the new branch yet but I'd like to note that the flash binary included in the CMSIS-Pack has different function entry point names, and the python conversion script included in pyocd was unable to find the required flash routines. The flash binary in this pull request is actually newly written.

ttt-t3r:
> @ttt-t3r Please see #1478 for the new debug probe pins API. If you have any issue with it or suggestions for improvement, please let me know!

Thank you for the PR, I will take a look and try to rebase this PR using the new API!

flit:
Thanks!

Issue #1444, opened 2022-08-19
stlink_probe.read_memory_block8 returning extra bytes on un-aligned 1 byte read


pdietl:
It seems to me that commit 46f1ac66c26169270b6c6e3d7992f3b106a8077b introduced a bug. The function stlink_probe.read_memory_block8 will return an array with two bytes in it if the addressed is aligned on a 2-byte boundary but a read of 1 byte (size=1) is requested. This ends up causing an assertion at `/memory.py(263)read_memory_block8()`: `0107462 E Unhandled exception in handle_message (b'v'): result size (2) != requested size (1) [gdbserver]`

I assume we need to add something along the lines of `array[0]` somewhere, though I am not sure about the appropriate place. 

@flit

flit:
Thanks for the detailed report! (And very sorry it took a long time.) Fortunately, it looks like this was fixed in #1428 and released in v0.34.1. If you still have a problem, please reopen this issue.

DariusBabrauskas:
v0.34.2 still have same problem. #1428  not fix this problem.  But error in same place.  Error occurs if (address & 3) >1   and read size 1 or 2.

PR #1476, opened 2022-11-23
probe: stlink: Fix 1-byte transfers

fkjagodzinski:
Internally the 1-byte transfers are handled in 3 phases:
1. read/write 8-bit chunks from every unaligned addresses until the first aligned address is reached,
2. read/write 32-bit chunks from all aligned addresses,
2. read/write 8-bit chunks from the remaining unaligned addresses.

Size of the first unaligned read/write is set to the result of address (4-byte) alignment check and can be either 1, 2, or 3 bytes (the value of `unaligned_count` calculated as `addr & 0x3`). This is incorrect and every transfer with the requested size smaller than `unaligned_count` is terminated with the following error:

> Unhandled exception in handle_message (b'm'): result size (3) != requested size (1) [gdbserver]

---

Skip the first unaligned transfer if the requested size is smaller than the result of the address alignment check. Handle the whole request in the second, unaligned read/write (phase 3. mentioned above).

### DIFF
--- a/pyocd/probe/stlink_probe.py
+++ b/pyocd/probe/stlink_probe.py
@@ -1,5 +1,5 @@
 # pyOCD debugger
-# Copyright (c) 2018-2020 Arm Limited
+# Copyright (c) 2018-2020,2022 Arm Limited
 # Copyright (c) 2021-2022 Chris Reed
 # SPDX-License-Identifier: Apache-2.0
 #
@@ -328,21 +328,28 @@ class STLinkMemoryInterface(MemoryInterface):
         csw = attrs.get('csw', 0)
         res = []
 
-        # read leading unaligned bytes
-        unaligned_count = addr & 3
-        if (size > 0) and (unaligned_count > 0):
+        # Transfers are handled in 3 phases:
+        #   1. read 8-bit chunks until the first aligned address is reached,
+        #   2. read 32-bit chunks from all aligned addresses,
+        #   3. read 8-bit chunks from the remaining unaligned addresses.
+        # If the requested size is so small that phase-1 would not even reach
+        # aligned address, go straight to phase-3.
+
+        # 1. read leading unaligned bytes
+        unaligned_count = 3 & (4 - addr)
+        if (size > unaligned_count > 0):
             res += self._link.read_mem8(addr, unaligned_count, self._apsel, csw)
             size -= unaligned_count
             addr += unaligned_count
 
-        # read aligned block of 32 bits
+        # 2. read aligned block of 32 bits
         if (size >= 4):
             aligned_size = size & ~3
             res += self._link.read_mem32(addr, aligned_size, self._apsel, csw)
             size -= aligned_size
             addr += aligned_size
 
-        # read trailing unaligned bytes
+        # 3. read trailing unaligned bytes
         if (size > 0):
             res += self._link.read_mem8(addr, size, self._apsel, csw)
 
@@ -355,8 +362,8 @@ class STLinkMemoryInterface(MemoryInterface):
         idx = 0
 
         # write leading unaligned bytes
-        unaligned_count = addr & 3
-        if (size > 0) and (unaligned_count > 0):
+        unaligned_count = 3 & (4 - addr)
+        if (size > unaligned_count > 0):
             self._link.write_mem8(addr, data[:unaligned_count], self._apsel, csw)
             size -= unaligned_count
             addr += unaligned_count


fkjagodzinski:
The issue was discovered in the latest release of pyOCD (`0.34.2`) when debugging `stm32u585aiix` target, and can be exposed e.g.

```
(gdb) x/128xb 0x20000000
0x20000000 <foo>:	0x00	0x00	Cannot access memory at address 0x20000002
(gdb) 
```

Matching pyocd log and traceback:

```
0155554 E Unhandled exception in handle_message (b'm'): result size (2) != requested size (1) [gdbserver]
  File "/usr/local/lib/python3.10/site-packages/pyocd/gdbserver/gdbserver.py", line 419, in handle_message
  File "/usr/local/lib/python3.10/site-packages/pyocd/gdbserver/gdbserver.py", line 840, in get_memory
    mem = self.target_context.read_memory_block8(addr, length)
  File "/usr/local/lib/python3.10/site-packages/pyocd/debug/cache.py", line 42, in read_memory_block8
    return self._memcache.read_memory_block8(addr, size)
  File "/usr/local/lib/python3.10/site-packages/pyocd/cache/memory.py", line 258, in read_memory_block8
    assert len(result) == size, "result size ({}) != requested size ({})".format(len(result), size)
AssertionError: result size (2) != requested size (1)
```

When inspected with pdb, this leads to `read_memory_block8`:

```
(Pdb) bt
  /usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py(973)_bootstrap()
-> self._bootstrap_inner()
  /usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py(1016)_bootstrap_inner()
-> self.run()
  <local-pyocd-source>/pyocd/gdbserver/gdbserver.py(333)run()
-> self._run_connection()
  <local-pyocd-source>/pyocd/gdbserver/gdbserver.py(385)_run_connection()
-> resp = self.handle_message(packet)
  <local-pyocd-source>/pyocd/gdbserver/gdbserver.py(422)handle_message()
-> reply = handler(msg[msgStart:])
  <local-pyocd-source>/pyocd/gdbserver/gdbserver.py(840)get_memory()
-> mem = self.target_context.read_memory_block8(addr, length)
  <local-pyocd-source>/pyocd/debug/cache.py(42)read_memory_block8()
-> return self._memcache.read_memory_block8(addr, size)
  <local-pyocd-source>/pyocd/cache/memory.py(255)read_memory_block8()
-> combined = self._read(addr, size)
  <local-pyocd-source>/pyocd/cache/memory.py(137)_read()
-> uncachedData = self._read_uncached(uncached)
  <local-pyocd-source>/pyocd/cache/memory.py(97)_read_uncached()
-> data = self._context.read_memory_block8(uncachedIv.begin, uncachedIv.end - uncachedIv.begin)
  <local-pyocd-source>/pyocd/coresight/cortex_m.py(493)read_memory_block8()
-> data = self.ap.read_memory_block8(addr, size)
  <local-pyocd-source>/pyocd/utility/concurrency.py(29)_locking()
-> return func(self, *args, **kwargs)
  <local-pyocd-source>/pyocd/coresight/ap.py(1236)_accelerated_read_memory_block8()
-> return self._accelerated_memory_interface.read_memory_block8(addr, size,
> <local-pyocd-source>/pyocd/probe/stlink_probe.py(304)read_memory_block8()
-> addr &= 0xffffffff
(Pdb) 
```

fkjagodzinski:
Related (or pretty much the same 😉 ) issue -- #1444.

flit:
Thanks! And helpful detail in the description.

I'm about to push a change to remove Python 3.6 from the CI workflow, so it will run successfully.

DariusBabrauskas:
I think this commit not correct too. See fix at https://github.com/pyocd/pyOCD/pull/1428 in comments.

pyocd/probe/stlink_probe.py
```diff
@@ -329,8 +329,8 @@ def read_memory_block8(self, addr: int, size: int, **attrs: Any) -> Sequence[int
         res = []
 
         # read leading unaligned bytes
-        unaligned_count = addr & 3
-        if (size > 0) and (unaligned_count > 0):
+        unaligned_count = 3 & (4 - (addr & 3))
+        if (size > unaligned_count > 0):
```
flit, line 333: This still needs to read the unaligned bytes even if `unaligned_count` is less than or equal to `size`. So this should be:

```py
unaligned_count = min(size, 3 & (4 - addr))
if unaligned_count > 0:
```

(The `addr & 3` mask is unnecessary given how integers and bit masking works in Python.)

flit (changes requested):

pyocd/probe/stlink_probe.py
```diff
@@ -355,8 +355,8 @@ def write_memory_block8(self, addr: int, data: Sequence[int], **attrs: Any) -> N
         idx = 0
 
         # write leading unaligned bytes
-        unaligned_count = addr & 3
-        if (size > 0) and (unaligned_count > 0):
+        unaligned_count = 3 & (4 - (addr & 3))
+        if (size > unaligned_count > 0):
```
flit, line 359: Same as above.

pyocd/probe/stlink_probe.py
DariusBabrauskas: It looks fine.

pyocd/probe/stlink_probe.py
DariusBabrauskas: What... if size 1 and unaligned_count 2?
if (1> 2 > 0):
      # not exe

flit: This section should have the same change as requested above:

```py
unaligned_count = min(size, 3 & (4 - addr))
if unaligned_count > 0:
```

pyocd/probe/stlink_probe.py
fkjagodzinski: Thanks for the comments. 👍

I do agree the unaligned read still has to happen, but not necessarily here (in case of a small `size` value).

Please note that  `size <= unaligned_count` also means `size < 4`, so the second if statement is also skipped making the very last read (line 347) handle the whole (short) request. Please see the commit message:

> Skip the first unaligned transfer if the requested size is smaller than
the result of the address alignment check. Handle the whole request in
the second, unaligned read/write.

fkjagodzinski: If you think the request should be always handled within the first conditional, or that https://github.com/pyocd/pyOCD/pull/1476/commits/a353c96745559105dddf711bd489057da55a1c88#r1032679727 code is more readable, please let me know, so I can amend the patch this way.

flit: Ooh, totally missed that, really sorry!! Thanks for explaining it. (And I didn't see the commit message since GH collapses all but the commit subject by default.)

It would probably be best to have a comment describing how this works before this section of code.
Whether you keep the (strictly unnecessary) `addr & 3` bit is your choice.

fkjagodzinski: Done; I've removed the redundant `& 3` operation and added a comment in the code.

flit (approved):
Looks good! Thanks for the added comments.

Just noticed: you should probably add your copyright to the file header comment (meaning I'd prefer it 😄). I'll give you a chance before merging.

fkjagodzinski:
> Just noticed: you should probably add your copyright to the file header comment (meaning I'd prefer it 😄). I'll give you a chance before merging.

OK, updated the copyright. Too bad this cancelled the whole CI pipeline. 😅

flit:
Thanks!

flit:
Btw, I plan to cherry pick this fix to `main` and release a bug fix update quickly. Thanks again.

PR #1492, opened 2023-01-05
flash: Add option to prevent resetting of device after programming

nordicjm:
This adds a --no-reset command line argument that can be used to prevent a target being reset after a program operation has finished. The use-case for such a scenario includes when programming multiple images whereby one of the images enables security such as disabling the debugger, without this patch, only the first image would be flashed.

### DIFF
--- a/pyocd/flash/builder.py
+++ b/pyocd/flash/builder.py
@@ -394,7 +394,7 @@ class FlashBuilder(MemoryBuilder):
                 page = add_page_with_existing_data()
                 sector_page_addr += page.size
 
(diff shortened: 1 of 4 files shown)


flit:
Awesome! This was actually on my todo list. 😄

flit (approved):

flit:
Fyi, just noticed that this PR is targeting `main`. Because it's a feature addition, it should be targeted at `develop`. I'll merge it, cherry pick to `develop`, then delete the commit from `main` so you don't have to worry about it.

PR #1503, opened 2023-02-22
[Dependencies] Bump Pylink

avi-jois:
Pylink 1.0 has the following breaking change:

@denravonska: Changed .flash() to no longer erase chip on flash; users will now need to ensure they call .erase() prior to flashing a non-erased region of flash.

However, pyOCD doesn't use JLinks's flash routines so there is no call in the codebase to `.flash()` from `pylink`. This makes it safe to allow `pylink` packages with a major version of 1.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,7 +58,7 @@ install_requires =
     natsort>=8.0.0,<9.0
     prettytable>=2.0,<4.0
     pyelftools<1.0
-    pylink-square>=0.11.1,<1.0
+    pylink-square>=0.11.1,<2.0
     pyusb>=1.2.1,<2.0
     pyyaml>=6.0,<7.0
     six>=1.15.0,<2.0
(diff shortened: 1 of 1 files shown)


denravonska:
Worth noting is that JLink itself will erase the affected pages, at least on the devices I've tried it on, so an erase will happen. Just not on a chip level :)

flit:
Do you think we should bump the minimum pylink version too?

avi-jois:
I don't think it's necessary to bump the minimum version because there's no new feature that's needed, and keeping as wide a dependency range as possible minimizes possible dependency resolution conflicts.

flit (approved):

flit:
@avi-jois Do you mind if I rebase your branch so this can be merged? Thanks.

avi-jois:
I don’t mind, feel free to rebase and merge it.

flit:
Fyi, I bumped the minimum pylink-square version to 1.0. As long as that package follows semantic versioning, this should ensure API compatibility until version 2.0.

PR #1505, opened 2023-02-28
probe: cmsis-dap: set retry count to max value by default

maxd-nordic:
This patch sets the retry count for DAP transfers to the highest value by default.
This is needed since flash erase will stall a read operation for a long time on the nRF9160.

### DIFF
--- a/pyocd/probe/pydapaccess/cmsis_dap_core.py
+++ b/pyocd/probe/pydapaccess/cmsis_dap_core.py
@@ -294,7 +294,7 @@ class CMSISDAPProtocol(object):
 
         return resp[1]
 
-    def transfer_configure(self, idle_cycles=0x02, wait_retry=0x0050, match_retry=0x0000):
+    def transfer_configure(self, idle_cycles=0x02, wait_retry=0xFFFF, match_retry=0x0000):
         cmd = []
         cmd.append(Command.DAP_TRANSFER_CONFIGURE)
         cmd.append(idle_cycles)


maxd-nordic:
I think it should be no problem to have a very high retry value. Since the target still has to send ACK_WAIT responses, we still return failure on broken connections.

flit:
👋🏽 Thanks for the patch!

Do you happen to know the number of retries that are required on average? Mostly just curious, but it would also be good to know how close it is to the maximum. If it's too close, then pyocd should probably be performing its own retries in addition to the probe (that just hasn't been necessary in any other cases).

flit:
Yes, I agree that it shouldn't be a problem. But actually, the retry count should be configurable by the user and target. (Hmm, I'll add a "minimum required retry count" to my [Open-]CMSIS-Pack todo list.)

maxd-nordic:
> 👋🏽 Thanks for the patch!
> 
> Do you happen to know the number of retries that are required on average? Mostly just curious, but it would also be good to know how close it is to the maximum. If it's too close, then pyocd should probably be performing its own retries in addition to the probe (that just hasn't been necessary in any other cases).

That of course depends on the bus speed, but I saw 6232 retries on a 1MHz SWD connection and that was not even the slowest operation.

flit:
Ok, thanks. So with 6232 at 1 MHz, it should be safe to go up to somewhat faster than 10 MHz using a retry count of 64k. Probably alright for now. And one can always use a slower SWD clock if a chip erase is needed.

l0ud:
fyi: this change happened to fix Puya MCU flash programming which also seems to take a long time to erase. Got timeout errors before.

Issue #1515, opened 2023-03-16
Semihosting and STD*_FILENO


rgrr:
Hello,

when using semihosting and trying to send a message to stdout via `write(STDOUT_FILENO,...)`, output goes to nowhere under pyocd, openocd does the expected output.

This is because "STDOUT_FILENO=1" according to C standards but pyocd expects it to be "2".  The sources (https://github.com/pyocd/pyOCD/blob/main/pyocd/debug/semihost.py#64) shows "STDOUT_FD=2" .

If one uses "2" on the target, the output appears.  In the sources mentioned, it is stated that "The fds must be non-zero according to the ARM semihosting spec".  (I've searched the spec but cannot find it)

Now I'm wondering who is mistaken?  C lib defines "STDOUT_FILENO=1" for ages and this is certainly an incompatibility.

rgrr:
Perhaps I have found the place in the spec: https://developer.arm.com/documentation/dui0471/g/Semihosting/SYS-OPEN--0x01-?lang=en

It states

```
Return

On exit, R0 contains:

    a nonzero handle if the call is successful

    -1 if the call is not successful.
```

But writes to stdout do not need any open().  So the spec is wrong as well at https://developer.arm.com/documentation/dui0471/g/Semihosting/SYS-WRITE--0x05-?lang=en when it says:

```
Entry

On entry, R1 contains a pointer to a three-word data block:

word 1

    contains a handle for a file previously opened with SYS_OPEN
```

Because, as said, stdout/stderr don't have to be opened.

For compatibility with the C library correct behavior on pyocd side would be to use the same values for the std* handles as the C lib does.

flit:
Newlib definitely does the expected behaviour in regards to opening stdio prior to its use, eg opening a file called ":tt" with various modes. Hopefully it's possible to support both behaviours.

rgrr:
Hmmm... I guess the spec of the C library is a little bit older than the Arm semihosting spec.  The values of the special file descriptors for stdio made it even into wikipedia, see https://en.wikipedia.org/wiki/File_descriptor.

Even the comment in https://github.com/pyocd/pyOCD/blob/main/pyocd/debug/semihost.py#105 says, that the values are 0/1/2.  But effectively they are 1/2/3, which is clearly wrong.
OpenOCD behaves as expected: "write(1, buf, sizeof(buf))" sends characters to stdout, pyOCD needs "write(2, buf, sizeof(buf))".

[...]

PR #1517, opened 2023-03-18
Semihosting stdio fds

rgrr:
Renumber the stdio file descriptors so that there is a better fit with POSIX.  OpenOCD behaves the same.

See discussion here #1515


flit (approved):

flit:
Btw, what C library are you using?

rgrr:
> Btw, what C library are you using?

I'm using picolib.  Toolchain is built with https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm

PR #1519, opened 2023-03-23
flash: catch exceptions while fetching state

maxd-nordic:
This patch catches (DAPLink) probe timeouts while waiting for a flash operation. These timeouts can occur on targets that are stalled for a long time. It also reverts my previous attempt to fix the issue, which is unnecessary now and might cause other timeouts.


pyocd/flash/flash.py
flit, line 1: Probably should add a Nordic copyright.

flit (approved):

PR #1520, opened 2023-03-24
semihosting transparent io

### LINKED ISSUE #1518: Semihosting and binary file write
I guess there is a problem with writing binary data to files via semihosting.

semihosting::_get_string() converts data to (utf-8) strings regardless if the data is expected to be a string or not.

If the binary data from the target contains ill utf-8 data, pyocd throws exceptions, e.g.

```
...
  File "/home/hardy/.local/lib/python3.11/site-packages/pyocd/debug/semihost.py", line 538, in handle_sys_write
    return self.io_handler.write(fd, data_ptr, length)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hardy/.local/lib/python3.11/site-packages/pyocd/debug/semihost.py", line 223, in write
    data = self.agent._get_string(ptr, length)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/hardy/.local/lib/python3.11/site-packages/pyocd/debug/semihost.py", line 485, in _get_string
    return bytes(data).decode()
           ^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x81 in position 0: invalid start byte
```

I will try to look into this when there is time.

rgrr:
Semihosting IO no longer does internal encode() / decode() steps because this fails on binary data, e.g. coverage dumps.

There is an additional test to verify correct behavior for binary data.

Closes #1518


rgrr:
Hello Chris,

I have to change this to draft.  **Do not merge it.**  I've noticed that there must be some mix up of str / bytes during some API games.
Test system does not seem to detect this, so I will check the test cases as well.

rgrr:
I have changed the code a little bit (actually reverted parts) and added some text mode operations to the "unit"tests.  Everything seems to work.

Nevertheless I'm not completely happy, because text mode expects utf-8 because it does some invisible encode/decode operations.  Any idea how to resolve this?  Or should we stay with this solution because binary is now transparent and text mode is the same as before.

flit:
Thanks for the change! Sorry I disappeared for a bit, I was sick for a week with some intestinal thing…

pyocd/debug/semihost.py
```diff
@@ -479,32 +479,32 @@ def _get_args(self, args, count):
         else:
             return args
 
-    def _get_string(self, ptr, length=None):
+    def _get_data(self, ptr, length=None):
```
flit, line 482: Hmm, maybe we should just separate `._get_string()` from `._get_data()`? The only difference being null-termination support; return type would be the same. Not necessary, just thinking.

And probably should remove the leading underscore since it's not actually a private method and is called from outside the class. (My bad!)

flit:
Personally, I think it's probably ok. It's certainly an improvement.

What's you're saying about text mode is pretty much was I was getting at in the second option described in #1518:

> Just ignore the text/binary mode flag, open all files as binary. Assume that the data from the target is correctly encoded / the target can handle data encoded in the file's coding. This will probably work fine. It's a bigger change to the code, however. And it would make it impossible to perform encoding translations if that were desired (and implemented).

There's likely some downside to doing it this way that I haven't considered.

I think I'd like to merge this change as-is. Any further changes can be done in another PR. What do you think?

flit (approved):

flit:
Ah, forgot something in the review: would you mind adding your copyright to the file's header comment? Thanks!

test/pytest_report.html
flit, line 1: This file probably shouldn't be in the PR?

rgrr: gosh, thank you for finding this.

PR #1488, opened 2022-12-16
cmsis-packs: debug sequence support

flit:
This PR is a major feature to add support for [CMSIS debug sequences](https://open-cmsis-pack.github.io/Open-CMSIS-Pack-Spec/main/html/debug_description.html) (from Device Family Packs, or DFPs). Support for more details and features of the DFP debug description is necessarily included. Adding debug sequence support greatly improves reliability of DFP-based targets, as it allows pyocd to use special debug logic written by the silicon vendor. Basically, it brings DFP-based targets up to par with builtin targets.

There are _lots_ of related changes that go along with the main feature, mostly to implement features of debug sequences and debug definitions.

- Graph nodes now have names. The board and target have standard names of `board` and `soc`, while cores get their names from the CPU type or DFP debug description. Not really used much currently, but will enable setting options for specific cores and similar features.
- Cores have a `.supported_reset_types` property. Attempting to use `SW_VECTRESET` or `SW_SYSRESETREQ`, if unsupported, will result in a warning log and a fallback reset type being used.
- Memory regions can have subregions. This is used to support flash regions with multiple section sizes, as well as flash regions where the flash algo only covers a subset of the entire architecturally-defined region.
- Major rework of DFP flash region handling that fixes some outstanding issues.
- New and modified commands/values:
  - `show debug-sequences`
  - `show reset-type` / `set reset-type`
  - `show map`: show memory map subregions
  - `show cores`: print core names
  - `show graph`: print node names

Debug sequence support is a new feature that can greatly change how pyocd controls the target. It is also possible for DFPs to have bugs in their debug sequences (in addition to the other bugs that DFPs may contain such as broken flash algos). As such, pyocd provides a method to disable all sequences or specific ones by name using session options:

- `pack.debug_sequences.enable`: Bool, enabled by default, that controls whether any debug sequences are executed.
- `pack.debug_sequences.disabled_sequences`: List of strings (comma-separated when set from command line), used to disable debug sequences by name. Names can optionally include a suffix with a colon and CPU core name in order to disable the given sequence for that core only.

Remaining work:

- ✅ Finish support for dbgconf variables, e.g. the definitions in `.dbgconf` files that get generated when a DFP containing debug sequences is used with Keil MDK or Keil Studio. Depending on how this is handled (not sure if `.dbgconf` files are the right solution for pyocd), there might need to be a way to display available dbgconf variable values. This might be handled in a follow-on PR.
- Finish documentation.
[...]

PR #1527, opened 2023-04-04
add logging function to rtt sub-command.

zjli-2019:
It will be handy to log data from up channel to a file, just like JLinkRTTLogger.

So, a new option `--log-file` is added to specify the file name and enable logging mode.

Other two options are also added to select up/down channels.

### DIFF
--- a/pyocd/subcommands/rtt_cmd.py
+++ b/pyocd/subcommands/rtt_cmd.py
@@ -22,6 +22,7 @@ import argparse
 import logging
 import sys
 from time import sleep
+import time
 from typing import List
 
 from pyocd.core.helpers import ConnectHelper
@@ -39,7 +40,7 @@ class RTTSubcommand(SubcommandBase):
     """@brief `pyocd rtt` subcommand."""
 
     NAMES = ["rtt"]
-    HELP = "SEGGER RTT Viewer."
+    HELP = "SEGGER RTT Viewer/Logger."
 
     @classmethod
     def get_args(cls) -> List[argparse.ArgumentParser]:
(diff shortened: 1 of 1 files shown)


rgrr:
Sorry to jump into this PR... I'm currently doing some API experiments and one of my ideas was to do some RTT console fetching in an example.  Unfortunately the current RTT implementation is not very flexible (underestimation).
How about extending the RTT module so that it exposes some methods like: SearchSignature() (perhaps in smaller bunches) and FetchData()?
If you already have your hands on this code, how about refactoring it a "little" bit?

zjli-2019:
@rgrr, `RTTControlBlock.from_target` in `pyocd.debug.rtt` seems just like `SearchSignature`.

rgrr:
> @rgrr, `RTTControlBlock.from_target` in `pyocd.debug.rtt` seems just like `SearchSignature`.

@zjli-2019 : perhaps we have different sources, but I cannot find something like pyocd.debug.rtt or RTTControlBlock in pyocd.  Do you have a link to a file?

zjli-2019:
@rgrr you can find it on the develop branch:

https://github.com/pyocd/pyOCD/blob/8ac1dbac77150acb22fd88617653d55e63eb5fb0/pyocd/debug/rtt.py#L120

rgrr:
> @rgrr you can find it on the develop branch:
> 
> https://github.com/pyocd/pyOCD/blob/8ac1dbac77150acb22fd88617653d55e63eb5fb0/pyocd/debug/rtt.py#L120

Thanks for pointing me into that direction.  That's exactly what I was looking for.  Question is when this will be integrated.

flit:
> Question is when this will be integrated.

I'm planning a release of everything on the `develop` branch as soon as the CMSIS-Pack debug sequence support (#1488) is merged in. The only thing remaining (I hope!) is a fix for an issue related to generation of memory regions for flash algorithms that don't have an explicit region definition, when there are certain alias regions present. Actively being worked on.

pyocd/subcommands/rtt_cmd.py
flit, line 1: File needs to end with an empty line.

flit, line 1: Please add your copyright to the file's header comment.

flit (changes requested):
Just the minor file-related requests.

On a more functional level, do you think it would make sense to have logging be independent of viewing/input? So you could simultaneously view and log, or save a log from an interactive session.

zjli-2019:
> Just the minor file-related requests.
> 
> On a more functional level, do you think it would make sense to have logging be independent of viewing/input? So you could simultaneously view and log, or save a log from an interactive session.

It is nice to be able to simultaneously view and log, or save a log from an interactive session. 
If so, file format has to be considered: how to store input/output separately, how to parse it later, etc.

In my case, an up channel is used to transfer **binary** trace data.  Logging mode
is just like JLinkRTTLogger, and all data coming from the up channel is saved to a file that will be processed later.

flit (approved):

PR #1536, opened 2023-04-16
semihosting: SYS_GET_CMDLINE, feature bits, tilde expansion, type fixes

flit:
Several small features and some cleanup of the semihosting code.

- `SYS_GET_CMDLINE` request is handled, returning the value of the `semihost.commandline` session option.
- `SYS_HEAPINFO` is stubbed to return all zeroes.
- Tilde expansion for home directories in all file paths.
- Type annotations and type fixes.
- The `:semihosting-features` special file is supported. Added a unit test case for this. Supported feature bits are:
        - `SH_EXT_EXIT_EXTENDED` (byte 0 bit 0): =0, meaning `SYS_EXIT_EXTENDED` is not supported
        - `SH_EXT_STDOUT_STDERR` (byte 0 bit 1): =1, meaning both stdout and stderr are supported

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -476,6 +476,13 @@ Whether to use GDB syscalls for semihosting file access operations, or to have p
 operations. This is most useful if GDB is running on a remote system.
 </td></tr>
 
+<tr><td>semihost.commandline</td>
+<td>str</td>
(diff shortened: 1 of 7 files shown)


flit:
cc @rgrr

PR #1539, opened 2023-04-23
Update yuntu devices.

majorlin:


scripts/YTM32B1LE0_Main.FLM
flit, line 1: This .FLM file should be removed from the PR?

PR #1561, opened 2023-05-21
core: session: load the user script after the board is constructed

flit:
It's not clear why, originally, the user script was loaded before the board is created. In any case, the namespace update on .open() is retained. Even though all the objects added to the namespace are available now when the user script is loaded, they aren't really usable since nothing is opened/connected/inited.


yyjdelete:
After this commit, is it still possible to regist a new target by userscript like https://github.com/pyocd/pyOCD/issues/981 does?

PR #1562, opened 2023-05-23
target: family: psoc6: remove unnecessary sleep during reset.

te-johan:
self.reset() have a 0.5 second sleep when using hw reset and other reset types should not require any additional sleep.
by removing this sleep it is possible to aquire targets where the application disable the swd port.


flit:
Ah, that's where that sleep was!

flit (approved):

PR #1567, opened 2023-05-30
Added support for stm32 h743 and h723

unsanded:
Added support for these two targets. It is for a big part copy-pasted from another stm32 target. So I feel like this should be put all together in `target/family`, but i don't feel like i know enough to do that.


pyocd/target/builtin/target_STM32H743xx.py
```diff
@@ -0,0 +1,370 @@
+# pyOCD debugger
+# Copyright (c) 2020 Bartek Wolowiec
+# Copyright (c) 2021 Chris Reed
```
flit, line 3: Copyright.

pyocd/target/builtin/target_STM32H743xx.py
```diff
@@ -0,0 +1,370 @@
...
+
+        #sram3
+        RamRegion(   start=0x30040000, length=0x8000,
+                  is_powered_on_boot=False),
+        #sram4
+        RamRegion(   start=0x38000000, length=0x10000),
+        )
+
+    def __init__(self, session):
+        super(STM32H743xx, self).__init__(session, self.MEMORY_MAP)
```
flit, line 230: Remove args to `super()`.

pyocd/target/builtin/target_STM32H723xx.py
```diff
@@ -0,0 +1,317 @@
...
+        self.flash_keyr = self.flashaddr + 4
+        self.flash_optkeyr = self.flashaddr + 8
+        self.flash_optcr = self.flashaddr + 0x18
+        self.flash_cr = self.flashaddr + 0xc
+        self.flash_sr = self.flashaddr + 0x10
+        self.flash_optsr_cur = self.flashaddr + 0x1c
+        self.flash_optsr_prg = self.flashaddr + 0x20
+
+
+class MiniAP(object):
```
flit, line 55: Can this `MiniAP` class be moved to a common location so there is only one copy of the code? Maybe `pyocd.coresight.minimal_mem_ap.py`?

A common implementation should mask the `IDR` with 0x0fffe00f and check against 0x04770001. This will verify it's an Arm AHB-AP of any variant or revision.

flit (changes requested):

pyocd/target/builtin/target_STM32H723xx.py
unsanded: Yeah, i will write it, but it i am not in a position to test with the F767.

flit: Well, me either. 😄 But I can verify it by inspection and hand testing.

unsanded: I aded it like suggested, and tested on the two H7 devices. For the F7xx, it turns out that the exact MiniAP from the F767 worked on the H7 devices as well. So i used that. But I did add the `&` like you suggested.

flit:
Ah, yes, please rebase onto `develop`. Thanks!

pyocd/target/builtin/target_STM32H723xx.py
flit: Thanks. The mask just ensures it will work on different and future devices without needing any changes.

flit:
Thanks!

PR #1578, opened 2023-06-30
nRF91 family improvements

maxd-nordic:
This patch adds the following improvements to nRF91 support:
* UICR flashing
* unlocking newer models with hardened APPROTECT
* built-in flash algos


flit:
Thanks for the improvements. 😄 I've been on holiday but will be back tomorrow and will start reviewing in earnest.

maxd-nordic:
Do you know if the builtin algos can be automatically applied to the nrf9160_xxaa target?

maxd-nordic:
The algorithm here is modified slightly to work with pyocd. Since the UICR region is not present i the pack-based target, we could add it as an additional memory region there. How would I do that?

flit:
I see. To add the region, you can just call `self.memory_map.add_region()` (or `.add_regions()`) in your family class and pass it a new `FlashRegion` instance. It really can be done at any point prior to discovery. But the easiest is probably from the `.__init__()` method. Just add the region into the provided `memory_map` argument before it's passed to `super().__init__()`.

Similarly, you can modify the memory map at this point to override the flash algo. (The ability to modify memory region attributes after creation is a recent change. If you run into trouble, just let me know!)

maxd-nordic:
@flit Thanks! Just made the change to automatically add the UICR region if not present already.

pyocd/target/family/target_nRF91.py
```diff
@@ -108,14 +302,104 @@ def split_addr_range_into_chunks(range, chunk_size):
...
+        if memory_map.get_region_for_address(0x00ff8000) is None:
+            LOG.debug("Adding UICR region")
+            memory_map.add_region(FlashRegion(
+            start=0x00ff8000,
+            length=0x1000,
+            blocksize=0x1000,
+            is_testable=False,
+            is_erasable=False,
+            algo=FLASH_ALGO_UICR,
+        ))
```
flit, line 400: The indentation on this section looks incorrect. The closing parentheses are a level above the indent of the `if` block. Although it will parse ok as is, it should be fixed.

flit (changes requested):

flit:
Btw, I meant to mention earlier, it would be nice to add a command to perform the modem firmware update since you already have the code. Someone was asking how to do this with pyocd on the Nordic forum. Doesn't have to be in this PR, though.

To do this, create a command class like those in `pyocd/commands/commands.py`.

Starting from something like this:

```py
class nRF91ModemFirmwareUpdateCommand(CommandBase):
    INFO = {
            'names': ['nrf91-update-modem-fw'],
            'group': 'nrf91', # This is the group named passed to `.add_command_group()`.
            'category': 'nrf', # Could also be 'nrf91'? This is the category the command will be grouped under in help.
            'nargs': 1, # Set based on usage
            'usage': "FILL ME IN",
            'help': "Update modem firmware for an nRF91 target.",
            'extra_help': "",
            }

    def parse(self, args):
        pass

    def execute(self):
        pass
```

Then override `.add_target_command_groups()` on the family class.

```py
    def add_target_command_groups(self, command_set: CommandSet) -> None:
        command_set.add_command_group('nrf91')
```

As long as the command class is in a module loaded before `add_target_command_groups()` is called, it will be found automatically when you call `command_set.add_command_group()`.

maxd-nordic:
That's great! I just added a command like you described.

flit:
Awesome! 😄

flit (approved):

PR #1583, opened 2023-07-10
Fix/1553 rtt cb not found

tdasika:
pyocd is always searching for control_block_id with 16 bytes length (when the block id is less than 16 bytes, it is padding with ‘\0’)
So, when ‘SEGGER RTT’ is the block id to be found, it is searching for 4 words (16 bytes) with zero padding instead of 10 bytes(’SEGGER RTT’),  where some junk characters may be present and resulting in failure ‘Control block not found'

Fixing it by searching for only the length of the control block ID provided while setting up rtt

### DIFF
--- a/pyocd/debug/rtt.py
+++ b/pyocd/debug/rtt.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2021 Simon D. Levy <simon.d.levy@gmail.com>
 # Copyright (C) 2022 Johan Carlsson <johan.carlsson@teenage.engineering>
 # Copyright (c) 2022 Samuel Dewan
+# Copyright (C) 2023 Tejaswini Dasika <tejaswinidasika@gmail.com>
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -387,7 +388,7 @@ class GenericRTTControlBlock(RTTControlBlock):
 
     target: SoCTarget
     _cb_search_address: int
-    _cb_search_size_words: int
+    _cb_search_size_bytes: int
     _control_block_id: Sequence[int]
 
     def __init__(self, target: SoCTarget, address: int = None,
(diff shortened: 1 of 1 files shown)


pyocd/debug/rtt.py
```diff
@@ -436,20 +433,20 @@ def _find_control_block(self) -> Optional[int]:
         offset: int = 0
 
         while search_size:
-            read_size = min(search_size, 32)
-            data = self.target.read_memory_block32(addr, read_size)
+            read_size = min(search_size, 8)
```
flit, line 436: Reducing the read size will slow the search down noticeably. 
It's probably best to leave this at 32 for now.

(Really, the read size should be at least 512 or 1024 bytes for best performance, especially with HS USB probes. However, with large sizes like that you need to take into account the greater likelihood of running off the end of the memory region when not using the default RAM region.)

pyocd/debug/rtt.py
```diff
@@ -421,10 +421,7 @@ def __init__(self, target: SoCTarget, address: int = None,
             self._cb_search_size_words = 0
         else:
             self._cb_search_size_words = size // 4
```
flit, line 423: `self._cb_search_size_words` is set to the size in words here, but it's used as a byte count for `.read_memory_block8()` in the search routine.

So, please remove the divide by 4 and rename the attribute to `_cb_search_size_bytes`.

flit (changes requested):
Just a couple small changes.

If you feel like it, please also add your copyright to the file's header comment.

flit:
@tdasika Do you mind if I commit the requested changes to your branch so I can close out this PR? I'd like to make sure this fix is included in the next release. Thanks!

tdasika:
> @tdasika Do you mind if I commit the requested changes to your branch so I can close out this PR? I'd like to make sure this fix is included in the next release. Thanks!

@flit , I have addressed the requested changes.

flit:
> I have addressed the requested changes.

Even better! Thanks 😄

flit (approved):

PR #1592, opened 2023-07-22
Add Trace Funnel Coresight component

rapgenic:
In this PR I implemented a simple graph node for coresight trace funnels.

The `TraceFunnel` component is quite simple, it reads the number of available channels and enables them all at initialization. The rationale of this choice is that I think that in most of the cases the user wants everything to be working since the beginning.

Channels can be enabled and disabled with the `set_enable_channels` function, which provides a warning if trying to set more channels than available.

This is `show graph` output with this patch:

```
pyocd> show graph
- "board": Board
  - "soc": Mimxrt1176dvmaa
    - "cm7": CortexM
      - DWT
      - FPB
      - ITM
      - TraceFunnel
    - "cm4": CortexM
      - DWT
      - FPB
      - ITM
      - TraceFunnel
```

### DIFF
--- a/pyocd/coresight/funnel.py
+++ b/pyocd/coresight/funnel.py
@@ -34,21 +34,21 @@ class TraceFunnel(CoreSightComponent):
 
     def __init__(self, ap, cmpid=None, addr=None):
         """@brief Standard CoreSight component constructor."""
-        super(TraceFunnel, self).__init__(ap, cmpid, addr)
+        super().__init__(ap, cmpid, addr)
         self._available_channels = 2
 
     @property
-    def available_channels(self):
+    def available_channels(self) -> int:
         """@brief Number of input ports connected to the funnel"""
         return self._available_channels
 
-    def init(self):
+    def init(self) -> None:
         """@brief Reads Funnel connected channels and enables them all by default."""
         devid = self.ap.read32(self.address + TraceFunnel.DEVID)
         self._available_channels = devid & TraceFunnel.DEVID_PORTCOUNT_MASK
         self.enable()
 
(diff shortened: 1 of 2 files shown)


pyocd/coresight/funnel.py
```diff
@@ -0,0 +1,75 @@
...
+    # The addresses are offsets from the base address.
+    CSTF = 0x00000000
+    CSTF_ENSX_MASK = 0xFF
+
+    DEVID = 0x00000FC8
+    DEVID_PORTCOUNT_MASK = 0xF
+
+    def __init__(self, ap, cmpid=None, addr=None):
+        """@brief Standard CoreSight component constructor."""
+        super(TraceFunnel, self).__init__(ap, cmpid, addr)
```
flit, line 37: Although it works fine, you technically don't need the arguments to `super()` since Python 3. Though there are places in pyocd where that hasn't been cleaned up yet. Not a big deal.

pyocd/coresight/funnel.py
flit, line 1: Could you please add a blank line at the end of this file?

flit, line 1: If you have time, it would be nice to have type annotations on the methods.

flit (changes requested):
Mostly minor requests. The blank line on the end of `funnel.py` is the only required change. 

Thanks for the patch!

flit:
Sorry, I just realised… would you mind rebasing and retargeting this to the `develop` branch? Only fixes are allowed on `main`. After rebasing, you can edit this PR to modify the merge target branch.

flit (approved):

PR #1595, opened 2023-07-23
probe: cmsis-dap: usb backends: windows performance fix and more

flit:
- Fix performance regression for CMSIS-DAPv2 on certain machines running Microsoft Windows by using a real multithreaded queue with blocking support for passing data from receive thread.
- Re-raise exceptions that occur in the receive thread from .read().
- A few other small improvements.


flit:
cc @Guozhanxin If you have a chance to test this patch, that would be very helpful. Thanks!

PR #1608, opened 2023-08-10
Allow FlashBuilder to work when program page size is larger than sector erase size.

BrianPugh:
# Problem Description
Currently, `FlashBuilder`, used by `FileProgrammer`, makes the reasonable assertion that sector size must be larger than program size. However, on the `STM32H7B0`, the sector size is 8KB, while the program size is 32KB. Here's the relevant definition from `Keil.STM32H7xx_DFP.3.1.1.pack`.

```
struct FlashDevice const FlashDevice  =  {
   FLASH_DRV_VERS,             // Driver Version, do not modify!
   "STM32H7B0_Flash_128k",     // Device Name 
   ONCHIP,                     // Device Type
   0x08000000,                 // Device Start Address
   0x00020000,                 // Device Size in Bytes (128kB)
   0x8000,                       // Programming Page Size
   0,                          // Reserved, must be 0
   0xFF,                       // Initial Content of Erased Memory
   100,                        // Program Page Timeout 100 mSec
   6000,                       // Erase Sector Timeout 6000 mSec

// Specify Size and Address of Sectors
   0x2000, 0x000000,          // Sector Size  8kB (16 Sectors)
   SECTOR_END
};
```

# Solution
The goal of this PR is to allow the STM32H7B0 to be programmed, while making the smallest change possible to pyOCD. This is my first PR and don't understand the code base super well, so super open to feedback and if we would like to address this in a different manner.

I added a new field, `n_subsectors` to the `_FlashSector` class. This allows the `_FlashSector` to represent an integer number of actual sectors (defaults to 1). The attribute `size` is now a property thats the total size of all subsectors. A new property, `addrs`, represents all the addresses of the subsectors. Upon `_FlashSector` creation, we first check to see if `_FlashSector` is smaller than `_FlashPage`, and if so, assign a >1 number of subsectors.

Finally, when `FlashBuilder` erases sectors, we now erase all subsectors of each `_FlashSector` object.

### DIFF
--- a/pyocd/flash/builder.py
+++ b/pyocd/flash/builder.py
@@ -937,7 +937,7 @@ class FlashBuilder(MemoryBuilder):
             for sector in self.sector_list:
                 if sector.are_any_pages_not_same():
                     # Erase the sector
-                    for addr in self.sector.addrs:
+                    for addr in sector.addrs:
                         self.flash.erase_sector(addr)
 
                     # Update progress
--- a/pyocd/flash/builder.py
+++ b/pyocd/flash/builder.py
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2015-2019 Arm Limited
 # Copyright (c) 2021 Chris Reed
+# Copyright (c) 2023 Brian Pugh
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/pyocd/flash/builder.py
+++ b/pyocd/flash/builder.py
@@ -94,12 +94,22 @@ def _stub_progress(percent):
 
 class _FlashSector:
     """@brief Info about an erase sector and all pages to be programmed within it."""
-    def __init__(self, sector_info):
+    def __init__(self, sector_info, n_subsectors: int = 1):
         self.addr: int = sector_info.base_addr
-        self.size: int = sector_info.size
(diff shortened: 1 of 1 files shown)


flit:
Super big thanks for the patch! I'll have time to review this maybe tomorrow or the weekend at the latest.

BrianPugh:
I'd say (as you alluded to on slack) that biggest current limitation of this PR is that the final erase may be oversized. I.e. if attempting to program a 100KB firmware, 128KB will be erased instead of the ideal 104KB. Not sure what the best way to tackle that would be.

BrianPugh:
other things to consider: I didn't really look how the progress weight is calculated; it may need to be adjusted.

flit:
The solution to oversized erasing is to shrink the number of sectors erased on the last page a contiguous range of programmed memory to cover only the amount of data actually programmed. Eg, if programming the first 100kb of flash, only erase the sector covering 96k-100k (and all others before it of course).

This is the method used by the Arm Keil tools. Like I described in Slack, it has potential issues… However, since basically all vendors tests their flash algos with Keil MDK, it's safe enough to match that tool's behaviour. And there's no chance of problems if the page size < sector size.

flit (approved):

flit:
Would you mind adding your copyright to the `flash_builder.py` file?

BrianPugh:
added!

PR #1610, opened 2023-08-11
Add STM32H7B0 Target

BrianPugh:
Adds the STM32H7B0 target with the following features:
* Extended the flash definition for Bank 1 to 256KB.
* Added the undocumented Bank 2 (256KB).
* Added custom modified flash algorithm for supporting 256KB.

I've tested erasing and flashing >128KB firmware to both Bank 1 and Bank 2 with success.

Builds on-top of #1608, so that should be merged first.

Possible TODOs that I need help on:
1. Maybe detect silicon revision and display appropriate warning if an unknown revision is detected.

### DIFF
--- a/pyocd/target/builtin/__init__.py
+++ b/pyocd/target/builtin/__init__.py
@@ -131,6 +131,7 @@ from . import target_ytm32b1me0
 from . import target_ytm32b1md1
 from . import target_STM32H723xx
 from . import target_STM32H743xx
+from . import target_STM32H7B0xx
 from . import target_Air001
 
 ## @brief Dictionary of all builtin targets.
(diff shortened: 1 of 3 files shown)


pyocd/target/builtin/__init__.py
```diff
@@ -216,6 +217,7 @@
           'stm32l031x6' : target_STM32L031x6.STM32L031x6,
           'stm32h723xx' : target_STM32H723xx.STM32H723xx,
           'stm32h743xx' : target_STM32H743xx.STM32H743xx,
+          'stm32h7b0' : target_STM32H7B0.STM32H7B0,
```
flit, line 220: Would you mind adding xx's to the target name so it becomes `stm32h7b0xx`? (Granted the F051 is not like that, but that's historical…)

flit (approved):

flit (changes requested):
Only change request is the target name.

BrianPugh:
updated!

flit (approved):

flit:
Oops, looks like you'll have to rebase after that previous PR was merged.

BrianPugh:
rebased; I also made the  `stm32h7b0` -> `stm32h7b0xx` name change everywhere.

PR #1614, opened 2023-08-23
very basic implementation to get a "connect" LED status display

rgrr:
subject says it all...


flit (approved):

flit (changes requested):
Thanks for the work!
Would you mind retargeting this to the `develop` branch?

(Honestly… I think I'm going to change the role of the `develop` branch soon to be only for the next major version. It's been more of a pain than it's worth for minor version releases. 😓 But for now this is the branch policy.)

Issue #1258, opened 2021-12-07
program speed


loveshipting:
Hi,
  My test condition is jlink-v9, and windows 10. I use pyocd to program, the speed is about 10KB/S.
```
$ pyocd flash -t CY8C624AFNI-S2D43 ./cypress_flash.hex  --pack E:/pyocd_pack/Cypress.PSoC6_DFP.1.2.0.pack -f 5M
0001401:INFO:load_cmd:Loading E:\WorkSpace\pyOCD\cypress_flash.hex
[---|---|---|---|---|---|---|---|---|----]
[========================================]
0219615:INFO:loader:Erased 2097152 bytes (8 sectors), programmed 2097152 bytes (4096 pages), skipped 0 bytes (0 pages) at 9.44 kB/s
```

then, I run the "speed_test.py", the speed is 12KB/S
```
$ python ./test/speed_test.py
INFO:pyocd.board.board:Target type is cy8c6xxa
INFO:pyocd.board.board:Target type is cy8c6xxa
INFO:pyocd.coresight.dap:DP IDR = 0x6ba02477 (v2 rev6)
INFO:pyocd.coresight.ap:AHB-AP#0 IDR = 0x84770001 (AHB-AP var0 rev8)
INFO:pyocd.coresight.ap:AHB-AP#1 IDR = 0x84770001 (AHB-AP var0 rev8)
INFO:pyocd.coresight.ap:AHB-AP#2 IDR = 0x24770011 (AHB-AP var1 rev2)
INFO:pyocd.coresight.rom_table:AHB-AP#0 Class 0x1 ROM table #0 @ 0xf1000000 (designer=034 part=102)
INFO:pyocd.coresight.rom_table:AHB-AP#1 Class 0x1 ROM table #0 @ 0xf0000000 (designer=034 part=102)
INFO:pyocd.coresight.rom_table:[0]<e00ff000:ROM class=1 designer=43b part=4c0>
INFO:pyocd.coresight.rom_table:  AHB-AP#1 Class 0x1 ROM table #1 @ 0xe00ff000 (designer=43b part=4c0)
INFO:pyocd.coresight.rom_table:  [0]<e000e000:SCS v6-M class=14 designer=43b part=008>
INFO:pyocd.coresight.rom_table:  [1]<e0001000:DWT v6-M class=14 designer=43b part=00a>
INFO:pyocd.coresight.rom_table:  [2]<e0002000:BPU v6-M class=14 designer=43b part=00b>
[...]

PR #1618, opened 2023-08-29
Add memory interface for jlink AP 0

marian-m12l:
Hi :wave: 

This PR aims to improve (dramatically) transfer speeds with JLink probes by implementing a memory interface using pylink's high-level read/write commands.

I believe it's implemented the way it's described in issue #1258.

Memory interface is only enabled for AP 0 and if the probe is connected to the target (i.e. if the user specified the `jlink.device` option).

Memory interface is heavily inspired by the STLink implementation.

As I'm not fluent in python and the pyproject structure, could you give me some pointers on how to run the tests, and potentially add some (for instance dedicated tests with and without target connection)?

Thanks.

### DIFF
--- a/pyocd/probe/jlink_probe.py
+++ b/pyocd/probe/jlink_probe.py
@@ -21,12 +21,14 @@ from time import sleep
 import pylink
 from pylink.enums import JLinkInterfaces
 from pylink.errors import (JLinkException, JLinkWriteException, JLinkReadException)
-from typing import (TYPE_CHECKING, Optional, Tuple)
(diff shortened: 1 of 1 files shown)


flit (changes requested):
Everything looks great! The only request is that you add your copyright to the `jlink_probe.py` file header comment.

marian-m12l:
Thanks for the quick feedback!
Added the copyright and fixed a typo that I missed when preparing for the PR.

flit:
You're fast! 😄

Issue #1553, opened 2023-05-11
Control block not found when using rtt with pyocd 0.35.0


yannroth:
After upgrading to pyocd 0.35.0, rtt is not running anymore. I am testing on custom board with an nrf52832. Here is
```
(pyocd) yann@~/venv$ pyocd --version
0.35.0
(pyocd) yann@~/venv$ pyocd rtt -t nrf52
0000936 I Target type is nrf52 [board]
0001089 I DP IDR = 0x2ba01477 (v1 rev2) [dap]
0001165 I AHB-AP#0 IDR = 0x24770011 (AHB-AP var1 rev2) [discovery]
0001165 I AP#1 IDR = 0x02880000 (AP var0 rev0) [discovery]
0001166 I NRF52832 not in secure state [target_nRF52]
0001172 I AHB-AP#0 Class 0x1 ROM table #0 @ 0xe00ff000 (designer=244 part=006) [rom_table]
0001180 I [0]<e000e000:SCS v7-M class=14 designer=43b:Arm part=00c> [rom_table]
0001185 I [1]<e0001000:DWT v7-M class=14 designer=43b:Arm part=002> [rom_table]
0001189 I [2]<e0002000:FPB v7-M class=14 designer=43b:Arm part=003> [rom_table]
0001192 I [3]<e0000000:ITM v7-M class=14 designer=43b:Arm part=001> [rom_table]
0001197 I [4]<e0040000:TPIU M4 class=9 designer=43b:Arm part=9a1 devtype=11 archid=0000 devid=ca1:0:0> [rom_table]
0001202 I [5]<e0041000:ETM M4 class=9 designer=43b:Arm part=925 devtype=13 archid=0000 devid=0:0:0> [rom_table]
0001204 I CPU core #0 is Cortex-M4 r0p1 [cortex_m]
0001205 I FPU present: FPv4-SP-D16-M [cortex_m]
0001208 I This appears to be an nRF52832 QFAA E1 [target_nRF52]
0001209 I 4 hardware watchpoints [dwt]
0001212 I 6 hardware breakpoints, 4 literal comparators [fpb]
0005943 C Control block not found [__main__]
```
When running on pyocd 0.34.3, there is no issue:
```
(pyocd) yann@~/venv$ pyocd --version
0.34.3
(pyocd) yann@~/venv$ pyocd rtt -t nrf52
0000933 I Target type is nrf52 [board]
[...]

PR #1654, opened 2023-12-18
Reset the offset in finding control block

KenMacD:
This commit fixes an issue where the control block id was not found in cases where the bytes spanned two read blocks.

Example data block causing the original issue:

`01:08:18.444,458] <dbg> SEGGER R`

In this case the code skips 24 bytes, making the next loop start with:

`SEGGER RTT\x00\x00\x00\x00…`

Unfortunately `offset` is still `8`, which the first byte does not match. The offset is then reset, but by that time the first byte has already been skipped over.

Re: #1553

### DIFF
--- a/pyocd/debug/rtt.py
+++ b/pyocd/debug/rtt.py
@@ -436,6 +436,7 @@ class GenericRTTControlBlock(RTTControlBlock):
         while search_size:
             read_size = min(search_size, 32)
             data = self.target.read_memory_block8(addr, read_size)
+            offset = 0
 
             if not data:
                 break


elfmimi:
Superseded by #1669 (proposal)

amcelroy:
+1, had the same issue and can confirm this works well.

DavidLesnjak:
Thanks for the fix and the analysis.
This is superseded by [#1669](https://github.com/pyocd/pyOCD/pull/1669), so I am closing this PR.
If you notice that your case is still not handled correctly there, please feel free to reopen this.

PR #1660, opened 2024-01-12
gdbserver: be stricter about matching qC

tlyu:
Don't match only the prefix of qC. Otherwise, compare-sections in gdb will fail, due to qCRC returning a thread ID.


ithinuel (approved):

ithinuel:
hello, would you mind rebasing your PR so that we can re-run CI please ? <3

PR #1664, opened 2024-01-17
add support for EG6832 mcu

microcai:
EG6832 is made by EGmicro.com.
this little MCU has 8k sram +64k flash.
flash is located at 0x00000000 rather than 0x8000000 :)

must first use  pyocd erase -t eg6832 --chip, then pyocd load -t eg6832 xxx.hex

now tested by the eval board selled at [EG6832 eval board](https://item.taobao.com/item.htm?abbucket=6&id=724129170854&ns=1&skuId=5208082079091)


microcai:
why these checks are not run?

ithinuel:
This is a good question, I'm trying to understand that. (thank you for leaving this message as it helps track when it started to fail).

PR #1669, opened 2024-02-04
Fix RTT Control Block Search

elfmimi:
- Revival of memory transfer in chunks. introduced in #1375
- Fix the case when control block ID is located across read block boundary.
- Addressing the possibility of the control block ID being proceeded by fraction image of itself. ex. `b'SEGGSEGGER RTT'`


jscatena88 (approved):
LGTM, fixes the issue we are having

jscatena88:
Is there anything holding up this PR? I am running into the bug as well and this seems to fix it. Anything I can do to help get it over the line?

ithinuel:
Hi, this seems to conflict with https://github.com/pyocd/pyOCD/pull/1583
The would need to be rebased and the conflict fixed please.

PR #1670, opened 2024-02-06
dependencies: remove capstone

jsiverskog:
the implementation already supports that capstone may not be installed. capstone is somewhat slow to install on some platforms (e.g. raspberry pi), and not always needed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,7 +50,6 @@ python_requires = >=3.7.0
 # importlib_resources is used instead of stdlib importlib.resources because we
 # want the selectable entry_points API, which is not present until Python 3.10.
 install_requires =
-    capstone>=4.0,<5.0
     cmsis-pack-manager>=0.5.2,<1.0
     colorama<1.0
     hidapi>=0.10.1,<1.0; platform_system != "Linux"
@@ -75,6 +74,7 @@ pemicro =
 test =
     pytest>=6.2
     pytest-cov
+    capstone>=4.0,<5.0
     coverage
     flake8
     pylint
@@ -111,4 +111,3 @@ exclude =
     test_user_script.py,
     # Ignore gdb test script for similar reasons.
     gdb_test_script.py
-


ithinuel (commented):
Hello, Thank you for the contribution.

As I understand this, this is a matter of weighing the benefit for people using this step command and relying on it printing the asm versus people installing pyocd from pip on "slow" systems.

I don't personally have an opinion on that matter and don't feel legitimate to make the call on this.

carlosperate:
Not indicating if this should be removed or kept in `install_requires`, but just to add a small comment that:
 _if_ it is going to be removed from `install_requires` it might be better if it goes into a different `extras_require`.
 Maybe a new one, like `pemicro`? and maybe at that point it'd be good as well to add a third called `all` to include all options.
On top of that it would also be good to update all the documentation indicating how to install the different dependencies.
 
Of course, I understand not wanting to do the extra work without knowing if a PR is going to be accepted, but I think it's worth pointing this out to have it in consideration as well.

ithinuel:
@carlosperate Thank you! I am not familiar enough with this aspect of python to do this myself but I’m happy to review/approve/merge such PR (provided there’s some explanations for me to understand what I’m approving).

carlosperate:
Yeah, I'm happy to help!

The main idea is to create a new group for `capstone`, we can call it something like `dissassembler`, and add that info to the [docs/installing.md](https://github.com/pyocd/pyOCD/blob/main/docs/installing.md) and [docs/installing_on_non_x86.md](https://github.com/pyocd/pyOCD/blob/main/docs/installing_on_non_x86.md) pages.
Additionally, for convinience, we can add an `all` group to include [`pemicro`, `tests` and `dissassembler`].

Ah, and we should update this error message as well:
https://github.com/pyocd/pyOCD/blob/90ee63a57a73ccca6da10b4ad5e3ebd5a6775e58/pyocd/commands/commands.py#L1732-L1734

The way it would then work is that by default `pip install pyocd` won't include these extra dependencies, and `pip install pyocd[all]` would do (or `pip install pyocd[dissasembler]` would only includes the `dissasembler`, etc).

As the `capstone` version is still locked to 4.x this is useful for platforms like Raspberry Pi and macOS arm64, as there aren't wheels for those platforms before v5.0.2, so they might end up building it from source during the installation:
- https://pypi.org/project/capstone/4.0.2/#files

https://github.com/pyocd/pyOCD/blob/90ee63a57a73ccca6da10b4ad5e3ebd5a6775e58/setup.cfg#L53

Alternatively we could increase the max `capstone` version, that might require some testing as well, but I've never looked into this specific feature, so not sure how much there might be to test. Not sure what you would prefer @ithinuel?

@jsiverskog once we know the next steps would you like to update this PR yourself? I'd like to make sure your contribution is recognised and credited, but I also understand it's been a year and maybe you've moved onto other things.

cederom:
I am FreeBSD user and Capstone always caused troubles here as system package is different from required by pyOCD, when used in virtual environment, had to built from sources, had to patch the upstream, etc etc :D 

Also on rPI building this package every time in a new environment adds to already time consuming bootstrap.

I know this is really useful feature but not always required. Having this as runtime not build time option would be of a great benefit :-)

PR #1673, opened 2024-02-17
stm32h743xx: Add AXI SRAM memory map

agausmann:
Reference: RM0433, Table 7 (Memory map)
https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf


agausmann:
I fail to see how it "no longer applies". `main` has not moved since I created this.

If you'd like, I can recreate the scenario / the commands I tried to run that do not work without this patch.

agausmann:
> On the Nucleo-H743ZI, I see that AXI ram is mapped (as `RAM_D1`).

Why would on-chip RAM be declared in the board file instead of the MCU target file?

PR #1680, opened 2024-03-24
RTT_COMMAND: Allow no down-channels

laurensmiers:
If no down-channel can be found, the target only supports printing RTT messages.

Before this commit, this target-configuration was only possible to do when supplying a log file.
Then commands were ignored and only the up-channel was written to the file.
If no log-file was supplied, we detected the missing down-channels and returned an error, never starting the viewer_loop so no up-channel messages were printed on stdout.

With this commit,
instead of returning an error when no down-channels are configured, we inform the user and start the viewer_loop
with a down-channel (parameter down_chan) with value 'None'.

Inside the viewer_loop we detect the None-value of down_chan, drop the user supplied command and continue our viewer_loop.

### DIFF
--- a/pyocd/subcommands/rtt_cmd.py
+++ b/pyocd/subcommands/rtt_cmd.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2022 Johan Carlsson <johan.carlsson@teenage.engineering>
 # Copyright (C) 2022 Samuel Dewan
 # Copyright (C) 2022 Zhengji Li
+# Copyright (C) 2024 Laurens Miers
 # SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


laurensmiers:
Not sure what is stopping the jobs from starting? If I need to do something, do tell.

szg251:
This is a must if someone uses https://docs.rs/defmt-rtt/latest/defmt_rtt/ which only creates an up channel for logging

ithinuel:
IURC approval is required from an authorised person for first time contributors.
I have to find how to do that.

ithinuel (approved):
Looks good to me. I need to find how to get CI to run.

Issue #1700, opened 2024-06-11
STM32H750 flash corruption with double buffering enabled


nattgris:
Hi,

It seems there is a problem with the flash write routines for STM32H750. A simple program like the below gives corrupted flash for about one out of five attempts for a 30 kB file, without any type of sign that something went wrong.

```python
#!/usr/bin/env python3
from pyocd.core.helpers import ConnectHelper
from pyocd.flash.file_programmer import FileProgrammer

with ConnectHelper.session_with_chosen_probe(target_override='stm32h750vbtx') as session:
    FileProgrammer(session, chip_erase="sector", smart_flash=False).program("file.elf")
```

Looking at the diff, it appears that some pages are only partially written with only the first part containing correct data, and the rest of the page being 0xff.

I was very surprised to find that neither the FileProgrammer helper, nor any of the underlying classes, support built in verification of the flash contents after programming. This is a fundamental feature of any flash programming tool. Of course you can do a manual compare separately, but that command does not have the same file format support as FileProgrammer.

I added verification support to the FlashBuilder class for easier debug of this problem. I can send a PR after some cleanup and exposing through FileProgrammer if there is any interest.

After testing a lot of times with the verify support active, it seems corruption only happens with double buffering enabled, and regardless of whether using sector erase or chip erase (which is a *lot* slower on STM32H750 for some reason). I don't see any way to switch off double buffering except changing the builder code, since the builder method to disable it doesn't seem to be exposed all the way up to FileProgrammer.

nattgris:
I would think #1696 has the same root cause, since the chips are from the same family and probably uses the same flash algo.

nattgris:
I found a builtin target stm32h743xx which works much better than the Keil.STM32H7xx pack, after being modified for the H750 flash layout. I made 100 programmings in a row without any problems, whereas the CMSIS H750 pack flash loader failed after on average 2 programmings.

By the way, I think the H743 support added in #1567 lists the wrong sector size for the flash, which should be 128K, but is listed as 8K in the flash algo dict, and 32K in the memory map. I assume this must be a mistake.

Is this the correct issue tracker for CMSIS-PACK bugs? Otherwise I think this issue can be closed. I will send a PR adding the verify support and a new builtin target for H750.

newbrain:
Just to add a data point:
I'm having the same symptoms, while trying to program the STM32H750 on a Daisyseed board.
The success rate is about one out of three tries.
[...]

Issue #1696, opened 2024-05-20
STM32H753 binary fails to run


Dillonmcewan:
I'm attempting to flash a very basic LED blinky app onto a NUCLEO-H753ZI. When I flash the board with pyOCD I see output indicating the flash was successful: `Erased 131072 bytes (1 sector), programmed 54272 bytes (53 pages), skipped 0 bytes (0 pages)`. But the app doesn't appear to start even after issuing a reset command. I started stepping through the code and it appears to hit a hard fault somewhere during RCC init. The weird thing is that if I flash the same binary using [probe-rs](https://probe.rs/) it runs without any issue, which makes me believe there's some problem in the flashing process itself rather than the binary.

nattgris:
You should be able to use the builtin target `stm32h743xx` which is the same chip but without crypto peripherals. I've found its flash loader to be working, compared to the CMSIS pack version, which corrupts the flash with high probability.

However note my https://github.com/pyocd/pyOCD/issues/1700#issuecomment-2162965594 that I think the sector sizes are specified incorrectly for that target. I'll send a PR with a fix, perhaps you can assist in testing it on a H743?

Dillonmcewan:
Thanks @nattgris, using the built-in target seems to work. I'd be happy to help with testing if you link the PR.

nattgris:
Great, #1703 is changing the sector sizes and should (still) work for your chip. Not sure if the incorrect sector size actually posed a problem, could depend on whether doing chip or sector erase.

I also noticed by just using that H743 target unmodified on my H750, that it was constantly reported by pyocd commander to be in security locked state, due to a bug in parsing the RDP value. Is that something you noticed with the builtin target? It should be fixed in my PR.

Dillonmcewan:
Got some time to test this today. I was able to flash my H753 with the `stm32h750xx` target several times without any issues.

I also checked the h743 target and did not see the same issue with getting stuck in a security locked state

PR #1702, opened 2024-06-13
Verify flash contents after programming

nattgris:
Add an option, on by default, to read out flash contents after programming and compare it to the programmed data. Without it, bugs in flash loader algos and other problems will silently leave flash corrupted.

Ref: #1700, #1696

### DIFF
--- a/pyocd/flash/builder.py
+++ b/pyocd/flash/builder.py
@@ -132,6 +132,11 @@ class _FlashSector:
         for page in self.page_list:
             page.same = False
 
+    def mark_all_pages_unknown(self):
+        """@brief Sets the same flag to False for all pages in this sector."""
+        for page in self.page_list:
+            page.same = None
+
     def __repr__(self):
         return "<_FlashSector@%x addr=%x size=%x wgt=%g pages=%s, subsectors=%d>" % (
             id(self), self.addr, self.size, self.erase_weight, self.page_list, self.n_subsectors)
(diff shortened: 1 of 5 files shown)


nattgris:
Sorry, should pull-requests be based on develop or main?

ithinuel:
`develop` is 28commits behind `main` (ie there are 28 extra commits on `main`) so I'd assume main to be the main target branch (I'm not 100% sure though).

pyocd/flash/builder.py
```diff
@@ -571,6 +572,23 @@ def program(self, chip_erase=None, progress_cb=None, smart_flash=True, fast_veri
...
+            self._scan_pages_for_same(progress_cb)
+            failed_pages = [page for page in self.page_list if page.same is False]
+            if failed_pages:
+                LOG.debug("Verify failed for pages {}".format(failed_pages))
+                raise FlashProgramFailure('flash verify failure', address=failed_pages[0].addr)
+
+        # Cleanup flash algo and reset target after programming.
+        self.flash.cleanup()
+
+        if no_reset is not True:
```
ithinuel, line 589:
```suggestion
        if no_reset is False:
```

ithinuel (changes requested):
Hi, Thank you for your contribution.

I think there should be some tests added to make sure that it does capture cases where things don't match and possible future regression.

nattgris:
> I think there should be some tests added to make sure that it does capture cases where things don't match and possible future regression.

I unfortunately have no idea how those tests could be written. I did test manually using the broken flash driver mentioned in the referenced issues.

SpaceCheetah:
This would be nice to have

PR #1703, opened 2024-06-13
Add builtin target support for STM32H750

nattgris:
Add support for STM32H750

Same as STM32H743 but with only one flash sector of 128K.

Adding a new builtin target because the `Keil.STM32H7xx_DFP` v3.1.1 CMSIS Pack flash loader is broken for at least this sub-family.

Ref: #1700

Also fix the existing targets `stm32h743xx` and `stm32h723xx` which had incorrect sector sizes specified and a bug in reading the security status.


Dillonmcewan:
Tested this with a NUCLEO-H753ZI and everything worked as expected

ithinuel (approved):

PR #1710, opened 2024-07-13
Fix test failure

jjakob:


test/unit/test_autoflush.py
```diff
@@ -29,11 +29,11 @@ class TestAutoflush:
     def test_flushed(self, mock_obj):
         with Autoflush(mock_obj):
             pass
-        assert mock_obj.flush.called
+        assert mock_obj.flush.assert_called
```
frogamic, line 32:
```suggestion
        mock_obj.flush.assert_called()
```

If I understand the `mock` documentation correctly, these asserts are functions that should be called in tests

test/unit/test_autoflush.py
```diff
@@ -29,11 +29,11 @@ class TestAutoflush:
...
 
     def test_transfer_err_not_flushed(self, mock_obj):
         with pytest.raises(TransferError):
             with Autoflush(mock_obj):
                 raise TransferError("bad joojoo")
-        assert mock_obj.flush.not_called
+        assert mock_obj.flush.assert_not_called
```
frogamic, line 38:
```suggestion
        mock_obj.flush.assert_not_called()
```

frogamic (changes requested):
I came here to raise the same PR since pyocd is a dependency of a package I maintain in NixOS and this broken test is causing issues as we migrate to python 3.12.

test/unit/test_autoflush.py
jjakob: I think you're right

jjakob:
I'm not sure how this test is supposed to work but at least after @frogamic suggestions it's now syntactically correct, and it passes, so I'll leave it up to pyOCD developers to see if it actually does what they want it to do.

PR #1716, opened 2024-08-06
pack: flash: Support flash_algos with no RW section

konkers:
The latest version of https://github.com/rp-rs/flash-algo does not produce an elf file with an RW section.


RobertRostohar:
Apologizes for not replying earlier.

In the meantime, this problem has already been addressed. Commit 5a4105f among other things also enables flash algorithms without RW sections.

PR #1717, opened 2024-08-07
coresight: discovery: Avoid double offset in root component creation

konkers:
Because the APAccessMemoryInterface includes and offset and the component factory API includes an address base, the base address offset at cpmid.address was being applied twice.

### DIFF
--- a/pyocd/coresight/discovery.py
+++ b/pyocd/coresight/discovery.py
@@ -262,7 +262,7 @@ class ADIv6Discovery(CoreSightDiscovery):
             memif = APAccessMemoryInterface(self.dp, ap_address)
 
             # Instantiate the component and attach to the target.
-            component = cmpid.factory(memif, cmpid, cmpid.address)
+            component = cmpid.factory(memif, cmpid, 0)
             self.target.add_child(component)
             component.init()
         except exceptions.Error as e:


pyocd/coresight/discovery.py
```diff
@@ -262,7 +262,7 @@ def _create_root_component(self, cmpid):
             memif = APAccessMemoryInterface(self.dp, ap_address)
 
             # Instantiate the component and attach to the target.
-            component = cmpid.factory(memif, cmpid, cmpid.address)
+            component = cmpid.factory(memif, cmpid, 0)
```
ithinuel, line 265: I tracked down this parametter to be set to [`CoreSightComponent._address`](https://github.com/pyocd/pyOCD/blob/456cbf876228382af10247cf5d210b6ffea6ff67/pyocd/coresight/component.py#L36).

So if I understand correctly, this should be set to None instead.

I can see that the AP Address is used as an offset in [`APAccessMemoryInterface`](https://github.com/pyocd/pyOCD/blob/456cbf876228382af10247cf5d210b6ffea6ff67/pyocd/coresight/dap.py#L1044-L1057). Could that be the issue ?
Are all read/write operations to this root AP meant to be relative to the base address?

EG, hypothetically, if there’s an SCS right on the root AP, then there could be some read/write in the 0xExxx_xxxx space even if the root’s base address is in 0xFxxx_xxxx, couldn’t it?

Could be please point to where it is applied a second time?

konkers: > I tracked down this parametter to be set to [`CoreSightComponent._address`](https://github.com/pyocd/pyOCD/blob/456cbf876228382af10247cf5d210b6ffea6ff67/pyocd/coresight/component.py#L36).
> 
> So if I understand correctly, this should be set to None instead.

`None` makes sense instead of `0` here.

> I can see that the AP Address is used as an offset in [`APAccessMemoryInterface`](https://github.com/pyocd/pyOCD/blob/456cbf876228382af10247cf5d210b6ffea6ff67/pyocd/coresight/dap.py#L1044-L1057). Could that be the issue ? Are all read/write operations to this root AP meant to be relative to the base address?
> 
> EG, hypothetically, if there’s an SCS right on the root AP, then there could be some read/write in the 0xExxx_xxxx space even if the root’s base address is in 0xFxxx_xxxx, couldn’t it?
> 
> Could be please point to where it is applied a second time?

I was debugging an issue where a funnel was faulting on access: https://github.com/pyocd/pyOCD/blob/main/pyocd/coresight/funnel.py#L47

Since it's using the instantiated `APAccessMemoryInterface` which is applying `cmpid.address` as an offset to all accesses the `self.address + TraceFunnel.DEVID` results in a double offset.

I don't have enough of a high level view to understand which application of this makes sense.  Perhaps instantiating `APAccessMemoryIterface` w/o an offset (or an offset derived from somewhere else ) makes sense?

konkers: Since it's today instead of yesterday I can say that the chip I was debugging was the rp2350 (see #1718).   The error ends up being ignored and you can flash, erase, reset the rp2350 fine.

RobertRostohar:
Sorry for picking this up so late.

Similar issue was seen also Alif E7 (offset address applied twice).

This was fixed with Commit 3a78de3, making this PR obsolete.

PR #1732, opened 2024-10-29
HDSC MCU updated targets and added new PNs.

lennvn:
- Updated targets HC32F448/460/4A0(2).
 - New targets HC32F115/155/334/467/472.
 - New PNs: hc32a448, hc32f4a2, hc32f115, hc32f155, hc32f334, hc32f467, hc32f472.


lennvn:
Should I exclude Python3.7 from MAC?

PR #1735, opened 2024-10-31
Add support for ARM China Star-MC2.

Quincy-Zh:
- Add support for ARM China Star-MC2.


RobertRostohar:
Star-MC2 is the same as Cortex-M52 which is now supported with #1852.

PR #1741, opened 2024-11-11
target: add support for nRF54L15

maxd-nordic:
This patch adds support for nRF54L15 to pyOCD.


pyocd/coresight/ap.py
```diff
@@ -798,7 +798,10 @@ def _init_rom_table_base() -> None:
             is_adiv5_base = (base & AP_BASE_FORMAT_MASK) != 0
             is_base_present = (base & AP_BASE_ENTRY_PRESENT_MASK) != 0
             is_legacy_base_present = not is_adiv5_base and not is_base_present
-            if is_legacy_base_present:
+            if base == 0xe00fe002:
```
maxd-nordic, line 801: Any suggestions welcome on how to do the rom table lookup properly. Seems like we don't set that bit, but e.g. Segger doesn't seem to care about it, even without specific chip support.

maxd-nordic:
I have tested with the a Rev B chip and all the relevant parts work nicely: unlocking, UICR writing, mass-erase. Note that since this is not actually a flash based device, programming speeds are going to be a bit worse than they could be. I still think this is very usable.

pyocd/coresight/ap.py
maxd-nordic: @ithinuel found and pushed a solution! :)

ithinuel (approved):

PR #1748, opened 2024-12-17
fix typing of chip_erase from bool -> str

NilsIrl:


ithinuel (approved):
I added an extra change on `loader.py:141`.
That’s a good catch. Thank you 👍

PR #1752, opened 2025-01-13
emit an error log when setting a breakpoint fails in gdbserver

koalatux:
I found having an error logged in pyOCD useful, when the IDE silently ignored the error reported by gdbserver.


ithinuel:
Hi, thank you for your contribution. I think it would be better to add that in the breakpoint manager instead (where other `set_breakpoint` are reported).

As far as I can see this would be in:
- `pyocd/coresight/fpb.py`
- `pyocd/debug/breakpoints/manager.py`
- `pyocd/debug/breakpoints/software.py`

PR #1753, opened 2025-01-13
Avoid potential exception

selescop:
when 'product_name' is not specified
(a call to to_str_safe() will trigger an exception)


RobertRostohar:
Sorry for not looking into this earlier.

When would `deviceInfo['product_string']` be None?

Looking at `hid.enumerate` from [cython-hidapi 0.14](https://github.com/trezor/cython-hidapi/blob/0.14.0/hid.pyx):
When there is no `product_string`, the NULL C pointer is converted to `""` and not `None`.

PR #1755, opened 2025-01-15
Fix For CMSIS Dap Bulk SWO

jonathanherbstgrapple:
This is how I was able to fix the issue I created https://github.com/pyocd/pyOCD/issues/1754.  Not sure if there are larger issues at play here.

### DIFF
--- a/pyocd/probe/pydapaccess/interface/pyusb_v2_backend.py
+++ b/pyocd/probe/pydapaccess/interface/pyusb_v2_backend.py
@@ -150,6 +150,7 @@ class PyUSBv2(Interface):
 
     def start_swo(self):
         thread_name = "SWO receive (%s)" % self.serial_number
+        self._swo_thread_did_exit = False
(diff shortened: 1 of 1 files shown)


ithinuel:
Hi, thank you for the contribution.

I see that the `rx_task` also sets `_swo_thread_did_exit` to true. I’m not sure if that’s the intended behaviour of if those two should be using a different `*_thread_did_exit`.

crypto-lars:
> Hi, thank you for the contribution.
> 
> I see that the `rx_task` also sets `_swo_thread_did_exit` to true. I’m not sure if that’s the intended behaviour of if those two should be using a different `*_thread_did_exit`.

I think this was not intended when this code was written and my pull request fixes that.

TeoMahnic:
Hi @jonathanherbstgrapple, thank you for your contribution.
I am closing this PR as it has been superseded by #1824. If the issue persists, please open a new issue or PR.

PR #1758, opened 2025-02-03
Add missing return to UserScriptFunctionProxy

gotlaufs:
Commit 4ece4ed introduced optional user script function arguments, but it looks like the `return` went missing.

Without it it is not possible to override `start_debug_core`, because return value is always `None`.


ithinuel (approved):
LGTM

PR #1777, opened 2025-05-09
Filter out Cypress KitProg3 bridge in CMSIS-DAP probe

TeoMahnic:


pyocd/probe/pydapaccess/interface/common.py
```diff
@@ -97,6 +97,8 @@
 CMSIS_DAP_IDS_TO_FILTER_BY_USAGE_PAGE: List[VidPidPair] = [
```
mathias-arm, line 97: I suggest also changing the comment above `Currently these are only NXP probes.` to `Currently these are only NXP and Cypress probes.`

PR #1791, opened 2025-07-23
Drop downstream data in rtt_server if no down channels

amari-ca:
Similar to !1680, this change allows rtt_server to function if there are no down channels; previously the downstream data would be appended to a non-existent entry in self.down_buffers and pyOCD would crash.


DavidLesnjak:
@amari-ca , thanks for the contribution.

This is already addressed by PR [#1941](https://github.com/pyocd/pyOCD/pull/1941), so I’m closing this PR as superseded.

The fix has also been released, so please try again with a current released version of pyOCD. If you still see the same problem, feel free to reopen this PR.

PR #1814, opened 2025-09-18
Add: realtek 8773G builtin

realmcu:
Added support for RTL8773G


TeoMahnic:
Hi @realmcu, thank you for providing support for this target.

Our preference is to rely on CMSIS packs for adding new device support, but since there doesn't seem to be a pack available for the Realtek 8773G device, we can proceed with adding internal support instead.

Could you please share details on how this device support was implemented and tested? It would also be helpful to include information on the hardware used for testing.

PR #1823, opened 2025-10-09
rtos: zephyr: use thread states from modern Zephyr versions

mathieuchopstm:
The existing thread states were coming from a very old Zephyr version.

Align to more modern definitions (found here: https://github.com/zephyrproject-rtos/zephyr/blob/8d202cb9485558b24310334ef000c171de94ff07/include/zephyr/kernel_structs.h#L51-L73)

Also handle the special case of idle threads which don't have a proper state when not running (for now, this is based on thread name matching; the more solid approach would require a deeper rework + some help from Zephyr)

### DIFF
--- a/pyocd/rtos/zephyr.py
+++ b/pyocd/rtos/zephyr.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2016-2020 Arm Limited
 # Copyright (c) 2022 Intel Corporation
 # Copyright (c) 2022 Chris Reed
+# Copyright (c) 2025 STMicroelectronics
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


MiloradCvjetkovic:
Hi @mathieuchopstm, 

thank you for the contributions.

Can you please explain what is the reason and added value of changing current `idle` thread behavior?

It looks that current display of `idle` thread can alternate between `Ready` and `Running` state which seems to be correct from behavior point of view.
Also, `idle` thread has the lowest priority defined as `K_IDLE_PRIO` which is displayed as such, yes, it does not compete with other threads of the same priority but it does not seem wrong.

On the screenshot below you can see comparison of old `idle` display vs your suggested :

Idle_old_new

pyocd/rtos/zephyr.py
```diff
@@ -172,7 +179,7 @@ def __init__(self, targetContext, provider, base, offsets):
         self._base = base
         self._thread_context = ZephyrThreadContext(self._target_context, self)
         self._offsets = offsets
-        self._state = ZephyrThread.READY
+        self._state = ZephyrThread.QUEUED
```
RobertRostohar:
```suggestion
        self._state = 0
```

pyocd/rtos/zephyr.py
```diff
@@ -224,6 +231,25 @@ def name(self):
...
+        # directly from 'base.thread_state'; as such, it will be
+        # valid even for idle threads, unlike all other values.
+        #
+        # HACK: assume threads called "idle" are idle threads.
+        # A proper method would check that this is one of the threads
+        # in the "z_idle_threads" array (or that is matches one of
+        # _kernel.cpus[].idle_thread, which is the same thing).
+        if self._name == "idle":
+            if self.state == ZephyrThread.RUNNING:
+                return "Running; <idle thread>"
```
RobertRostohar:
```suggestion
                return "Running"
```

pyocd/rtos/zephyr.py
```diff
@@ -224,6 +231,25 @@ def name(self):
 
     @property
     def description(self):
+        # Idle threads must be handled separately: when not running,
+        # they are neither SLEEPING, PENDING nor QUEUED and simply
+        # exist in a "limbo state" outside scheduler consideration.
```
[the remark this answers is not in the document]
MiloradCvjetkovic: Thank you @mathieuchopstm .

PR #1829, opened 2025-10-17
refactor: Migrate to PEP621, add uv.lock

ithinuel:
This makes packaging easier on newer systems and facilitates
reproducibility of the builds.

This also migrates to capstone 5 as capstone 4 does not provide wheels
for aarch64 targets.

Note: I do not have the expertise around setup.py & python packaging in
general. Part of this commit has been co-developed with an LLM.
I checked it works on my system.


flit:
Without checking the `pyproject.toml` entries in detail, this looks good to me and is overdue. Thanks!

ithinuel:
#1793 introduced the switch to capstone 5 in the meantime, so i'm removing the mention from this PR :)

PR #1831, opened 2025-10-20
load: improve programming of multiple files

TeoMahnic:
- fix bug where smart flash was not used for pack targets
- change loading reset mechanism: selected reset and halt before programming and reset using nSRST at the end
- flash algorithm extra symbols are added only if they exist in the binary
- builder: do init and uninit only once for all pages in a sector


TeoMahnic:
Rebased on top of develop branch.

PR #1843, opened 2025-10-24
Add reset type option to the load command

afeinman-snap:
pyOCD v.040.0 introduces a forced reset at the end of load command. Currently, this reset is fixed to N_SRST type. This fails on hardware that uses debug interface three-wire connection (SWDIO/SWDCLK/GND). 
This PR addressed the issue https://github.com/pyocd/pyOCD/issues/1842 by adding a reset type option (same syntax as for the reset command) and defaulting to SW reset (same as reset command)


TeoMahnic:
Closing, as it was superseded by #1861

PR #1853, opened 2025-11-19
target: add support for nRF54LM20A

StarSphere-1024:
This patch adds support for nRF54LM20A to pyOCD.


StarSphere-1024:
Can someone tell me what I need to do to get these workflows started?

PR #1855, opened 2025-11-25
target: only load SVD file if it exists

TeoMahnic:


TeoMahnic:
Rebased onto develop and squashed commits.

RobertRostohar (approved):

PR #1873, opened 2025-12-15
coresight: discovery: don't add cores not described in debug topology

TeoMahnic:
- When using cbuild-run/CMSIS-Pack targets, skip adding cores that are not present in the debug topology. These cores remain accessible through direct AP read/write operations, but are no longer inserted into the discovered core list.


TeoMahnic:
This PR applies the changes proposed in https://github.com/pyocd/pyOCD/issues/1800#issuecomment-3646071882.

PR #1889, opened 2025-12-23
target: add support for nRF54LM20A

StarSphere-1024:
This patch adds support for nRF54LM20A to pyOCD.

### DIFF
--- a/docs/builtin-targets.md
+++ b/docs/builtin-targets.md
@@ -795,6 +795,11 @@ title: Built-in targets
     <td>Nordic Semiconductor</td>
     <td>NRF54L15</td>
     </tr>
+    
+    <tr><td><code>nrf54lm20a</code></td>
+    <td>Nordic Semiconductor</td>
+    <td>NRF54LM20A</td>
+    </tr>
 
     <tr><td><code>nrf91</code></td>
     <td>Nordic Semiconductor</td>
--- a/pyocd/target/builtin/__init__.py
+++ b/pyocd/target/builtin/__init__.py
@@ -61,6 +61,7 @@ from . import target_nRF51822_xxAA
 from . import target_nRF52832_xxAA
 from . import target_nRF52833_xxAA
 from . import target_nRF52840_xxAA
+from . import target_nRF54LM20A
 from . import target_nRF54L15
 from . import target_nRF91xx
 from . import target_S32K344
@@ -216,6 +217,7 @@ BUILTIN_TARGETS = {
           'nrf52840' : target_nRF52840_xxAA.NRF52840,
           'nrf91' : target_nRF91xx.NRF91XX,
           'nrf54l' : target_nRF54L15.NRF54L15,
+          'nrf54lm20a' : target_nRF54LM20A.NRF54LM20A,
           's32k344': target_S32K344.S32K344,
           'stm32f103rc': target_STM32F103RC.STM32F103RC,
           'stm32f051': target_STM32F051T8.STM32F051,
--- a/pyocd/target/builtin/target_nRF54LM20A.py
+++ b/pyocd/target/builtin/target_nRF54LM20A.py
@@ -1,0 +1,107 @@
+# Copyright (c) 2025 StarSphere. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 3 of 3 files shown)


StarSphere-1024:
Can someone help me deal with PR as soon as possible?

StarSphere-1024:
Why is there no response at all? I have already verified that this works on the nRF54LM20DK development board. Please, we really need this.

TeoMahnic:
Hi @StarSphere-1024,

Apologies for the delay in reviewing this PR, and thank you for your contribution.

Our current preference is to use CMSIS Packs for target support whenever possible, and we aim to keep new additions aligned with that approach. However, as `nRF54LM20A` does not appear to have a CMSIS Pack available at this time, we can make an exception in this case.

Regarding the nRF54LM20 DK: I do not have access to this hardware for local validation, so I will rely on your confirmation that the implementation has been tested and works as expected on that device.

StarSphere-1024:
Hi @TeoMahnic,

Thank you for the review.

I can confirm that I have successfully tested this implementation on the nRF54LM20 DK. Using the onboard J-Link, I was able to flash firmware via pyOCD.

I've addressed the requested changes. Should I rebase to squash these commits?

StarSphere-1024:
Hi @TeoMahnic 
I removed the `'begin_data' : 0x20000000 + 0x1000` line as you requested, but now flashing fails with:
```
0000261 W NRF54LM20A is not in a secure state [target_nRF54LM20A]
0000363 C Error: 'begin_data' [__main__]
```
I checked the codebase and found `begin_data` is still used in many places:
- `pyocd/flash/flash.py:146` - `self.begin_data = flash_algo['begin_data']`
- `pyocd/flash/flash.py:332-341` - `compute_crcs()`
- `pyocd/flash/flash.py:395-399` - `program_page()`
- `pyocd/flash/flash.py:457-461` - `program_phrase()`

Are you sure this attribute has been deprecated?
Or do I need to update `page_buffers` instead?

TeoMahnic:
Hi @StarSphere-1024,

The [`begin_data` attribute was removed](https://github.com/pyocd/pyOCD/commit/077b1e0c5f1a616c51cd8209231734452e64bd04) in pyOCD [v0.43](https://github.com/pyocd/pyOCD/releases/tag/v0.43.0). Please rebase onto latest `develop` branch, and try flashing again.

There are currently some unrelated issues with functional tests in CI, but once you confirm that flashing works correctly we can proceed with merging this PR.

StarSphere-1024:
Hi @TeoMahnic,

Thanks for the clarification on begin_data. I've rebased onto the latest develop branch, removed the deprecated attribute, and confirmed that flashing works on the nRF54LM20 DK when the chip already has firmware.

However, while testing with a J-Link probe, I hit a critical crash (Segmentation Fault) in pyocd/probe/jlink_probe.py that happens before the flash algorithm even runs. This looks like a pre-existing bug in the J-Link integration that affects other targets too, not just my new one.

## The Issue:
When using a J-Link, pyOCD crashes immediately on connection:
```
0000277 C No emulator with serial number 1051807121 found. [main]
Segmentation fault (core dumped)
```

(Note: The same command works fine with CMSIS-DAP.)

## Root Cause:
I traced this to two bugs in jlink_probe.py:
Wrong call order in open(): disable_dialog_boxes() is calledbefore link.open(). On some J-Link DLL versions, this causes the subsequent open() to fail, leaving the DLL in a bad state.
Resource conflict in get_all_connected_probes(): This method creates a temporary JLink instance to list devices but never closes it. Then JLinkProbe.init creates a second instance. Having two active instances causes a race condition; the second open fails, and the cleanup triggers a segfault.

## The Fix:
I verified these changes fix the crash and allow stable flashing on both empty and programmed chips:

pyocd/probe/jlink_probe.py

Fix 1: Close the temporary JLink instance
```python
@classmethod
def get_all_connected_probes(cls, unique_id=None, is_explicit=False):
    try:
        jlink = cls._get_jlink()
        if jlink is None:
            return []
        
        serial_numbers = [cls._format_serial_number(info.SerialNumber) for info in jlink.connected_emulators()]
        
        # FIX: Close the temporary instance before creating JLinkProbe objects
        jlink.close() 
        
        return [cls(sn) for sn in serial_numbers]
    except JLinkException as exc:
        raise cls._convert_exception(exc) from exc
```
Fix 2: Correct order in open()
```python
def open(self):
    assert self.session
    try:
        # FIX: Open the device FIRST
        self._link.open(self._serial_number_int)
        self._is_open = True

        # THEN disable dialog boxes
        if self.session.options.get('jlink.non_interactive'):
            self._link.disable_dialog_boxes()
            
        # ... rest of existing code
```

Since this fixes a major crash for all J-Link users.
Should I include these fixes in the current PR, or split them out into a separate PR for the J-Link probe??


Best,
@StarSphere-1024

TeoMahnic:
Hi @StarSphere-1024,

Thank you for confirming that flashing works with the added internal device support.

>When using a J-Link, pyOCD crashes immediately on connection
...
Should I include these fixes in the current PR, or split them out into a separate PR for the J-Link probe??

Please keep the current PR focused on the internal device support. The J-Link fix should be addressed in a separate PR, as also mentioned in https://github.com/pyocd/pyOCD/pull/1927#issuecomment-4053391621.

TeoMahnic (approved):

PR #1890, opened 2026-01-05
pyocd: core: helpers.py: reset colour for invalid choice

alexapostolu:
The yellow foreground of `Invalid choice` leaks into the `Probe/Board` selection

Screenshot 2026-01-05 at 4 05 18 PM


RobertRostohar:
Thanks for spotting this and sorry for the long response time.

PR #1903, opened 2026-02-10
ci: add nightly workflow

TeoMahnic:


TeoMahnic:
Correct. The nightly workflow does not include the .whl artifact.
The artifact names already include a short version of the commit hash (similar to the release build workflow).
Example: `pyocd-windows-0.42.1.dev10+gc025cf693`

PR #1905, opened 2026-02-11
Adapt to fnmatch.translate() using \z instead of \Z in Python 3.14

elfmimi:
fnmatch.translate() returns slightly different result in Python 3.14 .

https://docs.python.org/3.13/library/fnmatch.html
> '(?s:.*\\.txt)\\Z'

https://docs.python.org/3.14/library/fnmatch.html
> '(?s:.*\\.txt)\\z'

this causes `pyocd pack find X` to only match device names ending with X .


RobertRostohar:
@elfmimi, thanks for looking into this and providing a solution.

I would slightly change the logic. Therefore, I suggest we continue this in #1993, which also registers the Pyhton 3.14 to classifiers and CI.

PR #1916, opened 2026-02-23
target: family/rp2: missing 256KB memory region

issa-tseng:
The RP2350 has two individually-striped 256KB regions followed by the 2 independent 4KB scratch regions the RP2040 also had. (the RP2040 only has one 256KB region before scratch). Update the memory map so GDB knows it can access these addresses. If this script is also used by 2040 targets then something has to be done to split the logic here.


issa-tseng:
ah, i was fooled by the RP2 designation into thinking this was pico 2 only. i realized this and read the rest of the file and i see that the logic does have to be split. should i just copy the MEMORY_MAP out of base and add it to the other classes? i'm not very familiar with python. thanks!

PR #1927, opened 2026-03-13
probe: jlink: open emulator before issuing commands

dottspina:
## Problem description

Using a JLink probe (USB) with JLink DLL version 9.14a and above (latest immune version is 9.12), pyOCD will fail to connect to the emulator, raising  "No emulator with serial number XXXXXXXXXX" or "Could not connect to default emulator", depending on the command line arguments. 

I myself encountered this problem (programming/debugging an nRF52840-DK), as did at least one other pyOCD user [1], and one pylink user [2].
On my machine (Linux), pyOCD will additionally segfault in `JLINKARM_Close()` , probably on library unload (seen in coredump). 

## Proposed solution

It seems that the required emulator status (opened, connected) for executing some commands in  `JLink.disable_dialog_boxes()` have changed in the JLink SDK v9.14, triggering automatic connections the pylink interface does not take into account (yet ?).
They've started looking into it, but they're working (like most of us probably) without the JLink SDK documentation, and reverse engineering the state machine might not be straightforward: It is not even certain that `JLINKARM Open()` should not have to be called, in one way or another, before dispatching commands.

Proposed solution: in pyOCD `JLinkProbe.open()`, call `JLink.disable_dialog_boxes()` after `JLink.open()`.

This fix Works for me (TM), and `test/automated_test.py -d` succeeds, expect "Gdb Test" which is ignored (I don't have the right Python runtime for the versions of arm-none-eabi-gdb that I have at hand). 
Bellow is my `pyocd.yaml`:
``` yaml
probes:
  XXXXXXXXX # My probe's serial number.
    target_override:  nrf52
    test_binary:      l1_nrf52840-dk.bin
    frequency:        6000000
```

Hope this helps.

[1] pyocd/pyOCD#1925 Jlink V9.22 No emulator with serial number
[2] square/pylink#254 Can't connect to the emulator

### DIFF
--- a/pyocd/probe/jlink_probe.py
+++ b/pyocd/probe/jlink_probe.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2020,2025 Arm Limited
 # Copyright (c) 2021-2022 Chris Reed
 # Copyright (c) 2023 Marian Muller Rebeyrol
+# Copyright (c) 2026 Christophe Dufaza
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -187,11 +188,22 @@ class JLinkProbe(DebugProbe):
         try:
             # Configure UI usage. We must do this here rather than in the ctor because the ctor
             # doesn't have access to the session.
-            if self.session.options.get('jlink.non_interactive'):
+            non_interactive = self.session.options.get('jlink.non_interactive')
+
+            # With JLink DLL versions 9.14 and above, calling disable_dialog_boxes()
+            # before open() puts the pylink adapter into an inconsistent state.
+            # See:
+            # - #1925 "Jlink V9.22 No emulator with serial number"
+            # - #1927 "probe: jlink: open emulator before issuing commands"
+            # - square/pylink#259 "Fix exec command"
+            if non_interactive and self._link.version < "9.14":
                 self._link.disable_dialog_boxes()
 
             self._link.open(self._serial_number_int)
             self._is_open = True
+
+            if non_interactive and self._link.version >= "9.14":
+                self._link.disable_dialog_boxes()
 
             # Get available wire protocols.
             ifaces = self._link.supported_tifs()


StarSphere-1024:
Yes, almost simultaneously, I also discovered the issue and the solution:
https://github.com/pyocd/pyOCD/pull/1889#issuecomment-4051795717

dottspina:
> Yes, almost simultaneously, I also discovered the issue and the solution: [#1889 (comment)](https://github.com/pyocd/pyOCD/pull/1889#issuecomment-4051795717)

@StarSphere-1024, yes same issue and same root cause identified.

AFAICT #1889 does not include the fix, rightly since its purpose is to "add support for nRF54LM20A". I think we can merge both pull requests, one per topic.

Would that be good for you?

StarSphere-1024:
Sure.my PR (https://github.com/pyocd/pyOCD/pull/1889)  is strictly for nRF54LM20A support. Since I use DAPLink, this J-Link issue doesn't affect me personally. However, it definitely needs a quick fix. Hope it gets merged soon—the review process has been so slow (my PR waited ~4 months) that I had to switch to OpenOCD for my project.

dottspina:
Follow-up.

**Problem description (update)**

Currently, when using JLink DLL above version 9.12:
- semi-hosting unit tests (`test/unit/test_semihosting.py`) fail 
- automated tests (`automated_test.py`) fail (nrf52840 target)
- `pyocd` fails to open J-Link probes connected to USB (either default or with serial number), and will possibly segfault

It is confirmed that the issue comes from automatic connections created by `JLink.disable_dialog_boxes()` when called before `JLink.open()`, which is what pyOCD does in `JLinkProbe.open()`.
The error only occurs when using a version of JLINK DLL above 9.12 (v912). Observed with 9.14 to 9.28.

This is being investigated on the pylink side (square/pylink#259), and the real fix will eventually come from there.
AFAICT, the most promising approach seems to be to patch `disable_dialog_boxes()` to skip/defer the J-Link commands whose behavior has changed.
But this is still debated (there is no public documentation to rely on).

**Proposed change (update)**

The initial version of this PR, i.e. the reordering of calls to `JLink.disable_dialog_boxes()` and `JLink.open()`, fixed all the above problems (tests pass, including `automated_test.py -i "Gdb Test"`, and emulators are found), but could degrade the user experience: on Windows, some dialog boxes will appear on `JLinkProbe.open()`, even though they should have been disabled.

This second version instead adds a workaround that should:
- preserve the current behavior as much as possible, i.e. call `disable_dialog_boxes()` before `open()` when it is considered safe
- while allowing the use, but also the development and testing, of pyOCD with current versions of the JLink DLL

Once the root cause is fixed in pylink and the fix is available (PyPI), we should be able to:
- unconditionally call `disable_dialog_boxes()` before `open()`
- update pyOCD requirements with the appropriate versions of the pylink adapter (note that it is unlikely that the fix will be backported to the 1.7 branch of ~~pyocd~~ pylink)

Thanks.

michaelgross-segger:
Hello everyone, I would like to add following info regarding the issue:
In the current implementation, `disable_dialog_boxes()` executes multiple J-Link DLL commands, some of which are not allowed to be called before a call to JLINK_Open(). Because of this, these function calls return with an error. The commands in question in this case are `HideDeviceSelection`, `DisableInfoWinFlashDL`, and `DisableInfoWinFlashBPs`. Removing these function calls from the original `disable_dialog_boxes()` and moving them below `self._link.open(self._serial_number_int)` & `self.is_open=True` solves this issue. 

Please refer to our documentation for a list of all possible J-Link commands, where commands which are allowed to be called before JLINK_Open() are denoted as such:
https://kb.segger.com/J-Link_Command_Strings

Should there be other issues or potential bugs with J-Link commands or J-Link in general, please don't hesitate to reach out to us via our support system: https://support.segger.com

RobertRostohar:
@dottspina, thanks for the workarround.

PR #1930, opened 2026-03-23
probe: jlink: Fix possibly misleading is_open status

dottspina:
It was probably a typo: in `JLinkProbe`, the `is_open` property answers the `JLink.opened` method itself instead of its return value.

As a result, `is_open` always evaluates to True in boolean contexts, and may report misleading status.

Thanks.


RobertRostohar:
Thanks for reporting the typo.

PR #1953, opened 2026-04-07
coresight: cortex_m: gracefully handle inaccessible core after reset

fbnors:
When performing a post-reset check in `_inner_reset` to ensure the core is not halted, `get_state()` reads the DHCSR register via SWD. On certain MCUs (like the LPC55S28), the SWD link may drop or become temporarily inaccessible immediately after a reset, causing `get_state()` to throw a TransferFaultError.

Previously, this unhandled exception would bubble up and cause the flash operation to crash with a fatal status 1 exit code, even if the flash and reset were otherwise successful.

This adds a `try/except` block around the state check in `_inner_reset()` to catch `exceptions.Error`. If the core state cannot be read or resumed, it logs a warning instead of aborting the entire command.


TeoMahnic:
Hi @fbnors, thank you for providing this fix.

As of v0.44.1, the implicit resume after reset behavior has been removed, so this change is no longer applicable. Therefore I am closing this PR.

Issue #1767, opened 2025-04-26
How to set the flash algorithm size in memory


ccu6:
I now have a flm file for programming the off-chip spi flash, and exported the flash algorithm through the script.
But when I flash the device, a memory transfer fault occurred. I checked the debug information and found that the flash algorithm ram space set may be too small, because the flm file provider suggested setting the ram space in keil to 0xF000, and the debug information seemed to only provide 0x8000 space. How can I solve this problem? This is my user script file and debug info. Thank you for your answer.

debug info:
> 0002824 D Flash algo info: Flash Device:
  name=b'STM32H750xx'
  version=0x101
  type=1
  start=0x8000000
  size=0x20000
  page_size=0x400
  value_empty=0xff
  prog_timeout_ms=100
  erase_timeout_ms=6000
  sectors:
    start=0x0, size=0x20000
 [flm_region_builder]
0002824 D flash algo: [stack=0x20007410; 0x7410 b] [b2=0x20007810,+0x7810] [b1=0x20007410,+0x7410] [code=0x20007c1c,+0x7c1c,0x3e4 b] (ram=0x20000000, 0x8000 b) [flash_algo]
0002825 D Running task notify [sequencer]
0002826 I Loading C:\Users\CCU6\Desktop\Hardware\STM32H750\H750_UART\build\Debug\H750_UART.hex [load_cmd]
0002870 D halting core 0 [cortex_m]
0002873 D set reset catch, core 0 [cortex_m]
0002873 D halting core 0 [cortex_m]
0002878 D reset, core 0, type=SW_SYSTEM [cortex_m]
0003002 D clear reset catch, core 0 [cortex_m]
0003581 D resuming core 0 [cortex_m]
0003582 D added=[] removed=[] [manager]
0003582 D bps after flush={} [manager]
0003595 D uninit session <pyocd.core.session.Session object at 0x0000025378948550> [session]
0003595 D uninit board <pyocd.board.board.Board object at 0x0000025378962270> [board]
0003598 D resuming core 0 [cortex_m]
0003598 D added=[] removed=[] [manager]
0003599 D bps after flush={} [manager]
0003609 D closing interface [hidapi_backend]
0003609 C Memory transfer fault @ 0x90000000-0x9000001f [__main__]

user script in file

[fk750_spiflash.py.txt](https://github.com/user-attachments/files/19922223/fk750_spiflash.py.txt)

PR #1976, opened 2026-06-08
pack: clamp flash-algo RAM to the device's RAM region

eakorolev:
`_select_flash_ram()` uses the pack's `<algorithm RAMsize>` as-is. When a DFP declares it larger
than the device's RAM, the loader (laid out top-down) lands past physical RAM and flashing faults
before erase.

Caught on `SiliconLabs.GeckoPlatform_EFR32MG21_DFP` / `EFR32MG21A020F768` (RAM `0x10000`, algo
`RAMsize=0x18000`): `Memory transfer fault (write) @ 0x20017cb0`.

Fix: clamp the algo RAM to the containing RAM region — same as the flash-range clamp in
`_add_flash_subregions()`. Added `test_ram_select_explicit_clamped_to_region` (fails without the fix).

---
Related: #1767 (same class of problem, unfixed); #1468 works around it per-target for MAX32660.

### DIFF
--- a/pyocd/target/pack/flm_region_builder.py
+++ b/pyocd/target/pack/flm_region_builder.py
@@ -1,6 +1,7 @@
 # pyOCD debugger
 # Copyright (c) 2022-2023 Chris Reed
 # Copyright (c) 2025 Arm Limited
+# Copyright (c) 2026 Evgeny Korolev
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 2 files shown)


TeoMahnic:
Hi @eakorolev, thanks for the PR and for looking into this. With #2006 merged, this should now be addressed, so I will close this PR as superseded.

eakorolev:
Thank you! I'm glad that bug is fixed now 🙌

PR #1994, opened 2026-06-30
drop Python 3.8 support

RobertRostohar:


TeoMahnic (approved):
LGTM

cederom:
any particular reason to remove python 3.8 like causing problems etc?

RobertRostohar:
There are some advantages when moving to higher version, mostly about cleaner syntax, better maintainability, ...
Another reason is CI - one less version to test and also potential issues with using a version which is EOL for almost two years.

PR #2012, opened 2026-07-20
pack: builder: allow explicit RAM range outside memory map

TeoMahnic:


jreineckearm (approved):
Code LGTM, couldn't test due to lack of HW at hand.

PR #2020, opened 2026-08-07
gdbserver: honor resume_on_disconnect on last client detach

jhqian:
The multi-client gdbserver resumes a halted target when the last GDB client detaches. This resume path did not consult the resume_on_disconnect session option, so the option had no effect on persistent gdbserver sessions where clients detach and reattach. Gate the resume on the option.

### DIFF
--- a/pyocd/gdbserver/gdbserver.py
+++ b/pyocd/gdbserver/gdbserver.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2006-2020,2025-2026 Arm Limited
 # Copyright (c) 2021-2022 Chris Reed
 # Copyright (c) 2022 Clay McClure
+# Copyright (c) 2026 Ryan QIAN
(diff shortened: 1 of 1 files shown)


jhqian:
The "Functional tests Mac" failure is infrastructure: the job was auto-failed with "no enabled agents online to process the request for more than 7 days" — no tests actually ran on macOS.

The 7 reported failed test cases (gdbserver on max32630/nrf51/nrf52833, Linux + Windows) are the long-standing Gdb Test failures that automated_test.py has carried in IGNORE_FAILURE_TESTS since 2021 (c5c33c10, "test: automated_test: ignore failures of connect and gdb tests"). The failure point is the DWT watchpoint exhaustion section, unrelated to the resume_on_disconnect gate added here. 

All other functional tests pass on Linux and Windows with this change.

PR #2021, opened 2026-08-17
probe: fix J-Link swo_read crash when swo_num_bytes() returns a negative value

### LINKED ISSUE #2019: J-Link SWVReader crashes when swo_num_bytes() returns a negative value
## Environment

- pyOCD: 0.45.1
- pylink-square: 1.4.0
- Python: 3.12
- J-Link DLL: 8.94
- Probe: SEGGER J-Link V9
- Target: STM32WL5MOCHx (STM32WL55 Cortex-M4)
- OS: Windows

## Description

When using SWVReader with a J-Link probe, the SWV reader thread crashes with:

IndexError: invalid index

**Note:** The same firmware and logger work correctly with an ST-LINK V3. The issue appears to be specific to the J-Link backend.

## Traceback
It appears that self._link.swo_num_bytes() occasionally returns a negative value, which is passed directly to swo_read().

The traceback points to: pyocd/probe/jlink_probe.py

```
def swo_read(self):
    return self._link.swo_read(
        0,
        self._link.swo_num_bytes(),
        True
    )
```

The exception originates from pylink:
```
buf = (ctypes.c_uint8 * num_bytes)(0)
IndexError: invalid index
```

## Possible Fix

I tested the following change:

```
def swo_read(self):
    count = self._link.swo_num_bytes()

    if count <= 0:
        return []

    return self._link.swo_read(0, count, True)
```

After applying this change (via a monkey patch), SWO logging works correctly with the J-Link, and I have not observed any further crashes.

The change only affects the case where swo_num_bytes() returns zero or a negative value and otherwise preserves the existing behavior.

adityaanikam:
Fixes #2019.

JLinkProbe.swo_read() passed self._link.swo_num_bytes() straight into swo_read() as the count.
pylink allocates a ctypes buffer sized to that count, so passing 0 -- which is normal when no SWO
data is available -- crashes with IndexError, which isn't a JLinkException and so isn't caught by
the existing except clause either; it kills the SWV reader thread outright.

self._link.swo_num_bytes() never returns a negative value (pylink validates and raises internally
if the underlying call fails), so the guard only needs to check for zero, not negative -- thanks to
@RobertRostohar for catching that in review.

Guards the count before the call and returns an empty bytearray() when it's zero, matching
DebugProbe.swo_read's documented return type.

No test added, JLinkProbe's constructor calls into the real J-Link DLL and raises ProbeError if it's
not present, so testing this cleanly would mean bypassing normal construction in a way nothing else
in this codebase does.

### DIFF
--- a/pyocd/probe/jlink_probe.py
+++ b/pyocd/probe/jlink_probe.py
@@ -3,6 +3,7 @@
 # Copyright (c) 2021-2022 Chris Reed
 # Copyright (c) 2023 Marian Muller Rebeyrol
 # Copyright (c) 2026 Christophe Dufaza
+# Copyright (c) 2026 Aditya Nikam
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -498,7 +499,10 @@ class JLinkProbe(DebugProbe):
 
     def swo_read(self):
         try:
-            return self._link.swo_read(0, self._link.swo_num_bytes(), True)
+            count = self._link.swo_num_bytes()
+            if count <= 0:
+                return bytearray()
+            return self._link.swo_read(0, count, True)
         except JLinkException as exc:
             raise self._convert_exception(exc) from exc
 
--- a/pyocd/probe/jlink_probe.py
+++ b/pyocd/probe/jlink_probe.py
@@ -500,7 +500,7 @@ class JLinkProbe(DebugProbe):
     def swo_read(self):
         try:
             count = self._link.swo_num_bytes()
-            if count <= 0:
+            if count == 0:
                 return bytearray()
             return self._link.swo_read(0, count, True)
         except JLinkException as exc:


RobertRostohar:
Actually self._link.swo_num_bytes() does not retrun a negative value (already caught in pylink with error raised).

It can however return 0 which is normal when there is no data. When 0 is passed to pylink swo_read() it crashes with IndexError.

The required guard should only check for zero value.

Please update the code and commit message.

pyocd/probe/jlink_probe.py
```diff
@@ -498,7 +499,10 @@ def swo_stop(self):
 
     def swo_read(self):
         try:
-            return self._link.swo_read(0, self._link.swo_num_bytes(), True)
+            count = self._link.swo_num_bytes()
+            if count <= 0:
```
RobertRostohar:
```suggestion
            if count == 0:
```

RobertRostohar (changes requested):

adityaanikam:
Good catch, thanks , you're right that swo_num_bytes() can't go negative since pylink validates that internally. Updated the guard to `count == 0` and rewrote the commit message/description to describe the zero case accurately instead of negative.

PR #2029, opened 2026-09-14
dap: fix incorrect shift of DPv3 root component base address

flit:
With a root component base address of 0 like most existing ADIv6 devices have, the bug was not visible. But with a non-zero base address, the low 32 bits of the computed address were incorrectly right-shifted by 12.

BASEPTR0_PTR_SHIFT is no longer used so was removed.


RobertRostohar:
This was already addressed with fccfe3e in `develop` branch.

